PR #1, opened 2024-01-16
Add css file and rename js file

trymebytes:
This PR adds the CSS file that will contain all the styles for the plugin.


psrpinto (approved):
Looks good! 👍

PR #2, opened 2024-01-16
Local environment with wp-env

psrpinto:
I wanted to be able to run the plugin locally, so I've set things up so that it runs in wp-env. Also added minimal documentation in a README. I've also set up composer so we can add scripts to `composer.json`.

Note this uses the `local` branch of GlotPress so we can easily set things up.

Hope this makes sense, let me know what you think.


wporg-gp-translation-events.php
```diff
@@ -12,14 +12,19 @@
  * @package Translation Events
  */
 
-require_once __DIR__ . '/includes/class-wporg-gp-translation-events-route.php';
+if ( class_exists( GP_Route::class ) ) {
```
psrpinto, line 15: This is here because when wp-env boots, it will try to activate both GlotPress, and this plugin. However, GlotPress does not activate because the permalink structure setting is not correct. So the `GP_Route` class (and I assume all other GlotPress code?) is not available.

Since the class is not found, wp-env fails to start with an error. That's why I added this check here, and also below, in the `register_routes()` function. This makes wp-env able to start correctly.

trymebytes (approved):
Tested and works for me!

psrpinto:
Cool, thanks for testing!

PR #3, opened 2024-01-16
Keep a log of actions users performed on translations during a translation event

### LINKED ISSUE #7: Store translation stats
We would create a new table `wp_wporg_gp_translation_events_actions` with the following columns;
- user_id
- event_id
- translation_id
- type
- datetime

Upon every translation save, we check for which events the user has opted in and whether it is active and applies and then add a row to this table.

psrpinto:
Fixes #7 

This PR adds a `wp_wporg_gp_translation_events_actions` table to the database, add adds entries to said table when new actions on translations happen in GlotPress (`created`, `rejected`, etc).

## Summary of changes

- Define database schema in `schema.sql`.
    - If you're using `wp-env`, you can apply the schema with `composer dev:db:schema`.
    - In production, the schema will need to be applied manually, the plugin does not apply it automatically.
- Create a translation _listener_, which "listens" to newly-created translations in GlotPress, and persists them into the `wp_wporg_gp_translation_events_actions` table.
    - An entry is created for each currently-ongoing event for which the user is registered for.

## TODO

- [x] What should happen to stored stats when the start or end datetimes of an event are edited?
- [x] Rename `created_at` to `happened_at`, so that it's not confusing because there will be different types of actions, and one of them is `created`
- [x] Add an `action` column
- [x] Add primary key for `event_id`, `translation_id`, `user_id`, `action`
- [x] Remove indexes (can be added back when/if needed)
- [x] Make sure that for a given event and translation, the user cannot do multiple actions of the same type

## Next steps

- https://github.com/WordPress/wporg-gp-translation-events/issues/16

## Screen captures
Screenshot 2024-01-24 at 13 08 16

Screenshot 2024-01-24 at 13 07 35

### DIFF
--- a/includes/class-wporg-gp-translation-events-translation-listener.php
+++ b/includes/class-wporg-gp-translation-events-translation-listener.php
@@ -2,17 +2,18 @@
 
 class WPORG_GP_Translation_Events_Translation_Listener {
 	const ACTIONS_TABLE_NAME = 'wp_wporg_gp_translation_events_actions';
+	const ACTION_CREATED = 'created';
 
 	public function start(): void {
 		add_action(
 			'gp_translation_created',
 			function ( $translation ) {
-				$this->handle_action( $translation );
+				$this->handle_action( $translation, self::ACTION_CREATED );
 			},
 		);
 	}
 
-	private function handle_action( GP_Translation $translation ): void {
+	private function handle_action( GP_Translation $translation, string $action ): void {
 		$now = new DateTime( 'now', new DateTimeZone( 'UTC' ) );
 
 		// Get events that are active now, for which the user is registered for.
@@ -20,14 +21,15 @@ class WPORG_GP_Translation_Events_Translation_Listener {
 		$events        = $this->select_events_user_is_registered_for( $active_events, $translation->user_id );
 
 		foreach ( $events as $event ) {
-			$this->persist( $event, $translation, $now );
+			$this->persist( $event, $translation, $now, $action );
 		}
 	}
 
 	private function persist(
 		WP_Post $event,
 		GP_Translation $translation,
-		DateTime $happened_at
+		DateTime $happened_at,
+		string $action
 	): void {
 		/** @var GP_Translation_Set $translation_set */
 		$translation_set = ( new GP_Translation_Set )->find_one( [ 'id' => $translation->translation_set_id ] );
@@ -40,6 +42,7 @@ class WPORG_GP_Translation_Events_Translation_Listener {
 				'user_id'        => $translation->user_id,
 				'translation_id' => $translation->id,
 				'happened_at'    => $happened_at->format( 'Y-m-d H:i:s' ),
+				'action'         => $action,
 				'locale'         => $translation_set->locale,
 			]
 		);
--- a/includes/class-wporg-gp-translation-events-translation-listener.php
+++ b/includes/class-wporg-gp-translation-events-translation-listener.php
@@ -3,16 +3,16 @@
 class WPORG_GP_Translation_Events_Translation_Listener {
 	const ACTION_TYPE_CREATED = 'created';
 
-	function start(): void {
+	public function start(): void {
 		add_action(
 			'gp_translation_created',
 			function ( $translation ) {
-				$this->persist( $translation, self::ACTION_TYPE_CREATED );
+				$this->handle_action( $translation, self::ACTION_TYPE_CREATED );
 			},
 		);
 	}
 
-	private function persist( GP_Translation $translation, string $action_type ): void {
+	private function handle_action( GP_Translation $translation, string $action_type ): void {
 		// TODO.
 	}
 }
--- a/includes/class-wporg-gp-translation-events-translation-listener.php
+++ b/includes/class-wporg-gp-translation-events-translation-listener.php
@@ -1,6 +1,7 @@
 <?php
 
 class WPORG_GP_Translation_Events_Translation_Listener {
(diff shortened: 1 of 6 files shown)


includes/class-wporg-gp-translation-events-translation-listener.php
```diff
@@ -0,0 +1,88 @@
...
+	 * @param WP_Post[] $events
+	 *
+	 * @return WP_Post[]
+	 */
+	private function select_events_user_is_registered_for( array $events, int $user_id ): array {
+		return array_filter(
+			$events,
+			function ( $event ) {
+				// TODO.
+				return true;
```
psrpinto, line 92: Currently, the user is registered for all events. Once the opt-in feature has been implemented, this needs to be adapted.

includes/class-wporg-gp-translation-events-translation-listener.php
```diff
@@ -0,0 +1,88 @@
...
+				'locale'         => $translation_set->locale,
+			]
+		);
+	}
+
+	/**
+	 * @return WP_Post[]
+	 */
+	private function get_active_events( DateTime $at ): array {
+		return get_posts(
```
akirk, line 60: Down the line we need to add caching here to avoid running this query on every save.

includes/class-wporg-gp-translation-events-translation-listener.php
```diff
@@ -0,0 +1,88 @@
...
+
+		global $wpdb;
+		$wpdb->insert(
+			self::ACTIONS_TABLE_NAME,
+			[
+				'event_id'       => $event->ID,
+				'user_id'        => $translation->user_id,
+				'translation_id' => $translation->id,
+				'created_at'     => $created_at->format( 'Y-m-d H:i:s' ),
+				'locale'         => $translation_set->locale,
```
akirk, line 50: We should store whether this was a `transation` (translation gets created) or a `review` event (translation gets modified: `status` moves from `waiting` to `current` or `rejected`).

schema.sql
```diff
@@ -0,0 +1,18 @@
+create or replace table wp_wporg_gp_translation_events_actions
+(
+    event_id       int(10)     not null comment 'ID of the event',
+    user_id        int(10)     not null comment 'ID of the user who authored the translation',
+    translation_id int(10)     not null comment 'ID of the translation',
+    created_at     datetime    not null comment 'When the translation was created, in UTC',
+    locale         varchar(10) not null comment 'Locale of the translation'
+)
```
akirk: We discussed to add a primary key for `event_id` `user_id` and `translation_id` and add a `type` field.

psrpinto:
Putting this back in draft until I address remarks.

includes/class-wporg-gp-translation-events-translation-listener.php
psrpinto: Opened #13 to track this.

schema.sql
```diff
@@ -0,0 +1,18 @@
+create or replace table wp_wporg_gp_translation_events_actions
+(
+    event_id       int(10)     not null comment 'ID of the event',
+    user_id        int(10)     not null comment 'ID of the user who authored the translation',
+    translation_id int(10)     not null comment 'ID of the translation',
+    created_at     datetime    not null comment 'When the translation was created, in UTC',
```
psrpinto: @akirk @amieiro We discussed and agreed on removing `created_at`, but I think we need to keep it. 

Because the start and end datetimes of the event can be edited, it's possible that the table contains entries that no longer belong in the event. For example

1. The event is created with start datetime A
2. The event starts (now = A)
3. User creates a translation
4. The event is edited to start at datetime B (which is later than A)
5. The table would contain a translation that no longer belongs in the event

To fix this, we could prevent the event start and end datetimes from being edited after the event has started. However, I think that would be poor UX for a user who creates/edits events.

If, for example, the user made a mistake and only realises it after the event started, they would probably create a new event, so you'd now have two entries for the same event, which would be confusing. So the user would also probably want to to delete events.

Instead, we could remove all these problems by having stats calculation filter out actions that took place in a datetime that does not fit the current start and end datetimes of the event.

To be able to do this, the `created_at` column needs to exist, I think.

This turned out to be a longer text than I originally thought.

Does this all make sense? I think it's important to ensure stats are correct.

psrpinto:
This is ready for re-review. We're currently only tracking `create` actions, I opened #16 for tracking other actions (`reject`, `approve`, etc), and would be tackling that in #17.

wporg-gp-translation-events.php
```diff
@@ -200,4 +187,14 @@ function register_translation_event_js() {
 add_action( 'init', 'register_event_post_type' );
 add_action( 'add_meta_boxes', 'event_meta_boxes' );
 add_action( 'save_post', 'save_event_meta_boxes' );
-register_routes();
+
+if ( class_exists( GP::class ) ) {
```
akirk: This should rather be done in a filter since it's not guaranteed that the GlotPress plugin will be loaded before this one.
```suggestion
add_action( 'gp_init', function() {
```

akirk (approved):
Tested locally and works well, I did have one recommendation. Obviously there is more to do but it works for the current iteration.

wporg-gp-translation-events.php
psrpinto: That makes a lot of sense. I fixed it in 9e535543602fe00211ca8471f221e088a3daaf8a.

schema.sql
akirk: I would prefer to just use `insert ignore` instead of `replace` so that this cannot happen.

PR #12, opened 2024-01-19
Use datetime-local and store time zone

### LINKED ISSUE #4: Save event start and end as datetime, and also save timezone
Currently, event start and end dates are stored as `YYYY-mm-dd`. This means that it's not possible to have events that last only a portion of the day, for example 18h00 to 21h00. I think it would be useful to support partial-day events, for example WordPress meetups and such.

Additionally, no timezone information is stored. I think the timezone is important because otherwise it's not possible to accurately tell whether a translation is part of an event or not.

trymebytes:
In this PR, we change the date fields to `datetime-local`. We are also now allowing organizers set the time zone for an event  storing the timezone as a post meta.

See screenshot below;
Screenshot 2024-01-25 at 00 44 49

### DIFF
--- a/assets/js/translation-events.js
+++ b/assets/js/translation-events.js
@@ -1,7 +1,9 @@
 ( function( $, $gp ) {
 jQuery(document).ready(function($) {
     $gp.notices.init();
-    selectUserTimezone();
+    if ( $('#create-event-form').length ) {
+        selectUserTimezone();
+    }
     validateEventDates();
     $('#submit-event, #edit-translation-event').on('click', function(e) {
         if ( $('#event-end').val() <= $('#event-start').val() ) {
--- a/assets/js/translation-events.js
+++ b/assets/js/translation-events.js
@@ -4,7 +4,7 @@ jQuery(document).ready(function($) {
     selectUserTimezone();
     validateEventDates();
     $('#submit-event, #edit-translation-event').on('click', function(e) {
-        if ( $('#event-end-date').val() <= $('#event-start-date').val() ) {
+        if ( $('#event-end').val() <= $('#event-start').val() ) {
             $gp.notices.error( 'Event end date and time must be later than event start date and time.' );
             return;
         }
@@ -25,8 +25,8 @@ jQuery(document).ready(function($) {
     });
 
     function validateEventDates() {
-        var startDateTimeInput = $('#event-start-date');
-        var endDateTimeInput = $('#event-end-date');
+        var startDateTimeInput = $('#event-start');
+        var endDateTimeInput = $('#event-end');
     
         startDateTimeInput.add(endDateTimeInput).on('input', function () {
             endDateTimeInput.prop('min', startDateTimeInput.val());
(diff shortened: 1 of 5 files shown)


akirk:
Can we select the user's timezone? Here is the JS snippet to do it:

```js
document.querySelector(`#event-timezone option[value="${Intl.DateTimeFormat().resolvedOptions().timeZone}"]`).selected = true 
```

assets/js/translation-events.js
```diff
@@ -1,7 +1,14 @@
 ( function( $, $gp ) {
 jQuery(document).ready(function($) {
-    $gp.notices.init(),
+    $gp.notices.init();
+    selectUserTimezone();
```
psrpinto: In the edit page, if the event has a timezone, this resets it to the user's timezone. I guess we should only call this if the event has no timezone?

wporg-gp-translation-events.php
```diff
@@ -168,6 +192,7 @@ function submit_event_ajax() {
 	}
 	update_post_meta( $event_id, '_event_start_date', $start_date );
```
psrpinto: AFAICT, the start and end datetimes are currently stored as values in the user's timezone. I think they should be stored in UTC instead, like discussed in https://github.com/WordPress/wporg-gp-translation-events/issues/4#issuecomment-1896466894

I'm imagining something along these lines (not tested :) ):

```php
$start_date = sanitize_text_field( $_POST['event_start_date'] );
$start_date = new DateTime( $start_date, new DateTimeZone( $event_timezone ) );
$start_date->setTimezone( DateTimeZone::UTC );
$start_date = $start_date->format( 'Y-m-d H:i:00' ); // The ':00' sets seconds to 0
```

Then, when displaying the event, we would need to convert from UTC to the event's timezone.

wporg-gp-translation-events.php
```diff
@@ -135,13 +153,19 @@ function submit_event_ajax() {
...
-	$title        = sanitize_text_field( $_POST['event_title'] );
-	$description  = sanitize_text_field( $_POST['event_description'] );
-	$start_date   = sanitize_text_field( $_POST['event_start_date'] );
-	$end_date     = sanitize_text_field( $_POST['event_end_date'] );
-	$locale       = sanitize_text_field( $_POST['event_locale'] );
-	$project_name = sanitize_text_field( $_POST['event_project_name'] );
+	$title          = sanitize_text_field( $_POST['event_title'] );
+	$description    = sanitize_text_field( $_POST['event_description'] );
+	$start_date     = sanitize_text_field( $_POST['event_start_date'] );
+	$end_date       = sanitize_text_field( $_POST['event_end_date'] );
```
psrpinto: Just a detail, we could rename `event_start_date` to `event_start` since it's not a date any more, but a datetime. Would also apply to `event_end_date`, and to the `meta_key` of each.

psrpinto (commented):
Looks good! Got a couple of remarks

psrpinto (approved):
👍

PR #21, opened 2024-01-27
Unique event slug

trymebytes:
Ensure that each event has a unique slug, in the case where an event with the same title exists, append a number to the end of the slug of the new event.

### DIFF
--- a/wporg-gp-translation-events.php
+++ b/wporg-gp-translation-events.php
@@ -134,6 +134,32 @@ function save_event_meta_boxes( $post_id ) {
 	}
 }
 
+/**
+ * Generate a unique event slug.
+ *
+ * @param  string $title The event title.
+ * @param  int    $post_id The event post ID.
+ * @return string The unique event slug.
+ */
+function generate_unique_event_slug( $title, $post_id = 0 ) {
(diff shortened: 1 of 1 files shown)


wporg-gp-translation-events.php
```diff
@@ -134,6 +134,32 @@ function save_event_meta_boxes( $post_id ) {
...
 }
 
+/**
+ * Generate a unique event slug.
+ *
+ * @param  string $title The event title.
+ * @param  int    $post_id The event post ID.
+ * @return string The unique event slug.
+ */
+function generate_unique_event_slug( $title, $post_id = 0 ) {
```
psrpinto, line 144: I noticed that there's a [`wp_unique_post_slug()` function](https://developer.wordpress.org/reference/functions/wp_unique_post_slug/). Would it make sense to use it here?

psrpinto (approved):

wporg-gp-translation-events.php
trymebytes: Thank you for this!  So, the `generate_unique_event_slug` function will no longer be needed. I'll go ahead and close this PR.

PR #22, opened 2024-01-29
Display stats in event page

### LINKED ISSUE #6: Display stats in event page
The event page should contain stats from the event, stats like;

- number of attendees
- number of strings translated in total
- number of translations approved
- number of translations by locale

psrpinto:
Fixes #6 

This PR makes it so that stats are displayed in the event page.

There's a lot of commits because I initially was calculating stats in PHP and then changed it to having MySQL do it. So it's probably best to review the end result instead of individual commits.

I don't think any indexes are needed, as it seems to be using the primary key (see the explain below).

### UI

Screenshot 2024-02-07 at 15 27 40

### Explain of the query

Screenshot 2024-02-07 at 15 27 40

## TODO

- [x] Display number of attendees
- [x] Group all review actions (approve, reject, etc) under one number
- [x] Display number of translations by locale
- [x] Change code so that calculations run in the database
- [x] Add necessary indexes to the table
- [x] Remove `happened_at` from the stats table
- [x] Change order of primary key to `event_id, user_id, translation_id`
- [x] Add an index for `event_id, locale` so that the query doesn't have to do filesorting.
- [x] Change `replace` to `insert ignore`
- [x] Simplify query so that it calculates total actions and compute (in code) the reviewed as the diff between total and created


includes/class-wporg-gp-translation-events-stats-calculator.php
```diff
@@ -0,0 +1,108 @@
...
+		global $wpdb;
+
+		$query = $wpdb->prepare( "
+				select locale,
+					   sum(if(action = 'create', 1, 0)) AS created,
+					   sum(if(action in ('approve', 'reject', 'request_changes', 'mark_fuzzy'), 1, 0)) AS reviewed,
+					   count(distinct user_id) as users
+				from $this->ACTIONS_TABLE_NAME
+				where event_id = %d
+				  and happened_at between cast('%s' as datetime) and cast('%s' as datetime)
```
akirk: Can we please drop this? Event entries should only go in the table when its the time of the event.

includes/class-wporg-gp-translation-events-stats-calculator.php
```diff
@@ -0,0 +1,108 @@
...
+				group by locale with rollup
```
akirk, line 70: Because of this we should add an `event_id, locale` index to the table so that doesn't have to do filesorting.

includes/class-wporg-gp-translation-events-stats-calculator.php
```diff
@@ -0,0 +1,108 @@
...
+	function for_event( WP_Post $event ): WPORG_GP_Translation_Events_Event_Stats {
+		$start = new DateTime( get_post_meta( $event->ID, '_event_start', true ), new DateTimeZone( 'UTC' ) );
+		$end   = new DateTime( get_post_meta( $event->ID, '_event_end', true ), new DateTimeZone( 'UTC' ) );
+		$stats = new WPORG_GP_Translation_Events_Event_Stats;
+		global $wpdb;
+
+		$query = $wpdb->prepare( "
+				select locale,
+					   sum(if(action = 'create', 1, 0)) AS created,
+					   sum(if(action in ('approve', 'reject', 'request_changes', 'mark_fuzzy'), 1, 0)) AS reviewed,
```
akirk:
```suggestion
					   sum(action = 'create') AS created,
					   count(*) AS total,
```

Can we make this simpler and calculate the reviewed as `total - created` below?

akirk: [MySQL's SUM function](https://dev.mysql.com/doc/refman/8.0/en/functions.html#function_sum) is defined as `SUM([DISTINCT] expr) [over_clause]` so you don't need the if.

includes/class-wporg-gp-translation-events-stats-calculator.php
akirk: While you're adding the index, could you please also make the primary index `event_id, user_id, translation_id`? I think it's much more likely that we might later want to do a query of all translations a certain user did in an event than which users have an entry on a translation id (which at most would be two). MySQL can only use the index in sequence, so when querying for `event_id` and `user_id`, it can only use the `event_id` part of an `event_id, translation_id, user_id` index but `event_id` and `user_id` for a `event_id, user_id, translation_id` index.

akirk (changes requested):

includes/class-wporg-gp-translation-events-stats-calculator.php
psrpinto: I'm convinced this is needed. I've documented why that is in a a previous PR: https://github.com/WordPress/wporg-gp-translation-events/pull/3#discussion_r1462170267

includes/class-wporg-gp-translation-events-stats-calculator.php
psrpinto: Ah, so that's where the filesorting is coming from, thanks! Agree on changing the order of columns in the index too, will fix both these issues.

includes/class-wporg-gp-translation-events-stats-calculator.php
akirk: Ok, I read the comment and I think the solution needs to be to replace the `replace` with an `insert ignore`. We should not be updating the row.

psrpinto: I'm having trouble understanding how the `insert ignore` would solve the problem, I think maybe I didn't explain what the problem is correctly.

Let's say I schedule WordCamp Lisbon for tomorrow. At 8am of the event's day a user adds a translation. However, at 9am I realize it's raining that day so I decide to reschedule the event to next week. 

Should the translation created by the user be included in the stats of the event?

akirk: What I am saying that during an event is active, submitting a translation or review should be counted towards the event. If you do some event date juggling so that the event happens again a second time then those translations should also count.
I don't think we need to account for these changes whatsoever at stats display. When the translation stats entry for an event_id exists, we can count it. Not that it's planned to add this kind of option, but what if we had a recurring Friday evening translations event?

psrpinto: > what if we had a recurring Friday evening translations event?

This is a great point, it made me realise that indeed stats counting should be "dumb" and just use what is in the table, like you mention.

So I will remove `happened_at` from the table and adapt the query so it doesn't take it in account. I will also change `replace` to `insert ignore` when storing the action.

includes/class-wporg-gp-translation-events-stats-calculator.php
psrpinto: Makes sense!

includes/class-wporg-gp-translation-events-stats-calculator.php
akirk: Thank you!

psrpinto:
I believe I've addressed all remarks, this is ready for re-review.

templates/event.php
```diff
@@ -23,4 +30,34 @@
...
+			<tr>
+				<td><?php echo esc_html( $locale ) ?></td>
+				<td><?php echo esc_html( $row->created ) ?></td>
+				<td><?php echo esc_html( $row->reviewed ) ?></td>
+				<td><?php echo esc_html( $row->users ) ?></td>
+			</tr>
+			<?php endforeach ?>
+			<tr class="event-details-stats-totals">
+				<td>Total</td>
+				<td><?php echo esc_html( $event_stats->totals()->created ) ?></td>
```
akirk, line 57: This will cause a `Fatal error: Uncaught Error: Typed property WPORG_GP_Translation_Events_Event_Stats::$totals must not be accessed before initialization` if the table has no stats.

akirk (changes requested):
Nice thanks! Almost there!

templates/event.php
psrpinto: Good catch, I had fixed it #35 but I've now cherry-picked the commit into here.

psrpinto:
@akirk Could you take one last look? Only one commit since you last reviewed :)

templates/event.php
```diff
@@ -43,14 +44,14 @@
...
-			<?php foreach ( $event_stats->rows() as $locale => $row ): ?>
+		<?php foreach ( $event_stats->rows() as $locale => $row ): ?>
 			<tr>
 				<td><?php echo esc_html( $locale ) ?></td>
 				<td><?php echo esc_html( $row->created ) ?></td>
 				<td><?php echo esc_html( $row->reviewed ) ?></td>
 				<td><?php echo esc_html( $row->users ) ?></td>
 			</tr>
-			<?php endforeach ?>
+		<?php endforeach ?>
```
akirk, line 54: I believe this misses a trailing semicolon. At least to the coding standards.

templates/event.php
```diff
@@ -60,4 +61,5 @@
 			</tbody>
 		</table>
 	</div>
+<?php endif ?>
```
akirk, line 64: Here also

akirk (changes requested):

akirk (approved):
I've changed my mind and will approve it now and we'll fix this in #44

PR #27, opened 2024-01-30
Fix broken route loading

### LINKED ISSUE #23: Routes not loading
The custom routes for events are no longer loading, they return a 404.

To test, switch to the `trunk` branch and visit the `glotpress/events/new` in your local install, it returns a 404.

image

trymebytes:
All the events route stopped return a 404, this has been fixed in this PR by reordering the routes registration.

### DIFF
--- a/wporg-gp-translation-events.php
+++ b/wporg-gp-translation-events.php
@@ -257,10 +257,10 @@ add_action(
 	'gp_init',
 	function() {
 		require_once __DIR__ . '/includes/class-wporg-gp-translation-events-route.php';
-		GP::$router->prepend( '/events/([a-z0-9_-]+)', array( 'WPORG_GP_Translation_Events_Route', 'events_details' ), 'get' );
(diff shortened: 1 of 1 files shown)


wporg-gp-translation-events.php
```diff
@@ -257,10 +257,10 @@ function register_translation_event_js() {
 	'gp_init',
 	function() {
 		require_once __DIR__ . '/includes/class-wporg-gp-translation-events-route.php';
+		GP::$router->prepend( '/events/([a-z0-9_-]+)', array( 'WPORG_GP_Translation_Events_Route', 'events_details' ), 'get' );
```
akirk: I'd prefer if we changed this from `prepend` to `add` and leave the order the same. That way it'll be easier for future addition of routes without having to realize that the order in the routing table will be reversed because of all the `prepend`s.

akirk (changes requested):

psrpinto (approved):

akirk (approved):

PR #31, opened 2024-02-01
Use single event form

### LINKED ISSUE #30: Single form for creating and editing events
We have two forms at the moment, one for creating events and the other for editing events. If a change is made in one and not made in the other, this could break the system. We need to ensure that just one form is used for both event creation and event editing.

trymebytes:
Fixes #30 

**Create Event Form**
image

In local GlotPress
translationevent2


includes/class-wporg-gp-translation-events-route.php
```diff
@@ -37,7 +37,19 @@ public function events_create() {
...
+		$event_form_name    = 'create_event';
+		$css_show_url       = 'hide-event-url';
+		$event_id           = null;
+		$event_title        = '';
+		$event_description  = '';
+		$event_timezone     = '';
+		$event_start        = '';
+		$event_end          = '';
+		$event_locale       = '';
+		$event_project_name = '';
```
psrpinto, line 50: In the future, we could consider refactoring this so that there is an `Event` object that gets passed to the template, instead of passing individual variables.

I think having an `Event` class would also simplify code in other parts of the plugin. For example, conversion of dates to/from UTC could be centralized in the Event class.

wporg-gp-translation-events.php
```diff
@@ -153,38 +153,35 @@ function validate_event_dates( $event_start, $event_end ) {
...
-	$locale         = sanitize_text_field( $_POST['event_locale'] );
-	$project_name   = sanitize_text_field( $_POST['event_project_name'] );
-	$event_timezone = sanitize_text_field( $_POST['event_timezone'] );
+	$title          = isset( $_POST['event_title'] ) ? sanitize_text_field( $_POST['event_title'] ) : '';
+	$description    = isset( $_POST['event_description'] ) ? sanitize_text_field( $_POST['event_description'] ) : '';
+	$event_start    = isset( $_POST['event_start'] ) ? sanitize_text_field( $_POST['event_start'] ) : '';
+	$event_end      = isset( $_POST['event_end'] ) ? sanitize_text_field( $_POST['event_end'] ) : '';
+	$locale         = isset( $_POST['event_locale'] ) ? sanitize_text_field( $_POST['event_locale'] ) : '';
+	$project_name   = isset( $_POST['event_project_name'] ) ? sanitize_text_field( $_POST['event_project_name'] ) : '';
+	$event_timezone = isset( $_POST['event_timezone'] ) ? sanitize_text_field( $_POST['event_timezone'] ) : '';
```
psrpinto, line 168: If we eventually introduce an `Event` class (like suggested in another comment), this could be moved to a "static constructor" on that class, something like this:


```php
class Event {
	private DateTime $start;
	// ...

	private function __construct(
		DateTime $start,
		// ...
	) {
		$this->start = $start;
		// ...
	}

	public static function fromForm(): Event {
		$start = new DateTime(isset( $_POST['event_start'] ) ? sanitize_text_field( $_POST['event_start'] ) : '');

		return new Event(
			$start,
			// ...
		);
	}
}
```

psrpinto (approved):
Makes a lot of sense to have a single form.
Tested locally and it works 👍

assets/js/translation-events.js
```diff
@@ -1,27 +1,43 @@
...
         var $form = $('.translation-event-form');
+
         $.ajax({
             type: 'POST',
             url: $translation_event.url,
             data:$form.serialize(),
             success: function(response) {
-                $gp.notices.success(response.data);
+                if ( response.data.eventId ) {
+                    history.replaceState('','', '/glotpress/events/edit/' + response.data.eventId)
```
akirk: This should use the `home_url()`

templates/events-form.php
```diff
@@ -0,0 +1,64 @@
...
+		<label for="event-locale">Locale:</label>
+		<input type="text" id="event-locale" name="event_locale" value="<?php echo esc_attr( $event_locale ); ?>">
+	</div>
+	<div>
+		<label for="event-project-name">Project Name:</label>
+		<input type="text" id="event-project-name" name="event_project_name" value="<?php echo esc_attr( $event_project_name ); ?>">
+	</div>
+	<div class="submit-btn-group">
+		<?php if ( ( isset( $event_status ) && 'draft' === $event_status ) || ! $event_id ) : ?>
+			<button class="button is-primary save-draft submit-event" type="submit" data-event-status="draft">Save Draft</button>
```
akirk: Can we update the text to "Update Draft" when you first save the draft.

psrpinto (approved):
Looks good!

PR #34, opened 2024-02-01
Ignore mark as fuzzy and only track one action per user per translation

### LINKED ISSUE #29: Ignore "mark as fuzzy" actions
Currently `mark_fuzzy` is being tracked and is included in the "Translations reviewed" stat, along with `approve`, `reject` and `request_changes`.

If I understood correctly, @amieiro mentioned that "Mark as fuzzy" might be used as "postponing" the review of the translation to the future, and that it is in practice the same as leaving the translation in the Waiting queue.

So should we:

1. Track it in the database table, and include it in the stats under "Translations reviewed"? (how it works today)
2. Track it in the database table, but don't include it in the stats? (useful if we think we might change our mind later and want to start including it in the stats)
3. Don't track it in the database table.

@amieiro @akirk @trymebytes

### LINKED ISSUE #32: Don't track multiple actions on a given translation by the same user
Instead, keep only the latest action. For more info see discussion at https://github.com/WordPress/wporg-gp-translation-events/commit/20aec8a0f38d169d32d0243089aedc9363105c33#r138151268

psrpinto:
Fix #29 #32

### DIFF
--- a/includes/class-wporg-gp-translation-events-stats-calculator.php
+++ b/includes/class-wporg-gp-translation-events-stats-calculator.php
@@ -65,7 +65,7 @@ class WPORG_GP_Translation_Events_Stats_Calculator {
 		$query = $wpdb->prepare( "
 				select locale,
 					   sum(if(action = 'create', 1, 0)) AS created,
-					   sum(if(action in ('approve', 'reject', 'request_changes', 'mark_fuzzy'), 1, 0)) AS reviewed,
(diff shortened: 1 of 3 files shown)


trymebytes (approved):
Looks good to me!

psrpinto:
I'm merging this into #22 as further changes are required there which would result in conflicts here. Since there's only two commits here, I think that's fine and makes things easier.

PR #35, opened 2024-02-01
Cache list of active events

### LINKED ISSUE #13: Cache list of active events
To avoid running the query on every action on a translation.

We could cache events happening before or after X hours from when the caching takes place, and do extra filtering in code to extract the actually active events.

Cache will also need to be invalidated when something relevant happens (new event created, event times changed, etc).

psrpinto:
This PR caches the results of the query that retrieves events that are active "now", so that we're not hitting the database with an expensive query for all actions on a translation.

We use the [`wp_cache_*` functions](https://developer.wordpress.org/reference/classes/wp_object_cache). In development, we use [SQLite Object Cache](https://wordpress.org/plugins/sqlite-object-cache/) as the caching backend.

We cache events for which start is before `now + d` (where `d` is the cache duration) AND end is after `now` (see picture below). We then do extra filtering in code to extract the actually active events.

We cache events that partially or fully overlap with the "cache" box, and we exclude events that don't overlap:

IMG_20240205_164008 jpg_compressed

## Testing instructions

1. Checkout `cache-active-events`
2. `composer dev:start`
3. Go to permalink settings and click the save button
4. Create a couple of translation events which last the whole of 2024
    - I'm using the whole year as a example, feel free to use other periods, just make sure they are active "now".
3. Create a translation event which ended before 2024
6. Create a translation event that will start in 2025
7. Place a `die(var_export($events, true));` in an `else` [here](https://github.com/WordPress/wporg-gp-translation-events/blob/cache-active-events/includes/class-wporg-gp-translation-events-translation-listener.php#L138)
8. Go to glotpress and review a translation
    - Make sure you don't add a translation as that might result in an immediate approve, which would mess up with these tests. Instead just reject or approve a "waiting" translation. If there are no waiting translations you can manually change the status of existing entries in `wp_gp_translations` to `waiting`.
10. The `die()` MUST NOT have been called
6. Go to glotpress and review another translation
7. The `die()` MUST have been called and it output the event ids of only the events that are active now.
8. Edit an event
7. The `die()` MUST NOT have been called

## TODO

- [x] Install a [persistent caching plugin](https://developer.wordpress.org/reference/classes/wp_object_cache/#persistent-cache-plugins) in development environment ([SQLite Object Cache](https://wordpress.org/plugins/sqlite-object-cache/))
- [x] Expire cache after a given duration (`d`)
- [x] Cache events for which start is before `now + d` AND end is after `now` (see picture below). Do extra filtering in code to extract the actually active events.
- [x] Invalidate cache when events are created, deleted, or their start, end or timezone are edited.
- [x] Cache event start and end in addition to the id
- [x] Do extra filtering in code to extract the actually active events.
- [x] Add an `Event` class
- [x] Pass `Event` instances to the cache and have the cache return an array of `Event`

## References
- [WP_Object_Cache](https://developer.wordpress.org/reference/classes/wp_object_cache/)


includes/class-wporg-gp-translation-events-active-events-cache.php
```diff
@@ -0,0 +1,84 @@
...
+}
+
+class WPORG_GP_Translation_Events_Active_Events_Cache {
+	public const CACHE_DURATION = 60 * 60 * 24; // 24 hours.
+	private const KEY = 'translation-events-active-events';
+	private static bool $invalidated_on_event_save = false;
+
+	public function __construct() {
+		// Invalidate cache when events are modified.
+		add_filter(
```
psrpinto: This should be moved out of the constructor because if there are multiple instances of this class, we'll be listening to the hook multiple times.

includes/class-wporg-gp-translation-events-translation-listener.php
```diff
@@ -50,68 +56,123 @@ function ( $translation, $translation_before ) {
...
+						'compare' => '>',
 						'type'    => 'DATETIME',
 					],
 				],
 			],
 		);
+
+		$cache_entries = [];
+		foreach ( $event_ids as $event_id ) {
+			$meta = get_post_meta( $event_id );
```
psrpinto: Is there a way to get the post meta of multiple posts in one query?

akirk: AFAIK just with a custom query like
```
    $sql = $wpdb->prepare(
        "SELECT post_id, meta_value FROM $wpdb->postmeta WHERE meta_key = %s AND post_id IN ( $post_ids )",
        $meta_key
    );

```

includes/class-wporg-gp-translation-events-active-events-cache.php
```diff
@@ -0,0 +1,84 @@
...
+				// However, we only need to invalidate the cache once, so we check first if we already did it.
+				if ( self::$invalidated_on_event_save ) {
+					return;
+				}
+				$event_meta_keys = [ '_event_start', '_event_end', '_event_timezone' ];
+				if ( ! in_array( $meta_key, $event_meta_keys ) ) {
+					return;
+				}
+				try {
+					$this->invalidate();
```
akirk: Can this work? You're using `$this` and `self` inside a closure.

includes/class-wporg-gp-translation-events-active-events-cache.php
```diff
@@ -0,0 +1,84 @@
...
+class WPORG_GP_Translation_Events_Active_Events_Cache {
+	public const CACHE_DURATION = 60 * 60 * 24; // 24 hours.
+	private const KEY = 'translation-events-active-events';
+	private static bool $invalidated_on_event_save = false;
+
+	public function __construct() {
+		// Invalidate cache when events are modified.
+		add_filter(
+			'update_post_metadata',
+			function ( $ignore, int $object_id, string $meta_key ) {
```
akirk: I am worried about this filter. Can't we just add an explicit invalidation after we updated the post? We control that part of the code and it's not used elsewhere. Here we're also not checking the post_type of the object.

includes/class-wporg-gp-translation-events-translation-listener.php
```diff
@@ -50,68 +56,123 @@ function ( $translation, $translation_before ) {
...
 
 	/**
-	 * @return WP_Post[]
+	 * @return int[]
+	 * @throws Exception
 	 */
-	private function get_active_events( DateTime $at ): array {
-		return get_posts(
+	private function get_active_events( DateTimeImmutable $at ): array {
+		$cache_entries = $this->active_events_cache->get();
```
akirk: Can we just rename this to 
```suggestion
		$entries = $this->active_events_cache->get();
```

includes/class-wporg-gp-translation-events-translation-listener.php
```diff
@@ -50,68 +56,123 @@ function ( $translation, $translation_before ) {
...
+		if ( null === $cache_entries ) {
+			$cache_entries = $this->cache_events( $at );
```
akirk: This is a bit unusual to me. It makes the code look like it relies on a cache. What if no cache is configured? I believe this would still work but the naming seems off.

The usual way I see this implemented is that you have the "retrieve active events" logic right here, and then cache the results.

Something like this:

```php
$events = wp_cache_get( $cache_key );
if ( ! $events ) {
   $events = $wpdb->get_results( $wpdb->prepare( ... ) );
   wp_cache_set( $cache_key, $events);
}
```

akirk (changes requested):
I think we should rethink the usage of the `update_post_metadata` filter and I'd prefer if we'd move the database call out of the cache_events function.

includes/class-wporg-gp-translation-events-active-events-cache.php
psrpinto: I agree this should be done explicitly, will change it 👍 

The reason I did it this way was mostly because of the fact that events can be edited through wp-admin as well. So we'll also need to invalidate the cache in that case. I opened an issue about that: https://github.com/WordPress/wporg-gp-translation-events/issues/40

includes/class-wporg-gp-translation-events-translation-listener.php
psrpinto: Totally agree, will change it.

includes/class-wporg-gp-translation-events-translation-listener.php
psrpinto: Do you think it's worth implementing this, or calling `get_post_meta()` for each event would be fine? This code will only run when the cache is empty, so once per day, or when events change.

psrpinto:
I think I addressed all remarks, this is ready for another review round.

includes/class-wporg-gp-translation-events-translation-listener.php
akirk: It's fine to use `get_post_meta` for now. We can always optimize later.

includes/class-wporg-gp-translation-events-translation-listener.php
```diff
@@ -50,71 +56,102 @@ function ( $translation, $translation_before ) {
...
+			}
+
+			$this->active_events_cache->cache( $events );
+		}
+
+		// Filter out events that aren't actually active at $at.
+		return array_filter(
+			$events,
+			function ( $event ) use ( $at ) {
+				return $at >= $event->start() && $at <= $event->end();
```
akirk: Just mentioning this as an alternative formatting which makes it a tad clearer that the $at is between the two events.
```suggestion
				return $event->start() <= $at && $at <= $event->end();
```

akirk (approved):
Looks good now!

PR #36, opened 2024-02-05
Filter and style event list

### LINKED ISSUE #24: Style and filter events list
Add better css styling to the events list page and filter events to show only upcoming events.

trymebytes:
Fixes #24

**Screenshot**
image

**Update**
image

### DIFF
--- a/templates/events-list.php
+++ b/templates/events-list.php
@@ -1,14 +1,9 @@
 <?php
 gp_title( __( 'Translation Events' ) );
 gp_tmpl_header();
-
+gp_tmpl_load( 'events-header', get_defined_vars(), dirname( __FILE__ ) );
 ?>
-<div class="event-list-top-bar">
-	<ul class="event-list-nav">
-		<li><a href="<?php echo esc_url( gp_url( '/events/my-events/' ) ); ?>">My Events</a></li>
-		<li><a class="button is-primary" href="<?php echo esc_url( gp_url( '/events/new/' ) ); ?>">Create Event</a></li>
-	</ul>
-</div>
+
 
 <h2 class="event_page_title">Upcoming Translation Events</h2>
 <div class="event-left-col">
--- a/templates/events-list.php
+++ b/templates/events-list.php
@@ -6,14 +6,23 @@ gp_tmpl_header();
 
 <h2 class="event_page_title">Upcoming Translation Events</h2>
 <?php
-
-
-$_paged = ( get_query_var( 'paged' ) ) ? get_query_var( 'paged' ) : 1;
-$args   = array(
+$current_datetime_utc = ( new DateTime( null, new DateTimeZone( 'UTC' ) ) )->format( 'Y-m-d H:i:s' );
+$_paged               = ( get_query_var( 'paged' ) ) ? get_query_var( 'paged' ) : 1;
+$args                 = array(
 	'post_type'      => 'event',
 	'posts_per_page' => 10,
 	'paged'          => $_paged,
 	'post_status'    => 'publish',
+	'meta_query'     => array(
+		array(
+			'key'     => '_event_start',
+			'value'   => $current_datetime_utc,
+			'compare' => '>=',
+			'type'    => 'DATETIME',
+		),
+	),
+	'orderby'        => 'meta_value',
+	'order'          => 'ASC',
 );
 
 $query = new WP_Query( $args );
--- a/templates/events-list.php
+++ b/templates/events-list.php
@@ -33,9 +33,12 @@ if ( $query->have_posts() ) :
 		<?php
 		while ( $query->have_posts() ) :
 			$query->the_post();
+			$event_start = ( new DateTime( get_post_meta( get_the_ID(), '_event_start', true ) ) )->format('l, F j, Y');
 			?>
-			<li>
+			<li class="event-list-item">
+				<span class="event-list-date"><?php echo esc_html( $event_start ); ?></span>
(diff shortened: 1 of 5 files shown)


templates/events-list.php
```diff
@@ -3,16 +3,33 @@
...
+</div>
 
 <h2 class="event_page_title">Upcoming Translation Events</h2>
+<div class="event-left-col">
 <?php
-
-
-$_paged = ( get_query_var( 'paged' ) ) ? get_query_var( 'paged' ) : 1;
-$args   = array(
+$current_datetime_utc = ( new DateTime( null, new DateTimeZone( 'UTC' ) ) )->format( 'Y-m-d H:i:s' );
```
akirk:
```suggestion
$current_datetime_utc = ( new DateTime( 'now', new DateTimeZone( 'UTC' ) ) )->format( 'Y-m-d H:i:s' );
```

Otherwise you get an error: Deprecated: DateTime::__construct(): Passing null to parameter #1 ($datetime) of type string is deprecated

templates/events-list.php
```diff
@@ -1,14 +1,9 @@
 <?php
 gp_title( __( 'Translation Events' ) );
 gp_tmpl_header();
-
+gp_tmpl_load( 'events-header', get_defined_vars(), dirname( __FILE__ ) );
```
psrpinto, line 4: 👌

templates/events-list.php
```diff
@@ -1,18 +1,30 @@
...
+			'value'   => $current_datetime_utc,
+			'compare' => '>=',
+			'type'    => 'DATETIME',
+		),
+	),
+	'orderby'        => 'meta_value',
+	'order'          => 'ASC',
 );
 
 $query = new WP_Query( $args );
```
psrpinto, line 30: I think it would make sense to move the query out of the template and into `WPORG_GP_Translation_Events_Route::events_list()`, so that the templates are logic-less. Looking at the other templates (e.g. `events-edit`), that seems to be the case there, I think it would be good to do that for this template too.

This might not be in scope for this PR though, so we could do it in a follow-up PR.

templates/events-list.php
```diff
@@ -23,9 +35,12 @@
 		<?php
 		while ( $query->have_posts() ) :
 			$query->the_post();
+			$event_start = ( new DateTime( get_post_meta( get_the_ID(), '_event_start', true ) ) )->format('l, F j, Y');
 			?>
-			<li>
+			<li class="event-list-item">
+				<span class="event-list-date"><?php echo esc_html( $event_start ); ?></span>
 				<a href="<?php the_permalink(); ?>"><?php the_title(); ?></a> by <span><?php the_author(); ?></span>
```
psrpinto, line 42: When testing this on wp-env, `the_permalink()` sends me to `/events/test-2/` which is the WordPress post page. Should it instead send to `/glotpress/events/test-2/`?

psrpinto (commented):
Tested locally and looks great! I found something that I'm not sure is a bug or not though

templates/events-list.php
trymebytes: Yea, nice catch. Thank you!

amieiro (dismissed):
In my local environment (https://glotpress.test/glotpress/events/), in the trunk branch, I have 2 events:

image

image

If I switch to this PR, in the same URL I don't have any event:

image

trymebytes:
@amieiro My guess is that the start dates of the events have passed, could you please confirm?

amieiro:
> @amieiro My guess is that the start dates of the events have passed, could you please confirm?

I have created two events for the whole month, starting on Feb 1 and finishing on Feb 29 and both didn't appear in the list.

Translation Events – Edit Event 2024-02-09 16-59-08
Translation Events – Edit Event 2024-02-09 17-01-06

This is the result:

image

These are the event post in the database:

image

And these are their meta information:

image

trymebytes:
Yes, they won't appear because we display only events that have their start dates equal to today or later.

trymebytes:
So today being February 9, only events starting from February 9 and later will be displayed. Events from February 8 and earlier will not be displayed. The time is also taken into account in this comparison.

amieiro:
Ok, I understood incorrectly the word "Upcoming". I think it should be something like "Available events", because if the event is active (e.g., Translation Month), the best option is to show it to the possible assistants, so they can participate there. What do you think?

trymebytes:
That's a good point. I see the possibility of events scheduled to hold for several days not being included on the list once their start date is past the current date. I have created this issue #46 for us to further discuss this and get more inputs.

templates/events-list.php
amieiro, line 42: This change will get the correct link to the template:
- Example of a previous link: https://glotpress.test/events/upcoming-event/
- Example of a new link: https://glotpress.test/glotpress/events/upcoming-event/ 

You need to add previously this code: 

```
$post_slug = get_post_field( 'post_name', get_the_ID() );
```

```suggestion
				<a href="<?php echo esc_url( gp_url( '/events/' . $post_slug ) ) ?>"><?php the_title(); ?></a> by <span><?php the_author(); ?></span>
```

akirk: As I already mentioned elsewhere, I think it'd be ok to take a look at this later. But if we need to do it, we should continue to use the permalink function like this:
```suggestion
				<a href="<?php echo esc_url( gp_url( get_the_permalink() ) ); ?>"><?php the_title(); ?></a> by <span><?php the_author(); ?></span>
```

trymebytes: I think we can fix this in a follow-up PR after we agree on the best approach.

amieiro (approved):

PR #38, opened 2024-02-07
Updates in the header fields

amieiro:
This PR adds or updates some fields in the header:
- Updates the `Plugin URI`, `Author` and `Author URI` from the `Automattic` to the `WordPress` organization. 
- Adds semantic version to the plugin version.
- Adds these fields:
  - Requires at least.
  - Tested up to.
  - Requires PHP.
  -  License URI.


psrpinto (approved):
👍

PR #43, opened 2024-02-09
Attend or un-attend an event

### LINKED ISSUE #9: Register attendees
For attendees to register their attendance, they click the `Opt In` button and the event id is stored to the user meta.

psrpinto:
Users can now attend or un-attend an event. Their attendance is taken in account when saving stats.

## Screen capture

https://github.com/WordPress/wporg-gp-translation-events/assets/550401/dec3c068-9d53-4fdd-bf13-58210bf17573

### DIFF
--- a/includes/class-wporg-gp-translation-events-route.php
+++ b/includes/class-wporg-gp-translation-events-route.php
@@ -5,7 +5,7 @@
  * @package wporg-gp-translation-events
  */
 class WPORG_GP_Translation_Events_Route extends GP_Route {
-
+	private const USER_META_KEY_ATTENDING = 'translation-events-attending';
 
 	/**
 	 * WPORG_GP_Translation_Events_Route constructor.
@@ -117,7 +117,7 @@ class WPORG_GP_Translation_Events_Route extends GP_Route {
 	 *
 	 * @param int $event_id
 	 */
-	public function events_attend( int $event_id) {
+	public function events_attend( int $event_id ) {
 		$user = wp_get_current_user();
 		if ( ! $user ) {
 			$this->die_with_error( 'Only logged-in users can attend events', 403 );
@@ -128,7 +128,20 @@ class WPORG_GP_Translation_Events_Route extends GP_Route {
 			$this->die_with_404();
 		}
 
-		// TODO
+		$event_ids = get_user_meta( $user->ID, self::USER_META_KEY_ATTENDING, true ) ?? [];
+		if ( ! $event_ids ) {
+			$event_ids = [];
+		}
+
+		if ( ! array_key_exists( $event_id, $event_ids ) ) {
+			// Not yet attending, mark as attending.
+			$event_ids[ $event_id ] = true;
+		} else {
+			// Currently attending, mark as not attending.
+			unset( $event_ids[ $event_id ] );
+		}
+
+		update_user_meta( $user->ID, self::USER_META_KEY_ATTENDING, $event_ids );
 
 		wp_safe_redirect( gp_url( "/events/$event->post_name" ) );
 		exit;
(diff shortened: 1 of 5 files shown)


includes/class-wporg-gp-translation-events-route.php
```diff
@@ -91,23 +91,66 @@ public function events_edit( $event_id ) {
...
 		}
+		$event_id           = $event->ID;
 		$event_title        = $event->post_title;
 		$event_description  = $event->post_content;
 		$event_start_date   = get_post_meta( $event->ID, '_event_start_date', true ) ?? '';
 		$event_end_date     = get_post_meta( $event->ID, '_event_end_date', true ) ?? '';
 		$event_locale       = get_post_meta( $event->ID, '_event_locale', true ) ?? '';
 		$event_project_name = get_post_meta( $event->ID, '_event_project_name', true ) ?? '';
+
+		$attending_event_ids = get_user_meta( $user->ID, self::USER_META_KEY_ATTENDING, true ) ?? [];
```
akirk: We need to use `?:` since get_user_meta doesn't return `null` in the case of not found but something falsy. Likely we need to change the other lines above this one too.
```suggestion
		$attending_event_ids = get_user_meta( $user->ID, self::USER_META_KEY_ATTENDING, true ) ?: [];
```

I stumbled on this because I received this error:  Uncaught TypeError: array_key_exists(): Argument #2 ($array) must be of type array, string given in includes/class-wporg-gp-translation-events-route.php:112

includes/class-wporg-gp-translation-events-translation-listener.php
```diff
@@ -114,11 +114,13 @@ private function get_active_events( DateTime $at ): array {
 	 * @return WP_Post[]
 	 */
 	private function select_events_user_is_registered_for( array $events, int $user_id ): array {
+		$attending_event_ids = get_user_meta( $user_id, WPORG_GP_Translation_Events_Route::USER_META_KEY_ATTENDING, true ) ?? [];
+		$attending_event_ids = array_keys( $attending_event_ids );
```
akirk: There is a bug. Below you are using `array_key_exists` but here you are creating a new array with the keys as values.
```suggestion
		$attending_event_ids = get_user_meta( $user_id, WPORG_GP_Translation_Events_Route::USER_META_KEY_ATTENDING, true ) ?? [];
```

includes/class-wporg-gp-translation-events-translation-listener.php
```diff
@@ -114,11 +114,13 @@ private function get_active_events( DateTime $at ): array {
...
+
 		return array_filter(
 			$events,
-			function ( $event ) {
-				// TODO.
-				return true;
+			function ( $event ) use ( $attending_event_ids ) {
+				return array_key_exists( $event->ID, $attending_event_ids );
```
akirk: I personally prefer this since it's visually clearer to me.
```suggestion
				return isset( $attending_event_ids[ $event->ID ] );
```

akirk (changes requested):
Nice. It only worked for me after I fixed the bug. Would be nice to also see in the overview which events you are attending but we can do it in a follow-up.

includes/class-wporg-gp-translation-events-route.php
psrpinto: Ooops, good catch. Fixed all of them.

includes/class-wporg-gp-translation-events-translation-listener.php
psrpinto: Oops, fixed.

akirk (approved):
Tested, works now! We'll fix the coding standards issues in #44

amieiro (approved):

PR #45, opened 2024-02-09
Convert event time to user's local time

### LINKED ISSUE #37: Show event date and time in user's timezone
Show event date and time in user's timezone

trymebytes:
Events datetime are stored in UTC. In this PR, we ensure that event date and time are displayed in readable format and in the user's local time zone.
image

Screenshot 2024-02-12 at 10 40 09

### DIFF
--- a/assets/js/translation-events.js
+++ b/assets/js/translation-events.js
@@ -5,6 +5,7 @@ jQuery(document).ready(function($) {
         selectUserTimezone();
     }
     validateEventDates();
+    convertToUserLocalTime();
 
     $('.submit-event').on('click', function(e) {
         e.preventDefault();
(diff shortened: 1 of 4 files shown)


amieiro (approved):
We need to improve some items here, but I think we can do them in another PR. I am going to use the Spanish situation to explain them:
- We don't use AM/PM, so we should have 24 hour time.
- We use [CET](https://en.wikipedia.org/wiki/Central_European_Time)/[CEST](https://en.wikipedia.org/wiki/Central_European_Summer_Time) time, so this should be the timezone used, not UTC/GMT offset.

image

A [good example](https://www.meetup.com/es-ES/aindustriosa/events/299082708/):

image

trymebytes:
Thank you for the feedback! Yes, we can fix that in another PR.

PR #48, opened 2024-02-12
Update the link for each upcoming event

amieiro:
In the event list, if you click in the event link:

image

The user goes to the default post for the 'event' CPT:

image

With this PR, the link changes and the user goes to the event page:

image


psrpinto (approved):
Thanks!

PR #52, opened 2024-02-12
Add the current events in the event list

### LINKED ISSUE #46: Sorting events on the events page
This issue is opened as a follow up to [this comment](https://github.com/WordPress/wporg-gp-translation-events/pull/36#issuecomment-1936227058);

Should we sort events based on their start dates or end dates. Sorting by event start dates means only events starting on the current date or later will be included in the list.

The issue with this is that, events that last for several days will not be included once their start date has passed the current date, thereby reducing visibility for such events.

amieiro:
Currently, the event list only supports the upcoming events, not the current events:

image

This PR adds the events that are happening right now: that have started in the past and end in the future. 

This PR does not cache requests.

image

It adds pagination for current and upcoming events (in the next screenshot, I have changed the number of items per page from 10 to 1 to take the screenshot).

image

I put the current events first before the upcoming events because I think it is more important to visualize the current ones than the future ones.

Fixes https://github.com/WordPress/wporg-gp-translation-events/issues/46.

### DIFF
--- a/templates/events-list.php
+++ b/templates/events-list.php
@@ -20,7 +20,7 @@ if ( $current_events_query->have_posts() ) :
 			?>
 			<li class="event-list-item">
 				<span class="event-list-date"><?php echo esc_html( $event_start ); ?></span>
-				<a href="<?php echo esc_url( gp_url( wp_make_link_relative( get_the_permalink() ) ) ) ?>"><?php the_title(); ?></a> by <span><?php the_author(); ?></span>?></span>
+				<a href="<?php echo esc_url( gp_url( wp_make_link_relative( get_the_permalink() ) ) ) ?>"><?php the_title(); ?></a> by <span><?php the_author(); ?></span>
 				<p><?php the_excerpt(); ?></p>
 			</li>
 		<?php
@@ -52,7 +52,7 @@ if ( $upcoming_events_query->have_posts() ) :
 			?>
 			<li class="event-list-item">
 				<span class="event-list-date"><?php echo esc_html( $event_start ); ?></span>
-				<a href="<?php echo esc_url( gp_url( wp_make_link_relative( get_the_permalink() ) ) ) ?>"><?php the_title(); ?></a> by <span><?php the_author(); ?></span>?></span>
+				<a href="<?php echo esc_url( gp_url( wp_make_link_relative( get_the_permalink() ) ) ) ?>"><?php the_title(); ?></a> by <span><?php the_author(); ?></span>
 				<p><?php the_excerpt(); ?></p>
 			</li>
 		<?php
--- a/templates/events-list.php
+++ b/templates/events-list.php
@@ -20,7 +20,7 @@ if ( $current_events_query->have_posts() ) :
 			?>
 			<li class="event-list-item">
 				<span class="event-list-date"><?php echo esc_html( $event_start ); ?></span>
-				<a href="<?php the_permalink(); ?>"><?php the_title(); ?></a> by <span><?php the_author(); ?></span>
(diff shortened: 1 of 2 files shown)


trymebytes:
Nice job! Regarding the placement of the lists, I wonder if it makes for a better user experience if  the list of current/happening events is placed on the right as shown in the mockup below;

Screenshot 2024-02-13 at 10 37 22

amieiro:
> Nice job! Regarding the placement of the lists, I wonder if it makes for a better user experience if the list of current/happening events is placed on the right as shown in the mockup below;
> 
> Screenshot 2024-02-13 at 10 37 22

As it is an important content, I think the best place is the wide column. But, of course, we can iterate it in the future.

templates/events-list.php
```diff
@@ -6,41 +6,75 @@
...
-			$event_start = get_post_meta( get_the_ID(), '_event_start', true );
+		while ( $current_events_query->have_posts() ) :
+			$current_events_query->the_post();
+			$event_start = ( new DateTime( get_post_meta( get_the_ID(), '_event_start', true ) ) )->format('l, F j, Y');
 			?>
 			<li class="event-list-item">
-				<span class="event-list-date"><time class="event-utc-time" datetime="<?php echo esc_attr( $event_start ); ?>"></span>
-				<a href="<?php echo esc_url( gp_url( wp_make_link_relative( get_the_permalink() ) ) ) ?>"><?php the_title(); ?></a> by <span><?php the_author(); ?></span>
+				<span class="event-list-date"><?php echo esc_html( $event_start ); ?></span>
+				<a href="<?php the_permalink(); ?>"><?php the_title(); ?></a> by <span><?php the_author(); ?></span>
```
psrpinto: I'm guessing this change wasn't intended?

```suggestion
				<a href="<?php echo esc_url( gp_url( wp_make_link_relative( get_the_permalink() ) ) ) ?>"><?php the_title(); ?></a> by <span><?php the_author(); ?></span>?></span>
```

templates/events-list.php
```diff
@@ -6,41 +6,75 @@
...
+	<h3>Upcoming events</h3>
+	<ul>
+		<?php
+		while ( $upcoming_events_query->have_posts() ) :
+			$upcoming_events_query->the_post();
+			$event_start = ( new DateTime( get_post_meta( get_the_ID(), '_event_start', true ) ) )->format('l, F j, Y');
+			?>
+			<li class="event-list-item">
+				<span class="event-list-date"><?php echo esc_html( $event_start ); ?></span>
+				<a href="<?php the_permalink(); ?>"><?php the_title(); ?></a> by <span><?php the_author(); ?></span>
```
psrpinto: I guess we should use the glotpress link here.

```suggestion
				<a href="<?php echo esc_url( gp_url( wp_make_link_relative( get_the_permalink() ) ) ) ?>"><?php the_title(); ?></a> by <span><?php the_author(); ?></span>?></span>
```

psrpinto (commented):
Tested and it works 👌

I noticed a couple of things though

templates/events-list.php
psrpinto: I noticed we need the event URL in multiple places, and since it's a long chain of function calls, in the future we could add a function somewhere that takes the event id as argument and returns the URL.

amieiro: Thank you for these updates! Related to these chain of functions, they are in a loop, so we should compute in the `WPORG_GP_Translation_Events_Route` class who calls the current template. But there, in the `events_list` method, we only use 2 `WP_Query` queries to pass them to the template, so we will overengineer the final solution, so for me, it is fine the current approach.

psrpinto (approved):

PR #59, opened 2024-02-12
Align the labels in the creation form

amieiro:
The form for the event creation and update has the labels unaligned.

image

This PR vertically align them and fixes the "Event URL" link.

image

image


psrpinto (approved):
Much better!

PR #60, opened 2024-02-13
Register an user to attend or un-attend to an event

amieiro:
The [Attend or un-attend an event #43](https://github.com/WordPress/wporg-gp-translation-events/pull/43) PR tried to merge into the [filter-style-list](https://github.com/WordPress/wporg-gp-translation-events/tree/filter-style-list) branch, not in the trunk branch, but we merge [the PR](https://github.com/WordPress/wporg-gp-translation-events/pull/36) for this [filter-style-list](https://github.com/WordPress/wporg-gp-translation-events/tree/filter-style-list) branch before we approve the [#43](https://github.com/WordPress/wporg-gp-translation-events/pull/43) PR, so the code was not merged in the trunk branch, and now it has a lot of merge conflicts. 

This PR resolve the merging conflicts and add the same functionality than #43.

### DIFF
--- a/includes/class-wporg-gp-translation-events-route.php
+++ b/includes/class-wporg-gp-translation-events-route.php
@@ -6,7 +6,7 @@
  * @package wporg-gp-translation-events
  */
 class WPORG_GP_Translation_Events_Route extends GP_Route {
-
+	public const USER_META_KEY_ATTENDING = 'translation-events-attending';
 
 	/**
 	 * WPORG_GP_Translation_Events_Route constructor.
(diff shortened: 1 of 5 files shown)


includes/class-wporg-gp-translation-events-route.php
```diff
@@ -106,15 +106,21 @@ public function events_edit( $event_id ) {
...
 
-		$event_title       = $event->post_title;
-		$event_description = $event->post_content;
-		$event_start       = get_post_meta( $event->ID, '_event_start', true ) ?? '';
-		$event_end         = get_post_meta( $event->ID, '_event_end', true ) ?? '';
+		$event_id = $event->ID;
+		$event_title         = $event->post_title;
+		$event_description   = $event->post_content;
+		$event_start         = get_post_meta( $event->ID, '_event_start', true ) ?? '';
+		$event_end           = get_post_meta( $event->ID, '_event_end', true ) ?? '';
```
psrpinto: Should be `?:` instead of `??`.

includes/class-wporg-gp-translation-events-route.php
```diff
@@ -106,15 +106,21 @@ public function events_edit( $event_id ) {
...
+		$event_locale        = get_post_meta( $event->ID, '_event_locale', true ) ?: '';
+		$event_project_name  = get_post_meta( $event->ID, '_event_project_name', true ) ?: '';
```
psrpinto: Project name and locale have been removed now, they're being added back here.

psrpinto (changes requested):
Thanks for fixing the conflicts. I noticed a couple of things that I think need to be fixed.

includes/class-wporg-gp-translation-events-route.php
amieiro: I think it is correct. This is the [Null coalescing operator](https://www.php.net/manual/en/migration70.new-features.php#migration70.new-features.null-coalesce-op). If `get_post_meta( $event->ID, '_event_start', true )` returns a valid value, `$event_start = get_post_meta( $event->ID, '_event_start', true )`. If it returns `false`, then  `$event_start = ''`.

includes/class-wporg-gp-translation-events-route.php
amieiro: Yes, we don't need it. Removed [here](https://github.com/WordPress/wporg-gp-translation-events/pull/60/commits/2c027c424cc911dfa5b86c4cb03aadaf3ea31bac). Thank you!

includes/class-wporg-gp-translation-events-route.php
psrpinto: Ok, I see. We changed this from `??` to `?:` because of [this remark](https://github.com/WordPress/wporg-gp-translation-events/pull/43#discussion_r1484308260), but now I'm not sure anymore. Should we change the other conditions below to also use `??`?

trymebytes (approved):
Looks good to me!

psrpinto (approved):

includes/class-wporg-gp-translation-events-route.php
amieiro: Yes, you are right. The `get_post_meta( something, something, true )` can return:
- The value of the metadata field.
- A false value.
- An empty string.

So the correct usage is `?:`. Removed [here](https://github.com/WordPress/wporg-gp-translation-events/pull/60/commits/9e80788583af709404001977a8e1920dd7c33ced). You can test it easily with this code:

```
<?php 
echo 1 ?? 'a';
echo false ?? 'b';
echo '' ?? 'c';

echo 1 ?: 'd';
echo false ?: 'e';
echo '' ?: 'f';
```

PR #64, opened 2024-02-13
Hide the editing options in wp-admin

### LINKED ISSUE #40: Should events be createable/editeable through wp-admin?
We briefly discussed this before but I'm not sure we reached a conclusion, so I'm opening an issue.

Currently events can be created or edited through either the custom pages in this plugin, or through wp-admin, as any other post. However, AFAICT, currently some things are only done when the event is edited through this plugin.

Examples are validating that the end is after the start, or invalidating the active events cache.

I see two options here:

1. Prevent events from being created/edited through wp-admin. Not sure that's a very WordPress-y thing to do.
2. Fully support creating/editing events through wp-admin (e.g. validate end is after start, invalidate cache, etc)

akirk:
Fixes #40.


amieiro (approved):
Works fine!

PR #68, opened 2024-02-13
Add 'My Events' page

### LINKED ISSUE #57: My Events page
This page shows all the events created by the user, including drafts with edit links.

image

trymebytes:
Fixes #57 

**Mockup**
image

**Updated screenshot**
image

### DIFF
--- a/templates/events-user-created.php
+++ b/templates/events-user-created.php
@@ -21,7 +21,7 @@ if ( $query->have_posts() ) :
 			?>
 			<li class="event-list-item">
 				<span class="event-list-date"><time class="event-utc-time" datetime="<?php echo esc_attr( $event_start ); ?>"></span>
-				<a href="<?php echo esc_url( gp_url( wp_make_link_relative( $permalink ) ) ); ?>"><?php the_title(); ?></a> by <span><?php the_author(); ?></span>
+				<a href="<?php echo esc_url( gp_url( wp_make_link_relative( $permalink ) ) ); ?>"><?php the_title(); ?></a>
 				<p><?php the_excerpt(); ?></p>
 			</li>
 			<?php
--- a/templates/events-user-created.php
+++ b/templates/events-user-created.php
@@ -20,11 +20,13 @@ if ( $query->have_posts() ) :
 			$permalink                     = str_replace( '%pagename%', $post_name, $permalink );
 			?>
 			<li class="event-list-item">
-				<span class="event-list-date"><time class="event-utc-time" datetime="<?php echo esc_attr( $event_start ); ?>"></span>
+				<div>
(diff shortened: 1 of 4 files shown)


wporg-gp-translation-events.php
```diff
@@ -65,7 +65,7 @@ function register_event_post_type() {
 		'menu_icon'   => 'dashicons-calendar',
 		'supports'    => array( 'title', 'editor', 'thumbnail', 'revisions' ),
 		'rewrite'     => array( 'slug' => $slug ),
-		'show_ui'     => false,
+		'show_ui'     => true,
```
akirk: Since you needed to activate that, should we just use a filter here so that you can override it with another plugin?

templates/events-user-created.php
```diff
@@ -0,0 +1,48 @@
+<?php
+gp_title( __( 'Translation Events' ) . ' - ' . esc_html( 'My Events' ) );
```
akirk:
```suggestion
gp_title( __( 'Translation Events' ) . ' - ' . esc_html__( 'My Events' ) );
```

akirk:
This feels a little visually overloaded for me with two boxes, one of them square and the other round. Could we try something as simple as this?

Screenshot 2024-02-14 at 15 07 08

amieiro:
You need to remove the link in the title if the event is in draft status.

image

akirk:
I think the link should be there so that you can view the event page before publishing it, see #75.

amieiro:
If we want to maintain the link to be able to see the event before publishing it, I will suggest a small tag only for the events in `draft` status. 

image

This is the same approach we are using with the inactive projects in GlotPress:

image

akirk:
That would work for me, feel free to experiment.

trymebytes:
I have added the flag back and it now looks like this;

image

amieiro:
I have changed the `'posts_per_page' => 10,` to `'posts_per_page' => 2,` [here](https://github.com/WordPress/wporg-gp-translation-events/pull/68/files#diff-98f145e1ea9a704b3a94b4e4152a72a5a2a5f677c9dee3d80ffc4a5644f20457R214) to have pagination, and it doesn't work. You have some pagination code in the events' main page.

image

trymebytes:
I have fixed this, please pull and try again.

templates/events-user-created.php
```diff
@@ -0,0 +1,58 @@
...
+			'total'     => $query->max_num_pages,
+			'current'   => max( 1, get_query_var( 'page' ) ),
+			'prev_text' => '&laquo; Previous',
+			'next_text' => 'Next &raquo;',
+			'format'    => '?page=%#%',
+
+
+			// 'current'   => max( 1, $current_events_query->query_vars['current_events_paged'] ),
+			// 'prev_text' => '&laquo; Previous',
+			// 'next_text' => 'Next &raquo;',
```
amieiro: Please, remove these comments. 

```suggestion
```

amieiro (approved):
It works fine for me.

PR #69, opened 2024-02-14
Add the delete button in the edit form of an event

### LINKED ISSUE #66: Add a Delete link to the Edit screen
- You are only allowed to delete an event if it doesn't have stats.
- If the event is active, the cache needs to be cleared.
- Deleting means to change the post status to trashed.

amieiro:
This PR adds the "Delete button"

image

Fixes https://github.com/WordPress/wporg-gp-translation-events/issues/66

### DIFF
--- a/assets/js/translation-events.js
+++ b/assets/js/translation-events.js
@@ -50,7 +50,29 @@ jQuery(document).ready(function($) {
         });
     });
 
-    function validateEventDates() {
+		$('.delete-event').on('click', function(e) {
+			console.log('delete event');
+			e.preventDefault();
+			if ( ! confirm( 'Are you sure you want to delete this event?' ) ) {
+				return;
+			}
+			var $form = $('.translation-event-form');
+			$('#event-form-action').val('delete');
+			$.ajax({
+				type: 'POST',
+				url: $translation_event.url,
+				data:$form.serialize(),
+				success: function(response) {
+					if ( response.data.success ) {
+						window.location.href = response.data.redirectUrl;
+					}
+				},
+				error: function(error) {
+					$gp.notices.error(response.data.message);
+				}
+			});
+		});
+	function validateEventDates() {
         var startDateTimeInput = $('#event-start');
         var endDateTimeInput = $('#event-end');
         if ( ! startDateTimeInput.length || ! endDateTimeInput.length ) {
@@ -65,7 +87,7 @@ jQuery(document).ready(function($) {
         });
     }
     function selectUserTimezone() {
-        document.querySelector(`#event-timezone option[value="${Intl.DateTimeFormat().resolvedOptions().timeZone}"]`).selected = true 
+        document.querySelector(`#event-timezone option[value="${Intl.DateTimeFormat().resolvedOptions().timeZone}"]`).selected = true
     }
 
     function convertToUserLocalTime() {
@@ -80,11 +102,11 @@ jQuery(document).ready(function($) {
             var userLocalDateTime = new Date(eventDateObj.getTime() - userTimezoneOffsetMs);
 
             var options = { weekday: 'short', year: 'numeric', month: 'long', day: 'numeric', hour: 'numeric', minute: 'numeric', hour12: true, timeZoneName: 'short' };
-    
+
             timeEl.textContent = userLocalDateTime.toLocaleString('en-US', options);
         });
     }
-    
+
 });
 }( jQuery, $gp )
 );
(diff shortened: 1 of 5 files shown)


trymebytes:
Just a personal observation, perhaps we can change the position of the delete button because having 3 buttons on the same row may be a bit confusing.
Screenshot 2024-02-15 at 11 23 34

includes/class-wporg-gp-translation-events-route.php
```diff
@@ -123,6 +126,22 @@ public function events_edit( $event_id ) {
 		$event_status                  = $event->post_status;
 		list( $permalink, $post_name ) = get_sample_permalink( $event_id );
 		$permalink                     = str_replace( '%pagename%', $post_name, $permalink );
+		$show_delete_button = false;
+		if ('Edit Event' === $event_form_title) {
+			$event = get_post( $event_id );
```
trymebytes: We can remove this if condition since we are already in the `events_edit` function. Also, `$event` has been assigned a value above so it's okay to remove this duplicate.

assets/js/translation-events.js
```diff
@@ -50,7 +50,29 @@ jQuery(document).ready(function($) {
...
+			var $form = $('.translation-event-form');
+			$('#form-name').val('delete_event');
+			$('#event-form-action').val( 'delete' );
+			$.ajax({
+				type: 'POST',
+				url: $translation_event.url,
+				data:$form.serialize(),
+				success: function(response) {
+					if ( response.data.eventId ) {
+						history.back();
```
trymebytes: After creating an event,  and I'm directed to the edit page, the `Delete` button doesn't show until the edit page is reloaded. We could try to add it in JS.

trymebytes (dismissed):

trymebytes:
In a follow up PR, I'll suggest we have a single AJAX function handle all event form submissions.

includes/class-wporg-gp-translation-events-route.php
```diff
@@ -123,6 +126,22 @@ public function events_edit( $event_id ) {
...
+			$stats_calculator = new WPORG_GP_Translation_Events_Stats_Calculator();
+			try {
+				$event_stats = $stats_calculator->for_event($event);
+			} catch ( Exception $e ) {
+				$this->die_with_error( 'Failed to calculate event stats' );
+			}
+			if ( empty( $event_stats->rows() ) ) {
```
psrpinto: We could add a function to `Stats_Calculator` that tells whether the event has stats. I suspect we'll need to know whether an event has stats in other places too. Something like the following:

```php
public function event_has_stats( WP_Post $event ): bool {
	try {
		$stats = $this->for_event( $event );
	} catch ( Exception $e ) {
		return false;
	}

	return ! empty( $stats->rows() );
}
```

wporg-gp-translation-events.php
```diff
@@ -213,18 +213,40 @@ function submit_event_ajax() {
 		);
 		$response_message = 'Event updated successfully!';
 	}
+	if ( 'delete_event' === $_POST['form_name'] ) {
```
psrpinto: I think we should use `$action` instead of `$_POST['form_name']`. `$action` has been sanitized.

amieiro:
> Just a personal observation, perhaps we can change the position of the delete button because having 3 buttons on the same row may be a bit confusing. Screenshot 2024-02-15 at 11 23 34

I like to have all the buttons in the same row, with different styles (one specific for deleting), rather than having them in 2 rows.

image

includes/class-wporg-gp-translation-events-route.php
amieiro: Done!

assets/js/translation-events.js
amieiro: I have solved this problem in [this commit](https://github.com/WordPress/wporg-gp-translation-events/pull/69/commits/22ed7167dc41a7ea7285d00dd17afc1fdeef3f95).

includes/class-wporg-gp-translation-events-route.php
amieiro: Done in [Add a method to know if an event has stats](https://github.com/WordPress/wporg-gp-translation-events/pull/69/commits/1d2e22123732eabf58a04a98c9dd192c2d54903b). Good improvement.

wporg-gp-translation-events.php
amieiro: Done [here](https://github.com/WordPress/wporg-gp-translation-events/pull/69/commits/9fd3d6282b5160da2aa4d38d8ea34a31a127a377). Thank you.

psrpinto (approved):
👍

PR #77, opened 2024-02-14
Lint main plugin file

psrpinto:
This PR applies linting to the main plugin file. I would suggest reviewing commit by commit, they should be self-explanatory. I did small refactors where it made sense.


akirk (approved):
Seems all reasonable to me! Tested and it still works :)

PR #78, opened 2024-02-14
Linting 2

psrpinto:
Another batch of linting. There will be one last after this one.

This should not interfere with currently-open PRs.

### DIFF
--- a/includes/class-wporg-gp-translation-events-stats-calculator.php
+++ b/includes/class-wporg-gp-translation-events-stats-calculator.php
@@ -62,8 +62,13 @@ class WPORG_GP_Translation_Events_Stats_Calculator {
 		$stats = new WPORG_GP_Translation_Events_Event_Stats();
 		global $wpdb;
 
-		$query = $wpdb->prepare(
-			"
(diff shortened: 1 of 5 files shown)


includes/class-wporg-gp-translation-events-stats-calculator.php
```diff
@@ -51,33 +53,40 @@ public function totals(): WPORG_GP_Translation_Events_Stats_Row {
...
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.SchemaChange
+		$rows = $wpdb->get_results(
+			$wpdb->prepare(
+				"
 				select locale,
 					   sum(action = 'create') as created,
 					   count(*) as total,
 					   count(distinct user_id) as users
-				from $this->ACTIONS_TABLE_NAME
+				from wp_wporg_gp_translation_events_actions
```
amieiro, line 76: Maybe in the next iteration we should add the database prefix to the table, avoiding to hardcode its name.

```suggestion
				from {$wpdb->prefix}_gp_translation_events_actions
```

amieiro (approved):
I think it is correct. We can continue to iterate and improve it in future PRs.

PR #79, opened 2024-02-15
Show prompt only when creating new events

### LINKED ISSUE #70: Updating an event should not show a confirmation dialog
Right now this dialog is also shown when updating an event, not just first publishing it:

Screenshot 2024-02-14 at 11 27 29

trymebytes:
Fixes #70


psrpinto (approved):
Tested and it works 👌

amieiro:
If you save the draft, and then you publish the event, you get a prompt. We can address this issue in another PR.

PR #80, opened 2024-02-15
Show a bubble when the event is in draft status

### LINKED ISSUE #75: Draft Event Page
When viewing a Draft event page, it needs to be visible that this is just a draft yet.

Maybe: only allow the author and admins view it?

amieiro:
This PR adds a filter to generate the event slug when it is stored as draft and a bubble when the user sees it, to indicate that is a draft.

image

Fixes https://github.com/WordPress/wporg-gp-translation-events/issues/75.


psrpinto (approved):
Looks great!

PR #83, opened 2024-02-15
Lint routes and templates

psrpinto:
This is the final PHP linting PR. I would suggest reviewing commit by commit, as each commit should be self-explanatory. I've made small refactors where appropriate.

### DIFF
--- a/includes/class-wporg-gp-translation-events-route.php
+++ b/includes/class-wporg-gp-translation-events-route.php
@@ -14,7 +14,8 @@ class WPORG_GP_Translation_Events_Route extends GP_Route {
 	 * @since 0.0.1
 	 */
 	public function __construct() {
-		$this->template_path = dirname( __FILE__ ) . '/../templates/';
+		parent::__construct();
+		$this->template_path = __DIR__ . '/../templates/';
 	}
 
 	/**
@@ -33,6 +34,7 @@ class WPORG_GP_Translation_Events_Route extends GP_Route {
 			'current_events_paged' => $_current_events_paged,
 			'paged'                => $_current_events_paged,
 			'post_status'          => 'publish',
+			// phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
 			'meta_query'           => array(
 				array(
 					'key'     => '_event_start',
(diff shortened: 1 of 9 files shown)


includes/class-wporg-gp-translation-events-route.php
```diff
@@ -117,12 +144,21 @@ public function events_edit( $event_id ) {
...
-		$event_start                   = self::convertToTimezone( get_post_meta( $event_id, '_event_start', true ), $event_timezone ) ?? '';
-		$event_end                     = self::convertToTimezone( get_post_meta( $event_id, '_event_end', true ), $event_timezone ) ?? '';
 		$event_status                  = $event->post_status;
 		list( $permalink, $post_name ) = get_sample_permalink( $event_id );
 		$permalink                     = str_replace( '%pagename%', $post_name, $permalink );
+		$event_url                     = get_site_url() . gp_url( wp_make_link_relative( $permalink ) );
+		$event_timezone                = get_post_meta( $event_id, '_event_timezone', true ) ?: '';
+
+		try {
+			$event_start = self::convertToTimezone( get_post_meta( $event_id, '_event_start', true ), $event_timezone ) ?? '';
```
akirk: Why the  `?? ''`, looks like convertToTimezone cannot return null.

templates/events-list.php
```diff
@@ -16,26 +21,29 @@
...
-			'next_text' => 'Next &raquo;',
+	echo wp_kses_post(
+		paginate_links(
+			array(
+				'total'     => $current_events_query->max_num_pages,
+				'current'   => max( 1, $current_events_query->query_vars['current_events_paged'] ),
+				'format'    => '?current_events_paged=%#%',
+				'prev_text' => '&laquo; Previous',
+				'next_text' => 'Next &raquo;',
+			)
```
akirk, line 46: Could you add a `?? ''` to both `paginate_links()` inside this PR to avoid the error  `Deprecated: preg_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated in /wp-includes/kses.php on line 1744`

tests/unit/dummy.php
```diff
@@ -1,7 +1,7 @@
 <?php
-
-// TODO: Remove this file once there are actual tests.
-
+/**
```
akirk, line 2: Should this be this?

```suggestion
/*
```

akirk (commented):
Overall looks good, could you just pick up that `paginate_links` fix and judge my other remarks, thanks!

tests/unit/dummy.php
psrpinto: The coding standards don't allow it somehow. It's viewed as the class comment so must start with `/**`. This will be removed soon once we add tests.

templates/events-list.php
psrpinto: Oops, good catch. I just now saw the deprecation message in the the events list page. Fixed.

akirk (approved):
Thanks, good to merge!

amieiro (approved):
I think it is ok to be merged.

PR #74, opened 2024-02-14
Improve UI

trymebytes:
Fixes #65 and the overall look and feel of other pages.

Screenshots
**Before**
image

**After**
improveui

PR #85, opened 2024-02-15
Lint JS and CSS

psrpinto:
This is the final linting PR. It's mostly just fixing whitespace. In the js, I refactored a little to extract functions out of functions.

This will likely cause conflicts with #74. I can help resolve conflicts there, or it if gets merged, I will resolve conflicts here.

[Diff ignoring whitespace changes](https://github.com/WordPress/wporg-gp-translation-events/pull/85/files?diff=unified&w=1)

### DIFF
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -5,8 +5,8 @@ on:
     types: [ opened, synchronize ]
 
 jobs:
-  php:
-    name: PHP
+  phpcs:
+    name: phpcs
     runs-on: ubuntu-latest
 
     steps:
--- a/assets/css/translation-events.css
+++ b/assets/css/translation-events.css
@@ -1,136 +1,135 @@
 .translation-event-form label {
-    display: inline-block;
-    width: 140px;
-    vertical-align: top;
+	display: inline-block;
+	width: 140px;
+	vertical-align: top;
 }
 .translation-event-form #event-title,
 .translation-event-form #event-description {
-    width: 30%;
+	width: 30%;
 }
 .translation-event-form input[type="text"],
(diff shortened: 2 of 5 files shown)


akirk (approved):
Looks good! I am just confused regarding the tooling, shall we add `wp-scripts` for this, like in https://github.com/GlotPress/gp-translation-helpers/blob/main/composer.json ?

psrpinto:
> Looks good! I am just confused regarding the tooling, shall we add wp-scripts for this, like in https://github.com/GlotPress/gp-translation-helpers/blob/main/composer.json ?

(I'm assuming you meant [package.json](https://github.com/GlotPress/gp-translation-helpers/blob/main/package.json#L17-L18) instead of `composer.json`)

It didn't occur to me to use `wp-scripts`, good point. I'm willing to make the change to use `wp-scripts` instead of `phpcs`, if you think it makes sense.

Currently we don't yet have any npm/node dependencies, so we could consider sticking with `phpcs` for now. (Would save 384MB in the `node_modules/` 😅 ).

psrpinto:
I rebased with trunk to fix conflicts.

akirk:
> (I'm assuming you meant [package.json](https://github.com/GlotPress/gp-translation-helpers/blob/main/package.json#L17-L18) instead of composer.json)

Oops, indeed.

> Currently we don't yet have any npm/node dependencies, so we could consider sticking with phpcs for now.

I wasn't aware that phpcs can do this reformatting. I guess it's ok for the moment :-)

psrpinto:
Ok, cool. I think it's a good time to merge this one now because there's only #74 which will be affected. Merging.

PR #86, opened 2024-02-16
Fix PHP Notice:  Undefined variable: attending_event_ids

trymebytes:
Stats for new events were not showing and the stats for old events were not updated after making a translation and PHP threw this notice;
`PHP Notice:  Undefined variable: attending_event_ids`. In this PR, the missing `$attending_event_ids` variable is added.

### DIFF
--- a/includes/class-wporg-gp-translation-events-translation-listener.php
+++ b/includes/class-wporg-gp-translation-events-translation-listener.php
@@ -154,6 +154,8 @@ class WPORG_GP_Translation_Events_Translation_Listener {
 	 * @param WPORG_GP_Translation_Events_Event[] $events Events.
 	 * @return WPORG_GP_Translation_Events_Event[]
 	 */
+	// phpcs:disable Generic.CodeAnalysis.UnusedFunctionParameter.FoundAfterLastUsed
+	// phpcs:disable Generic.CodeAnalysis.UnusedFunctionParameter.Found
 	private function select_events_user_is_registered_for( array $events, int $user_id ): array {
 		$attending_events = get_user_meta( $user_id, 'translation-events-attending', true );
 		return array_filter(
--- a/includes/class-wporg-gp-translation-events-translation-listener.php
+++ b/includes/class-wporg-gp-translation-events-translation-listener.php
@@ -154,13 +154,12 @@ class WPORG_GP_Translation_Events_Translation_Listener {
 	 * @param WPORG_GP_Translation_Events_Event[] $events Events.
 	 * @return WPORG_GP_Translation_Events_Event[]
 	 */
-	// phpcs:disable Generic.CodeAnalysis.UnusedFunctionParameter.FoundAfterLastUsed
-	// phpcs:disable Generic.CodeAnalysis.UnusedFunctionParameter.Found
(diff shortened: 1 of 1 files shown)


includes/class-wporg-gp-translation-events-translation-listener.php
```diff
@@ -157,10 +157,11 @@ function ( $event ) use ( $at ) {
 	// phpcs:disable Generic.CodeAnalysis.UnusedFunctionParameter.FoundAfterLastUsed
 	// phpcs:disable Generic.CodeAnalysis.UnusedFunctionParameter.Found
 	private function select_events_user_is_registered_for( array $events, int $user_id ): array {
+		$attending_events = get_user_meta( $user_id, 'translation-events-attending', true );
 		return array_filter(
 			$events,
-			function ( $event ) use ( $attending_event_ids ) {
-				return isset( $attending_event_ids[ $event->ID ] );
+			function ( $event ) use ( $attending_events ) {
```
akirk: Is the `$events` variable passed in from the outside not what we are looking for?
```suggestion
			function ( $event ) use ( $events ) {
```

trymebytes: The `$events` passed from outside is the array of [active events](https://github.com/WordPress/wporg-gp-translation-events/blob/9ae7586b67d83c02a79958d27424861f91ff678d/includes/class-wporg-gp-translation-events-translation-listener.php#L62). We then want to filter this array to ensure we have only the events that the user is attending.

akirk: Ok, I see. Could we then please use the `USER_META_KEY_ATTENDING` constant like it's used here? https://github.com/WordPress/wporg-gp-translation-events/blob/d21f915b7f3a0237fd5f444615606690057e1f63/includes/class-wporg-gp-translation-events-route.php#L184

akirk (approved):
Thanks!

psrpinto (approved):
Oops, thanks for fixing!

includes/class-wporg-gp-translation-events-translation-listener.php
```diff
@@ -157,10 +157,11 @@ function ( $event ) use ( $at ) {
 	// phpcs:disable Generic.CodeAnalysis.UnusedFunctionParameter.FoundAfterLastUsed
 	// phpcs:disable Generic.CodeAnalysis.UnusedFunctionParameter.Found
 	private function select_events_user_is_registered_for( array $events, int $user_id ): array {
+		$attending_event_ids = get_user_meta( $user_id, WPORG_GP_Translation_Events_Route::USER_META_KEY_ATTENDING, true );
 		return array_filter(
 			$events,
 			function ( $event ) use ( $attending_event_ids ) {
```
akirk: @amieiro in #60 you removed the `WPORG_GP_Translation_Events_Event` type hint. @trymebytes you added the `id()` below. So this means this is a `WPORG_GP_Translation_Events_Event` after all and we could add the type hint back?

akirk (dismissed):
Just to discuss https://github.com/WordPress/wporg-gp-translation-events/pull/86/files#r1492309532

psrpinto (approved):
🚢

PR #90, opened 2024-02-16
Introduce PHP namespace

### LINKED ISSUE #14: Some code is not namespaced to the plugin
I guess functions/variables/etc declared on the global scope need to be namespaced? For example, `submit_event_ajax()`.

Also, should the `post_type` be namespaced too? Currently it's `event`.

psrpinto:
Fixes #14 

I think namespaces make code considerably easier to understand because it allows shortening the names of classes, functions, variables and files. And of course it prevents clashes of names with other plugins.


akirk (dismissed):
I didn't investigate why but in this PR `/events/` doesn't load for me anymore.

psrpinto:
Mmm, weird, I tested locally and `/events/` and other routes seem to work for me. Could you check if there's something in the logs?

trymebytes:
Is there a reason for renaming the main plugin file to `plugin.php`? Otherwise, we could just leave it unchanged for the sake of consistency.

trymebytes:
The name change caused the plugin to be deactivated and would need to be re-activated. I guess that's why the routes don't load for @akirk .

psrpinto:
Ah yes, indeed, I forgot to mention that the plugin needs to be reactivated 🤦‍♂️ 

The reason I renamed the main plugin file is because the directory/repo already "carries" the name of the plugin. I did it for the purpose of shortening the name of the file.

I'm happy to revert this change though, if you want.

akirk:
Ah makes sense! Thanks @trymebytes !

akirk:
In WordPress it's the common practice to have the main plugin file with the same name as the directory+.php

akirk (changes requested):
You missed a rename in https://github.com/WordPress/wporg-gp-translation-events/pull/90/files#diff-4ad28b7297b89b696671daf3dbbda0f7e559703113a18bc042b3361ebbb04caaR189 `includes/route.php:189`and in `wporg-gp-translation-events.php:234`. If I rename those locally, it works!

psrpinto:
Good catch, fixed. Also merged trunk to resolve conflicts.

akirk (changes requested):
Unfortunately i just got another crash: `Fatal error: Uncaught Error: Class "Wporg\TranslationEvents\DateTime" not found in wp-content/plugins/wporg-gp-translation-events/templates/events-user-created.php:31`

psrpinto:
Oops, sorry, should've tested. I added the missing import, and tested and everything seems to work.

akirk (approved):
Looks good now!

PR #93, opened 2024-02-16
Add the date to "My events" and update the date order

### LINKED ISSUE #92: Order "My events" correctly, adding the date
Currently, "My events" doesn't have the events ordered correctly. I propose to order descending, so the old events will appear in the last places of the list.

amieiro:
Currently, "My events" doesn't have the date of each event and the order is not working properly.

image

This PR add the date of each event and order them descending, so the old events will appear in the last places of the list. It adds the footer to the template.

image

Fixes https://github.com/WordPress/wporg-gp-translation-events/issues/92.


psrpinto (approved):
👍

PR #99, opened 2024-02-19
Tests for stats listener

### LINKED ISSUE #15: Add unit tests for stats
To make sure stats are calculated correctly.

psrpinto:
This PR adds tests for the `Stats_Listener` (which "listens" to actions from GlotPress).

This took me longer than expected because it turns out that the test setup wasn't yet fully done. I looked at how https://github.com/GlotPress/gp-translation-helpers does it but since that repo doesn't use `wp-env`,  it didn't directly apply to here, so I had to figure some things out.

# TODO

- [x] Make tests that need GlotPress run on wp-env
- [x] Make tests that need GlotPress run on CI
- [x] Write tests for `Stats_Listener`


tests/bootstrap.php
```diff
@@ -23,10 +31,26 @@
...
+// Require GlotPress test code.
+require_once _glotpress_path( '/tests/phpunit/lib/testcase.php' );
+require_once _glotpress_path( '/tests/phpunit/lib/testcase-route.php' );
+require_once _glotpress_path( '/tests/phpunit/lib/testcase-request.php' );
+
+function _install_glotpress() {
+	require_once ABSPATH . 'wp-admin/includes/upgrade.php';
+	require_once _glotpress_path( '/gp-includes/schema.php' );
+	require_once _glotpress_path( '/gp-includes/install-upgrade.php' );
+	gp_upgrade_db();
```
akirk: Not sure I'd call this a suggested change. Just to show how one could make it work.
```suggestion
	gp_upgrade_db();
	foreach ( explode( ';', file_get_contents( __DIR__ . '/../schema.sql') ) as $sql ) {
		if ( trim($sql) ) {
			$wpdb->query( $sql );
		}
	}

```

akirk (changes requested):
The tests work well and pass, great job there! But I am not a fan of the local `wp-env` dependency. If I add what I suggested in my comment it can run without wp-env and by just running `./vendor/bin/phpunit`

Can we find a way that allows for both?

tests/bootstrap.php
psrpinto: Yep, agree. My intention was that it would work in all environments indeed (local, wp-env, CI), but I didn't consider that the table wasn't being created for local. I agree with the solution you proposed of creating the table in `bootstrap.php`, which will work for all environments, and is also consistent with what we're doing for glotpress (the glotpress tables get created in `bootstrap.php` too).

Will fix.

akirk (approved):
Thanks for the improvement!

PR #103, opened 2024-02-19
Fix a problem when you publish an event

### LINKED ISSUE #94: Editing published event will put it back to draft
Something in the logic isn't right here, a published event should stay public.

amieiro:
When a user creates or updates an event, it was set as draft. 

This PR updates some JavaScript to solve this problem. To test it, you can:
- Publish a new event.
- Publish a published event.
- Publish a drafted event. 

Fixes https://github.com/WordPress/wporg-gp-translation-events/issues/94

### DIFF
--- a/assets/js/translation-events.js
+++ b/assets/js/translation-events.js
@@ -14,7 +14,8 @@
 					'click',
 					function ( e ) {
 						e.preventDefault();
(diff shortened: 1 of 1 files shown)


akirk (approved):
Confirmed that it works now!

PR #104, opened 2024-02-19
Make links in description body clickable

### LINKED ISSUE #76: Edit Event Improvements
- [x] Allow storing HTML so that people can add links and images. From a PHP perspective, change `sanitize_text_field( $_POST['event_description'] )` to `make_clickable( $_POST['event_description'] )` so that URLs get autolinked. `wp_insert_post` will ensure that the `post_content` is sanitized.

trymebytes:
Fixes #76

### DIFF
--- a/templates/event.php
+++ b/templates/event.php
@@ -40,7 +40,7 @@ gp_tmpl_load( 'events-header', get_defined_vars(), __DIR__ );
 	<div class="event-details-left">
 		<div class="event-page-content">
 			<?php
-				echo wp_kses_post( nl2br( make_clickable( $event_description ) ) );
+				echo wp_kses_post( wpautop( make_clickable( $event_description ) ) );
 			?>
 		</div>
 		<?php if ( ! empty( $event_stats->rows() ) ) : ?>
--- a/templates/event.php
+++ b/templates/event.php
@@ -40,8 +40,7 @@ gp_tmpl_load( 'events-header', get_defined_vars(), __DIR__ );
 	<div class="event-details-left">
 		<div class="event-page-content">
 			<?php
-			// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
-			echo make_clickable( wp_kses_post( $event_description ) );
+				echo wp_kses_post( nl2br( make_clickable( $event_description ) ) );
 			?>
 		</div>
 		<?php if ( ! empty( $event_stats->rows() ) ) : ?>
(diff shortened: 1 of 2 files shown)


amieiro:
I am testing with 3 items:
1. A link.
2. An `<a>` item.
3. An `` item.

image

Only works the first item.

image

I don't know if it works correctly or not.

trymebytes:
Nice catch, thank you!

templates/event.php
```diff
@@ -34,7 +34,10 @@
 	</div>
 	<div class="event-details-left">
 		<div class="event-page-content">
-			<?php echo esc_html( $event_description ); ?>
+			<?php
+			// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+			echo make_clickable( wp_kses_post( $event_description ) );
```
akirk: Why not this which would not require the override?
```suggestion
			echo wp_kses_post( make_clickable( $event_description ) );
```

wporg-gp-translation-events.php
```diff
@@ -163,8 +163,10 @@ function submit_event_ajax() {
 		wp_send_json_error( 'Nonce verification failed' );
 	}
 
-	$title          = isset( $_POST['event_title'] ) ? sanitize_text_field( wp_unslash( $_POST['event_title'] ) ) : '';
-	$description    = isset( $_POST['event_description'] ) ? sanitize_text_field( wp_unslash( $_POST['event_description'] ) ) : '';
+	$title = isset( $_POST['event_title'] ) ? sanitize_text_field( wp_unslash( $_POST['event_title'] ) ) : '';
+	//@codingStandardsIgnoreStart
```
akirk: Is it an option to call `make_clickable()` already here? And why this style of disabling coding standards vs a `phpcs:ignore` style comment?

akirk (changes requested):
I am not sold on the output escaping, maybe we can improve it

wporg-gp-translation-events.php
trymebytes: After some tests, we could decide to call `make_clickable()` here or in the frontend it gives the same result. Do you suggest we rather call it here?

templates/event.php
```diff
@@ -34,7 +34,9 @@
 	</div>
 	<div class="event-details-left">
 		<div class="event-page-content">
-			<?php echo esc_html( $event_description ); ?>
+			<?php
+				echo wp_kses_post( nl2br( make_clickable( $event_description ) ) );
```
akirk: Does this also work?
```suggestion
				echo wp_kses_post( wpautop( make_clickable( $event_description ) ) );
```

akirk: I tested and I think we should use this.

akirk (approved):
Ok for me now!

PR #105, opened 2024-02-19
When a user creates an event, she should be attending right away

### LINKED ISSUE #88: When you create an event, you should be attending right away

amieiro:
When a user creates an event, she should be attending right away. 

To test this PR:
- Create a new event, clicking in the "Publish event" button just after filling the text fields. You should be attending to the event.
- Create a new event, clicking in the "Save Draft" button just after filling the text fields. You should not be attending to the event. Update some text field and click in the "Publish event" button. Now you should be attending to the event.

Fixes https://github.com/WordPress/wporg-gp-translation-events/issues/88.

### DIFF
--- a/wporg-gp-translation-events.php
+++ b/wporg-gp-translation-events.php
@@ -342,8 +342,6 @@ add_action( 'add_meta_boxes', 'Wporg\TranslationEvents\event_meta_boxes' );
 add_action( 'save_post', 'Wporg\TranslationEvents\save_event_meta_boxes' );
 add_action( 'transition_post_status', 'Wporg\TranslationEvents\event_status_transition', 10, 3 );
 
-
-
 /**
  * Add the events link to the GlotPress main menu.
  *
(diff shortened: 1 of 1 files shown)


wporg-gp-translation-events.php
```diff
@@ -222,6 +227,18 @@ function submit_event_ajax() {
 		);
 		$response_message = 'Event updated successfully!';
 	}
+	// The user who creates the event will assist to it when it's published.
+	if ( ( 'create_event' === $action && 'publish' === $event_status ) || ( 'edit_event' === $action && 'publish' === $event_status && 'draft' === $previous_status ) ) {
```
akirk: Would it be easier if we used the https://developer.wordpress.org/reference/hooks/transition_post_status/ hook?

amieiro: Done.

psrpinto (approved):
Tested 👌

PR #110, opened 2024-02-19
Remove the "Attend Event" if the event has passed

### LINKED ISSUE #53: It should not be possible to attend events that have ended
The "Attend Event" button should probably be hidden if the event ended before "now". If the user viewing the page attended the event, "You attended" could be shown instead.

amieiro:
Currently, it is possible to attend or to stop attending to a past event.

image

image

This PR removes the "Attend Event" and the "You're attending" buttons if the event has passed. It adds a bubble if the user has assisted in a past event.

image

image

Fixes https://github.com/WordPress/wporg-gp-translation-events/issues/53.

### DIFF
--- a/assets/css/translation-events.css
+++ b/assets/css/translation-events.css
@@ -121,7 +121,7 @@ h2.event_page_title {
 a.event-link-draft {
 	color:#80807f;
 }
-span.event-label-draft {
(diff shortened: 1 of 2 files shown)


psrpinto (approved):
👍

PR #111, opened 2024-02-19
Modify how date is displayed on the event list

### LINKED ISSUE #84: Current events should display event end date/time
- [x] Show how long the event still lasts, absolute when longer than 1 day, relative when less 
- [x] Also reformat the time to show it below,
- [x] remove the username.

Mockup:
Screenshot 2024-02-19 at 12-17-12 Translation Events

trymebytes:
Fixes #84 

**Before**
image

**After**
image

### DIFF
--- a/includes/event.php
+++ b/includes/event.php
@@ -52,4 +52,26 @@ class Event {
 	public function timezone(): DateTimeZone {
 		return $this->timezone;
 	}
+
+	/**
+	 * Generate text for the end date.
+	 *
+	 * @param string $event_end The end date.
+	 *
+	 * @return string The end date text.
+	 */
+	public static function get_end_date_text( string $event_end ): string {
+		$end_date_time     = new DateTimeImmutable( $event_end );
+		$current_date_time = new DateTimeImmutable( 'now', new DateTimeZone( 'UTC' ) );
+
+		$interval       = $end_date_time->diff( $current_date_time );
+		$hours_left     = ( $interval->d * 24 ) + $interval->h;
+		$hours_in_a_day = 24;
+
+		if ( $hours_left <= $hours_in_a_day ) {
+			return 'ends in ' . $hours_left . ' hours';
+		} else {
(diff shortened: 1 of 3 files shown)


includes/route.php
```diff
@@ -316,4 +316,26 @@ public function events_user_created() {
...
 	}
+
+	/**
+	 * Generate text for the end date.
+	 *
+	 * @param string $event_end The end date.
+	 *
+	 * @return string The end date text.
+	 */
+	public static function get_end_date_text( string $event_end ): string {
```
akirk: What about putting this in the `Event` class? Seems like a better fit?

includes/event.php
```diff
@@ -52,4 +52,26 @@ public function end(): DateTimeImmutable {
...
+	public static function get_end_date_text( string $event_end ): string {
+		$end_date_time     = new DateTimeImmutable( $event_end );
+		$current_date_time = new DateTimeImmutable( 'now', new DateTimeZone( 'UTC' ) );
+
+		$interval       = $end_date_time->diff( $current_date_time );
+		$hours_left     = ( $interval->d * 24 ) + $interval->h;
+		$hours_in_a_day = 24;
+
+		if ( $hours_left <= $hours_in_a_day ) {
+			return 'ends in ' . $hours_left . ' hours';
```
akirk: Please use translation functions here, specifically `_n()` it will give us the hour/hours handling for free.

includes/event.php
```diff
@@ -52,4 +52,26 @@ public function end(): DateTimeImmutable {
...
+		} else {
+			return 'until ' . $end_date_time->format( 'M j, Y' );
```
akirk: Let's avoid the string concat and use a `sprintf()` with a `__()` and a placeholder.

akirk (changes requested):
Works well! Just the strings require some more attention.

akirk (approved):
Worsk well in my testing!

PR #113, opened 2024-02-19
Prevent the usage of event titles that can conflict with event URLs

### LINKED ISSUE #89: Prevent creating events with slugs we use
- new
- my-events

amieiro:
Currently, we can create events with some problematic URLs, used in the routes:
- new
- edit
- attend
- my-events

This PR adds a check to avoid the creation of events with these titles:

image

Fixes https://github.com/WordPress/wporg-gp-translation-events/issues/89.

### DIFF
--- a/wporg-gp-translation-events.php
+++ b/wporg-gp-translation-events.php
@@ -160,7 +160,12 @@ function submit_event_ajax() {
 		}
 	}
 	if ( ! $is_nonce_valid ) {
-		wp_send_json_error( 'Nonce verification failed' );
+		wp_send_json_error( 'Nonce verification failed', 403 );
+	}
+	// This is a list of slugs that are not allowed, as they conflict with the event URLs.
+	$invalid_slugs = array( 'new', 'edit', 'attend', 'my-events' );
+	if ( isset( $_POST['event_title'] ) && in_array( sanitize_title( wp_unslash( $_POST['event_title'] ) ), $invalid_slugs, true ) ) {
+		wp_send_json_error( 'Invalid slug', 403 );
 	}
 
 	$title          = isset( $_POST['event_title'] ) ? sanitize_text_field( wp_unslash( $_POST['event_title'] ) ) : '';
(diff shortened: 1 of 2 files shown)


wporg-gp-translation-events.php
```diff
@@ -160,7 +160,12 @@ function submit_event_ajax() {
 		}
 	}
 	if ( ! $is_nonce_valid ) {
-		wp_send_json_error( 'Nonce verification failed' );
+		wp_send_json_error( 'Nonce verification failed', 403 );
+	}
+	// This is a list of slugs that are not allowed, as they conflict with the event URLs.
+	$invalid_slugs = array( 'new', 'edit', 'attend', 'my-events' );
+	if ( isset( $_POST['event_title'] ) && in_array( sanitize_title( wp_unslash( $_POST['event_title'] ) ), $invalid_slugs, true ) ) {
```
psrpinto: Below, on line 171 we're declaring the `$title` variable. We could use that variable here.

psrpinto (commented):
Tested and it works

wporg-gp-translation-events.php
amieiro: Done!

wporg-gp-translation-events.php
```diff
@@ -163,16 +163,15 @@ function submit_event_ajax() {
...
-		wp_send_json_error( 'Invalid slug', 403 );
-	}
-
+	$invalid_slugs  = array( 'new', 'edit', 'attend', 'my-events' );
 	$title          = isset( $_POST['event_title'] ) ? sanitize_text_field( wp_unslash( $_POST['event_title'] ) ) : '';
 	$description    = isset( $_POST['event_description'] ) ? sanitize_text_field( wp_unslash( $_POST['event_description'] ) ) : '';
 	$event_start    = isset( $_POST['event_start'] ) ? sanitize_text_field( wp_unslash( $_POST['event_start'] ) ) : '';
 	$event_end      = isset( $_POST['event_end'] ) ? sanitize_text_field( wp_unslash( $_POST['event_end'] ) ) : '';
 	$event_timezone = isset( $_POST['event_timezone'] ) ? sanitize_text_field( wp_unslash( $_POST['event_timezone'] ) ) : '';
+	if ( isset( $title ) && in_array( sanitize_title( $title ), $invalid_slugs, true ) ) {
```
psrpinto, line 172: It has been sanitized above, there's no need to sanitize, right? Though maybe above it should be changed to `sanitize_title()` instead of `sanitize_text_field()`?

psrpinto (approved):

wporg-gp-translation-events.php
amieiro: What I want in this line is the slug, to look for it in the array, so I think is the correct function, because the title has been sanitized before.

- [sanitize_text_field](https://developer.wordpress.org/reference/functions/sanitize_text_field/). Sanitizes a string from user input or from the database.
- [sanitize_title](https://developer.wordpress.org/reference/functions/sanitize_title/). Sanitizes a string into a slug.

psrpinto: Ah, I understand now, thanks 👍

PR #116, opened 2024-02-21
Add the footer in some templates

### LINKED ISSUE #91: Add footer to some templates
Some templates don't have the footer. So they miss the standard WordPress.org footer. We need to add the footer to them by including the gp-footer template?

amieiro:
Some templates don't have a footer. This PR adds some footers in these templates.

### Before

image

### After

image

## My events

This template had a footer, but I have to remove a `</div>`, because the `gp_tmpl_footer()` function adds an extra `</div>` just at the beginning.

### Before

image

### After

image

### Before

image

### After

image

### Before

image

### After

image

### Before

image

### After

image

Fixes https://github.com/WordPress/wporg-gp-translation-events/issues/91.

### DIFF
--- a/templates/event.php
+++ b/templates/event.php
@@ -106,4 +106,5 @@ gp_tmpl_load( 'events-header', get_defined_vars(), __DIR__ );
 		</div>
 		<?php endif; ?>
 	</div>
-</div>
+<div class="clear"></div>
+<?php gp_tmpl_footer(); ?>
--- a/templates/events-form.php
+++ b/templates/events-form.php
@@ -83,5 +83,5 @@ gp_tmpl_load( 'events-header', get_defined_vars(), __DIR__ );
 	<?php endif; ?>
 	</div>
 </form>
-
-</div>
+<div class="clear"></div>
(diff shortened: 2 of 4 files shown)


akirk:
@amieiro you requested review but it is still a draft.

Should we create our own footer template that then calls the GlotPress footer? I could see us wanting to have a standardized footer across all of the event pages. For example with links to help pages, or the P2 post where it was announced.

amieiro:
> @amieiro you requested review but it is still a draft.
> 
> Should we create our own footer template that then calls the GlotPress footer? I could see us wanting to have a standardized footer across all of the event pages. For example with links to help pages, or the P2 post where it was announced.

Sorry, Alex. By accident, I removed all the PR description, so I put it in draft just after requesting the review. Now you have the full description. We have the [gp_footer](https://github.com/GlotPress/GlotPress/blob/4d2521bb17ec78f7ab8efa7328d8a79610fa5404/gp-includes/template.php#L93) hook, so I think we can use it if we want to update some contents.

akirk (approved):

psrpinto (approved):

PR #118, opened 2024-02-21
Show time format based on locale

trymebytes:
Some locale use the 12-hour format while some use the 24-hour time format. In this PR, the time format is determined by the locale and not hardcoded.

**Before**
Screenshot 2024-02-21 at 14 36 35

**After** (de-De locale)
image


akirk (approved):
Works great, thanks!

PR #121, opened 2024-02-23
Add breadcrumb to pages

### LINKED ISSUE #82: Add breadcrumb navigation
Users need to be able to move easily across different pages.

trymebytes:
Fixes #82 

**Screenshots**

> image

> image

> image


akirk (changes requested):
It works but wouldn't it be easier to just implement the `gp_breadcrumb_items` filter and add the base breadcrumb there?

akirk (commented):
The Create Event page doesn't have a breadcrumb item to go back to the events page

trymebytes:
> The Create Event page doesn't have a breadcrumb item to go back to the events page

That's true, I could also add the breadcrumb to go back to the events page too but I chose this approach to ensure consistency with the way it's handled in GlotPress, see screenshot below;
Screenshot 2024-02-23 at 15 08 11

Also we have the `Events` link the top nav bar too.

amieiro:
@trymebytes FYI, today I merged a PR in GlotPress that improves the consistency in all the breadcrumbs. This is [the PR](https://github.com/GlotPress/GlotPress/pull/1789) and this is [the issue](https://github.com/GlotPress/GlotPress/issues/1788).

includes/route.php
```diff
@@ -385,4 +385,13 @@ public function my_events() {
 	public static function convertToTimezone( string $date_time, string $time_zone ): string {
 		return ( new DateTime( $date_time, new DateTimeZone( 'UTC' ) ) )->setTimezone( new DateTimeZone( $time_zone ) )->format( 'Y-m-d H:i:s' );
 	}
+
+	/**
+	 * Get the base breadcrumb for events page.
+	 *
+	 * @return array The base breadcrumb.
+	 */
+	public static function get_base_breadcrumb(): array {
```
psrpinto: Instead of adding a `get_base_breadcrumb()` function we could add a `set_breadcrumb()` function which would contain all the logic for breadcrumbs. Something like the following:

```php
public static function set_breadcrumb( string $name = '' ) {
	if ( '' === $name ) {
		gp_breadcrumb( array( __( 'Events' ) ) );
		return;
	}

	$breadcrumb   = array( gp_link_get( gp_url( '/events' ), __( 'Events' ) ) );
	$breadcrumb[] = esc_html( $name );
	gp_breadcrumb( $breadcrumb );
}
```

Then the templates can simply call:

```php
Route::set_breadcrumb( $event_title );
```

I think this would make for more maintainable code because the logic is concentrated on a single location instead of being in all templates.

templates/event.php
```diff
@@ -18,6 +18,9 @@
 
 /* translators: %s: Event title. */
 gp_title( sprintf( __( 'Translation Events - %s' ), esc_html( $event_title ) ) );
+$breadcrumb   = route::get_base_breadcrumb();
```
psrpinto: I was confused initially because I didn't realize `route` meant the `Route` class. I think it would be best to capitialize `route` so it matches the class name, `Route`. I'm surprised that it works with a lowercase `r`, and that `phpcs` doesn't complain.

trymebytes:
I wonder if the function `gp_breadcrumb_event` should be renamed?

templates/events-user-created.php
```diff
@@ -0,0 +1,72 @@
+<?php
+/**
+ * Template for My Events.
```
psrpinto: I guess this file was erroneously re-added, since this has now moved to `event-list.php`

akirk:
Yes, what about `gp_breadcrumb_translation_events()`?

psrpinto (commented):
Looks great!

includes/route.php
psrpinto: No longer relevant since a new approach is being taken now.

templates/event.php
psrpinto: No longer relevant since a new approach is being taken now.

psrpinto (changes requested):
There's a file that I think was added by accident

trymebytes:
> There's a file that I think was added by accident

Ah thank you, must have been added during a rebase I did.

psrpinto (approved):

akirk:
It feels a bit weird that this is two lines:
Screenshot 2024-02-27 at 13 34 34

Can you put the My Events / Create Events in the same line as the breadcrumbs so that we don't waste vertical space?

akirk (approved):
I discussed this with @trymebytes and we found that like on translate.wordpress.org the buttons should not be aligned with the breadcrumbs but with the headline. Tosin will work on this in a follow-up.

akirk:
I also noticed that the breadcrumbs reach all to the left. outside of the rest. Let's just make everything 100% width and not 80%.

PR #122, opened 2024-02-23
Add the "Events I attended" view

### LINKED ISSUE #120: Attended Events
You should be able to view the events you attended in the past.

amieiro:
We don't have a view to show the events attended by a user.

This PR adds:
- A new item in the top right menu to go to this option, with a vertical separator with the next item in the menu.
- A new route.
- A new method to get the events attended by a user and call to the template.
- A new template to see the events attended by a user.

image

The template adds pagination, as you can see in the next screenshot, with `'posts_per_page' => 1`.

image

Fixes https://github.com/WordPress/wporg-gp-translation-events/issues/120.


wporg-gp-translation-events.php
```diff
@@ -399,6 +399,7 @@ function () {
 		GP::$router->add( '/events/edit/(\d+)', array( 'Wporg\TranslationEvents\Route', 'events_edit' ) );
 		GP::$router->add( '/events/attend/(\d+)', array( 'Wporg\TranslationEvents\Route', 'events_attend' ), 'post' );
 		GP::$router->add( '/events/my-events', array( 'Wporg\TranslationEvents\Route', 'events_user_created' ) );
+		GP::$router->add( '/events/events-i-attended', array( 'Wporg\TranslationEvents\Route', 'events_i_attended' ) );
```
akirk: If we add something here we also need to add it to `$invalid_slugs`

akirk (commented):
It works well in my testing. I wonder though if would also work as another section of "My Events". I could see most people never creating events and then we'd have links in the top navigation with mostly no content. What do you think?

psrpinto (commented):
Nice! I tend to agree with Alex's suggestion of showing past events in the "My events" page.

amieiro:
So, what you are proposing is to add a new section in the "My events" page, with the "Events I Attended" 

```
My events (Events I have created)
- Event 1 
- Event 2

Events I attended
- Event 1 
- Event 2
```
Ok, I like this approach. I am going to update the PR with these changes.

akirk:
Yes, I suppose we can later finetune by hiding sections when empty, display a "create event" link in the first section etc. But let's keep the scope of this PR like you proposed.

amieiro:
I have added the "Events I attended" at the bottom of the "My events" template.

image

I added pagination for both tables.

image

I renamed the method from `events_user_created` to `my_events` and I did the same with the template, from `events-user-created.php` to `events-my-events.php`.

akirk (approved):
Tested and works well, also the pagination. Clever! Thanks!

PR #123, opened 2024-02-23
Tests for stats calculator

psrpinto:
This PR adds tests for the stats calculation, i.e. it tests that the table that is shown in the event page displays correct data.

### DIFF
--- a/tests/stats-calculator.php
+++ b/tests/stats-calculator.php
@@ -28,4 +28,16 @@ class Stats_Calculator_Test extends GP_UnitTestCase {
 
 		$this->assertFalse( $this->calculator->event_has_stats( $event ) );
 	}
+
+	public function test_tells_that_event_has_stats() {
+		$this->set_normal_user_as_current();
+		$user_id = wp_get_current_user()->ID;
+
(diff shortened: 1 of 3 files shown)


tests/stats-calculator.php
```diff
@@ -0,0 +1,94 @@
...
+		$user1_id = 42;
+		$user2_id = 43;
+		$user3_id = 44;
+
+		$event1_id = $this->event_factory->create_active( array( $user1_id ) );
+		$event2_id = $this->event_factory->create_active( array( $user1_id ) );
+
+		// For event1, aa locale, multiple users.
+		$this->stats_factory->create( $event1_id, $user1_id, 11, 'create' );
+		$this->stats_factory->create( $event1_id, $user2_id, 11, 'create' );
```
akirk: How could a user create two translations with the same id? I understand it's possible from a data perspective but it should not happen in reality.

akirk (commented):
Looks good but not sure about one test.

tests/stats-calculator.php
psrpinto: Makes sense, fixed.

akirk (approved):
Thank you!

PR #124, opened 2024-02-23
Add the past events in the main template

### LINKED ISSUE #119: Past Events view
After an event is over, there is no way to find it. We should create such a view.

amieiro:
This PR adds the "Past events" in the main template

image

It adds pagination (the next screenshot was taken with a pagination of 3 items per page).

image

Fixes https://github.com/WordPress/wporg-gp-translation-events/issues/119.


akirk (approved):
Tested and works well, thanks!

PR #126, opened 2024-02-26
Update the slug update logic and add a message for the published events to update their slug

### LINKED ISSUE #97: Slug management
It should be possible to change the slug before publishing an event. If you create an event that already has the same name, a `-2` is added etc but you cannot get rid of it if you change the event title.

- [x] When you change the title, the slug will update accordingly
- [x] When the event is published, you cannot change the title
- [x] When published, show a notice how to reach an admin to help you fix a typo in the title if you don't want the slug to change

amieiro:
This PR:

- Updates the slug generation, updating it with the title content only when:
  - The event status is `draft`.
  - The event status is `published` and it comes from a previous status of `draft`.
- Adds a message, only showed if the event status is `published` or just when its status changes from `draft`to `published`.

image

Fixes https://github.com/WordPress/wporg-gp-translation-events/issues/97.

### DIFF
--- a/wporg-gp-translation-events.php
+++ b/wporg-gp-translation-events.php
@@ -384,8 +384,10 @@ function generate_event_slug( array $data, array $postarr ): array {
 		if ( 'publish' === $data['post_status'] ) {
 			if ( is_numeric( $postarr['ID'] ) && 0 !== $postarr['ID'] ) {
 				$post = get_post( $postarr['ID'] );
-				if ( 'draft' === $post->post_status ) {
-					$data['post_name'] = sanitize_title( $data['post_title'] );
+				if ( $post instanceof WP_Post ) {
+					if ( 'draft' === $post->post_status ) {
+						$data['post_name'] = sanitize_title( $data['post_title'] );
+					}
 				}
 			}
 		}
--- a/wporg-gp-translation-events.php
+++ b/wporg-gp-translation-events.php
@@ -382,7 +382,7 @@ function generate_event_slug( array $data, array $postarr ): array {
 			$data['post_name'] = sanitize_title( $data['post_title'] );
 		}
 		if ( 'publish' === $data['post_status'] ) {
-			if ( is_numeric( $postarr['ID'] ) ) {
+			if ( is_numeric( $postarr['ID'] ) && 0 !== $postarr['ID'] ) {
 				$post = get_post( $postarr['ID'] );
 				if ( 'draft' === $post->post_status ) {
 					$data['post_name'] = sanitize_title( $data['post_title'] );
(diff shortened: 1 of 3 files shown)


trymebytes:
Please check in the failed tests and the PR is still in draft too.

wporg-gp-translation-events.php
```diff
@@ -366,24 +366,35 @@ function gp_event_nav_menu_items( array $items, string $location ): array {
...
-function generate_event_slug( array $data ): array {
-	if ( 'event' === $data['post_type'] && 'draft' === $data['post_status'] ) {
-		if ( empty( $data['post_name'] ) ) {
+function generate_event_slug( array $data, array $postarr ): array {
+	if ( 'event' === $data['post_type'] ) {
+		if ( 'draft' === $data['post_status'] ) {
 			$data['post_name'] = sanitize_title( $data['post_title'] );
 		}
+		if ( 'publish' === $data['post_status'] ) {
+			if ( is_numeric( $postarr['ID'] ) && 0 !== $postarr['ID'] ) {
```
psrpinto, line 385: I added the `0 !== $postarr['ID']` condition to make tests pass.

But this got me thinking: in the case where the event is not first saved as draft (i.e. user publishes directly), the `$data['post_name'] = sanitize_title( $data['post_title'] );` will never be called, right? I guess this is not what we want, but I'm not sure.

wporg-gp-translation-events.php
```diff
@@ -384,8 +384,10 @@ function generate_event_slug( array $data, array $postarr ): array {
 		if ( 'publish' === $data['post_status'] ) {
 			if ( is_numeric( $postarr['ID'] ) && 0 !== $postarr['ID'] ) {
 				$post = get_post( $postarr['ID'] );
-				if ( 'draft' === $post->post_status ) {
-					$data['post_name'] = sanitize_title( $data['post_title'] );
+				if ( $post instanceof WP_Post ) {
```
psrpinto, line 387: 👍

wporg-gp-translation-events.php
amieiro: > I added the `0 !== $postarr['ID']` condition to make tests pass.
> 
> But this got me thinking: in the case where the event is not first saved as draft (i.e. user publishes directly), the `$data['post_name'] = sanitize_title( $data['post_title'] );` will never be called, right? I guess this is not what we want, but I'm not sure.

If the event is not saved previously as a draft, the slug problems will be addressed by the [wp_insert_post](https://developer.wordpress.org/reference/functions/wp_insert_post/) function, so I think everything will be ok.

psrpinto: Ah, that makes sense 👍

psrpinto (approved):
Tested 👍

PR #135, opened 2024-02-26
Fix attending list bug

### LINKED ISSUE #134: Bug in the Events I'm Attending logic
A user is seen attending an event they have not opted to join.

To reproduce;
- Create and publish an event as User A.
- Switch to User B, you would see the event just created by User A in the "Events I'm attending" section even though User B has not opted to attend. 
Screenshot 2024-02-26 at 17 22 00

Go to the page of the event created by User A and the "Attend" button is still intact. See screenshots below;

Screenshot 2024-02-26 at 17 22 28

trymebytes:
Fixes #134

### DIFF
--- a/includes/route.php
+++ b/includes/route.php
@@ -126,7 +126,7 @@ class Route extends GP_Route {
 		);
 		$past_events_query = new WP_Query( $past_events_args );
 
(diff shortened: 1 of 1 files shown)


psrpinto (approved):
Thanks!

PR #137, opened 2024-02-27
Allow only event admin view event page for draft events

### LINKED ISSUE #102: Hide event page from non-admin when it's in draft status
The page of an event in draft status should only be viewable/accessible by the user that created the event.

trymebytes:
Fixes #102

### DIFF
--- a/includes/route.php
+++ b/includes/route.php
@@ -247,7 +247,7 @@ class Route extends GP_Route {
 		if ( ! $event ) {
 			$this->die_with_404();
 		}
-		if ( 'draft' === $event->post_status && intval( $event->post_author ) !== $user->ID ) {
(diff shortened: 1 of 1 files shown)


includes/route.php
```diff
@@ -247,6 +247,9 @@ public function events_details( string $event_slug ) {
 		if ( ! $event ) {
 			$this->die_with_404();
 		}
+		if ( 'draft' === $event->post_status && intval( $event->post_author ) !== $user->ID ) {
```
akirk: Would this also work but be would reach further?
```suggestion
		if ( 'published' !== $event->post_status && intval( $event->post_author ) !== $user->ID ) {
```

akirk: But I think we also need to allow admins to view/edit it or people with higher priviledges

psrpinto (approved):

PR #138, opened 2024-02-28
Enable the event creation, update and delete only to administrators

### LINKED ISSUE #136: Allow only administrators to create new events and update/delete the existing events
Allow only administrators to create new events and update/delete the existing events. Add a hook to be able to overwrite it at translate.w.org, so only the administrators and GTEs can create new events and update/delete the existing events.

amieiro:
This PR:
- Enables the event creation, update and delete only to administrators.
- Adds a hook to overwrite it.
- Updates some validations and strings, to be able to translate them.

Fixes https://github.com/WordPress/wporg-gp-translation-events/issues/136.

### DIFF
--- a/wporg-gp-translation-events.php
+++ b/wporg-gp-translation-events.php
@@ -24,6 +24,7 @@ use Exception;
 use GP;
 use WP_Post;
 use WP_Query;
+use function Symfony\Component\VarDumper\Dumper\esc;
 
 /**
  * Check if a slug is being used by another post type.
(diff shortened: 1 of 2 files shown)


wporg-gp-translation-events.php
```diff
@@ -24,6 +24,7 @@
 use GP;
 use WP_Post;
 use WP_Query;
+use function Symfony\Component\VarDumper\Dumper\esc;
```
trymebytes: I guess this is for debugging purposes.

amieiro: Yes, I think it was added by the IDE, because it is unused, and I don't use it to debug. I am going to remove it in a new commit. Thank you!

wporg-gp-translation-events.php
```diff
@@ -178,7 +206,7 @@ function submit_event_ajax() {
 	$event_end      = isset( $_POST['event_end'] ) ? sanitize_text_field( wp_unslash( $_POST['event_end'] ) ) : '';
 	$event_timezone = isset( $_POST['event_timezone'] ) ? sanitize_text_field( wp_unslash( $_POST['event_timezone'] ) ) : '';
 	if ( isset( $title ) && in_array( sanitize_title( $title ), $invalid_slugs, true ) ) {
-		wp_send_json_error( 'Invalid slug', 403 );
+		wp_send_json_error( esc_html__( 'Invalid slug.', 'gp-translation-events' ), 422 );
```
psrpinto, line 208: 422 👍

psrpinto (approved):
Looks good!

PR #139, opened 2024-02-28
Changes for Production

akirk:
This PR is actually quite far reaching and will require updates to your local db but I think we should do them based on feedback and to ensure things work well on translate.wordpress.org:
- Rename the table to `translate_event_actions` in accordance to other table names that start with `translate_`
- Add an autoinc primary key
- Make the current primary key a unique key to fulfil the same purpose
- Change the `action` varchar to an enum to save space and increase speed
- Change the `translation_id` to `original_id`, see below
- Change the post_type to `translation_event` to make this more unique. See below
- Use `dbDelta()` to create the table in the unit test
- Clean up `wporg-gp-translation-events.php` functions into a class, add a `CPT` class constant for this post type name.

#### Change the `translation_id` to `original_id`
I believe we should have done this from the start, since we also store the locale anyway. A user should get a credit per a translation and not per translation attempt. It could be easy to create thousands of wrong translations to boost your stats where as it takes more effort to do this per original. **You'll need to recreate the table [according to the new schema](https://github.com/WordPress/wporg-gp-translation-events/blob/improve-sql-schema/schema.sql)**

#### Change the post_type to `translation_event`
I think choosing just "event" as the post_type name was a bad idea since it could overlap with something else. I therefore think we should change it before we start. **You'll all have to run this in your db**: `update wp_posts set post_type = 'translation_event' where post_type = 'event'`


akirk:
> A user should get a credit per a translation and not per translation attempt.

Specifically, if a user makes a typo in his first translation attempt and then fixes it, they get credit for two translations. I don't think it should be the case and we achieve this by switching to `original_id`.

schema.sql
```diff
@@ -1,19 +1,11 @@
...
-create index wp_wporg_gp_translation_events_actions_event_id_locale_index
-    on wp_wporg_gp_translation_events_actions (event_id, locale);
+CREATE TABLE `translate_event_actions` (
+  `translate_event_actions_id` int(11) unsigned NOT NULL AUTO_INCREMENT,
+  `event_id` int(10) NOT NULL COMMENT 'Post_ID of the translation_event post in the wp_posts table',
+  `original_id` int(10) NOT NULL COMMENT 'ID of the translation',
+  `user_id` int(10) NOT NULL COMMENT 'ID of the user who made the action',
+  `action` enum('approve','create','reject','request_changes') NOT NULL COMMENT 'The action that the user made (create, reject, etc)',
+  `locale` varchar(10) CHARACTER SET latin1 COLLATE latin1_swedish_ci NOT NULL COMMENT 'Locale of the translation',
+  `happened_at` datetime NOT NULL COMMENT 'When the action happened, in UTC',
```
amieiro: We usually use in the translations table (not in all) a `date_added` for the creation of the row.
 
```suggestion
  `date_added` datetime NOT NULL COMMENT 'When the action happened, in UTC',
```

In some tables (e.g. in the `translate_translations` table), we have another field with the modification time, usually called `date_modified`. Does it make sense to create it now or is better to add it in the future if we need it?

schema.sql
```diff
@@ -1,19 +1,11 @@
...
+  PRIMARY KEY (`translate_event_actions_id`),
+  UNIQUE KEY `event_per_translated_original_per_user` (`event_id`,`locale`,`original_id`,`user_id`)
```
amieiro: A user can add several actions with the same item (locale and original string) that should be registered. E.g., a validator that rejects a translation and then adds a new translation for the same locale and original string. With the current key, it will count as one action, but it should be 2 actions. 

```suggestion
  UNIQUE KEY `event_per_translated_original_per_user` (`event_id`,`locale`,`original_id`,`user_id`, `action`)
```

amieiro (commented):
I have tested the PR, creating the table from scratch, and works fine. I have made some suggestions in the table structure. If some of them are taken into account, this changes should be propagated to the code.

schema.sql
akirk: We had actually agreed that we don't want to count this as two actions so that we don't encourage behavior where someone would create a translation as waiting and then approve it. I'd say rejecting and fixing counts as review an this is how it's implemented.

schema.sql
akirk: Makes sense, I changed it!

psrpinto: I called it `happened` instead of `added` or `created` because I find that those terms can be confusing, as actions can be other things than "translation added". E.g. a "reject" wasn't added, it happened :)

psrpinto: Another thing I find confusing with the new naming is that the field is called `date_` but its type is `datetime`.

akirk: I'm ok with going back to `happened_at` or `created_at`.

includes/route.php
```diff
@@ -64,7 +64,7 @@ public function events_list() {
 		// phpcs:enable
 
 		$current_events_args  = array(
-			'post_type'      => 'event',
+			'post_type'      => 'translation_event',
```
psrpinto: 👍

tests/bootstrap.php
```diff
@@ -29,22 +29,10 @@ function _glotpress_path( string $path ): string {
...
-		if ( ! $sql ) {
-			continue;
-		}
-		// phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery
-		// phpcs:disable WordPress.DB.DirectDatabaseQuery.NoCaching
-		// phpcs:disable WordPress.DB.PreparedSQL.NotPrepared
-		$wpdb->query( $sql );
-		// phpcs:enable
-	}
+	$statements = dbDelta( file_get_contents( $schema_path ) );
```
psrpinto: Didn't know about `dbDelta()`! 👌

schema.sql
```diff
@@ -3,9 +3,10 @@ CREATE TABLE `translate_event_actions` (
   `event_id` int(10) NOT NULL COMMENT 'Post_ID of the translation_event post in the wp_posts table',
   `original_id` int(10) NOT NULL COMMENT 'ID of the translation',
   `user_id` int(10) NOT NULL COMMENT 'ID of the user who made the action',
-  `action` ENUM('approve','create','reject','request_changes') NOT NULL COMMENT 'The action that the user made (create, reject, etc)',
-  `locale` varchar(10) NOT NULL COMMENT 'Locale of the translation',
+  `action` enum('approve','create','reject','request_changes') NOT NULL COMMENT 'The action that the user made (create, reject, etc)',
+  `locale` varchar(10) CHARACTER SET latin1 COLLATE latin1_swedish_ci NOT NULL COMMENT 'Locale of the translation',
```
psrpinto: Was this a deliberate change or accidental? If deliberate, why `swedish_ci`? Just curious :)

psrpinto (commented):
Thanks for working on this!

I wasn't sure where to leave this comment so leaving it here:

~[This commit](https://github.com/WordPress/wporg-gp-translation-events/pull/139/commits/ef16751f140c538bf535d08899d31a0b51a77a8c#diff-93ac378e918721e6698187d0123e47b161f8fe273064813e9cdef3433a9165a8) changes lots of things in `wporg-gp-translation-events.php` but I'm not sure I understand what's being changed. Could the file have been modified accidentally?~

Nevermind, I understand now, moving code into a class 👍

schema.sql
akirk: It's the exact same charset as the locale on the GlotPress tables. It's a MySQL standard (they are Swedish).

wporg-gp-translation-events.php
```diff
@@ -25,499 +25,501 @@
...
+		if ( isset( $title ) && in_array( sanitize_title( $title ), $invalid_slugs, true ) ) {
+			wp_send_json_error( 'Invalid slug', 403 );
+		}
+
+		$is_valid_event_date = false;
 		try {
-			$event_stats = $stats_calculator->for_event( $event );
-		} catch ( Exception $e ) {
-			wp_send_json_error( 'Failed to calculate event stats' );
+			$is_valid_event_date = validate_event_dates( $event_start, $event_end );
```
trymebytes:
```suggestion
			$is_valid_event_date = $this->validate_event_dates( $event_start, $event_end );
```

akirk:
@psrpinto I actually introduced a class to clean up what I've seen as a bit of a mess with spread-out `add_action/filter` calls, and add the CPT constant to the class.

wporg-gp-translation-events.php
```diff
@@ -25,499 +25,501 @@
...
+			}
+			wp_trash_post( $event_id );
+			$response_message = 'Event deleted successfully!';
+		}
+		if ( ! $event_id ) {
+			wp_send_json_error( 'Event could not be created or updated' );
+		}
+		if ( 'delete_event' !== $_POST['form_name'] ) {
+			try {
+				update_post_meta( $event_id, '_event_start', convert_to_utc( $event_start, $event_timezone ) );
```
trymebytes:
```suggestion
				update_post_meta( $event_id, '_event_start', $this->convert_to_utc( $event_start, $event_timezone ) );
```

wporg-gp-translation-events.php
```diff
@@ -25,499 +25,501 @@
...
+				update_post_meta( $event_id, '_event_end', convert_to_utc( $event_end, $event_timezone ) );
```
trymebytes:
```suggestion
				update_post_meta( $event_id, '_event_end', $this->convert_to_utc( $event_end, $event_timezone ) );
```

trymebytes (changes requested):
I couldn't create an event  because of undefined functions.

psrpinto:
Gotcha, thanks for clarifying. Indeed I reviewed that without whitespace and makes sense now.

akirk:
Thanks @trymebytes! I applied your suggestions and now it works for me.

schema.sql
psrpinto: Now that were tracking actions on the original instead of the translation, I'm wondering if we need to reconsider the "one action per user per locale" thing. Because indeed like @amieiro mentions, new translations or review actions on the same original by the same user won't count towards stats.

For example, if there are multiple waiting translations for the same original, and a reviewer approves one but rejects the other, only one of those review counts towards stats, while the reviewer did do the work twice.

It's a subtle thing, probably an edge case, but if we would want to address it, one simple way could be to also add the `translation_id` to the table, and include it in the unique key instead of `original_id`.

akirk: As soon as one suggestion is approved the other suggestions become `old`, so I see it as the review work is done only once.

psrpinto: After re-reading the reasoning for using `original_id` instead of `translation_id` in the PR description, I guess my suggestion above of also adding `translation_id` doesn't really make sense as it would defeat the purpose.

I guess it's a tradeoff between two edge cases: preventing users from gaming stats, or accurately counting actions on the same translation.

akirk: I see it this way: you can get one credit for working on a translation of an original, either translate it, or review it. You can get the credit in multiple languages for the same original but not twice in the same language.

psrpinto: > As soon as one suggestion is approved the other suggestions become old, so I see it as the review work is done only once.

That makes sense, I wasn't aware of this.

psrpinto (approved):

trymebytes (approved):
Tested and works!

PR #144, opened 2024-02-29
Check if the start and end dates have been filled in

### LINKED ISSUE #140: Error on creating new event related to start and end date.
Hi,

While testing locally, can't add an event because it gives an error message about the dates not being correct.

imagem

Actually the start date is before the end date, but the field values are empty.
I seems that these .val() are getting empty:
https://github.com/WordPress/wporg-gp-translation-events/blob/trunk/assets/js/translation-events.js#L40

amieiro:
Creating a new event, if you are using Chrome or another Chromium browser, when you select the start and end date, the hour is selected by default to the current one.

image

If you use Firefox, these hours are not set by default.

image

If you try to create an event without start or end time, you will get the "Event end date and time must be later than event start date and time." error.

image

This PR checks if both dates are completed, returning and error if one of them are not set. 

Fixes https://github.com/WordPress/wporg-gp-translation-events/issues/140.

### DIFF
--- a/assets/js/translation-events.js
+++ b/assets/js/translation-events.js
@@ -37,6 +37,14 @@
 		 * @param isDraft	  Whether the current event status is a draft or not
 		 */
 		function handleSubmit( eventStatus, isDraft ) {
+			if ( '' === $( '#event-start' ).val() ) {
+				$gp.notices.error( 'Event start date and time must be set.' );
+				return;
+			}
+			if ( '' === $( '#event-end' ).val() ) {
+				$gp.notices.error( 'Event end date and time must be set.' );
+				return;
+			}
 			if ( $( '#event-end' ).val() <= $( '#event-start' ).val() ) {
 				$gp.notices.error( 'Event end date and time must be later than event start date and time.' );
 				return;


assets/js/translation-events.js
```diff
@@ -37,6 +37,14 @@
 		 * @param isDraft	  Whether the current event status is a draft or not
 		 */
 		function handleSubmit( eventStatus, isDraft ) {
+			if ( '' === $( '#event-start' ).val() ) {
```
psrpinto, line 40: Do we need to check this on the backend as well? I'm not sure what the behaviour is [here](https://github.com/WordPress/wporg-gp-translation-events/blob/trunk/wporg-gp-translation-events.php#L142-L143) at the moment, maybe it already throws when the hour is not set, but I'm not sure.

amieiro: We check it in the backend [here](https://github.com/WordPress/wporg-gp-translation-events/blob/ae6049aa8ef5a69728fbba5116a4a602c08825ee/wporg-gp-translation-events.php#L184-L192).

psrpinto (approved):

Issue #117, opened 2024-02-21
Refactor the code that retrieves active events the user is attending


akirk:
@psrpinto [said](https://github.com/WordPress/wporg-gp-translation-events/pull/114/files#r1496047478): 
> I think maybe we should consider refactoring [the code that retrieves active events the user is attending](https://github.com/WordPress/wporg-gp-translation-events/blob/trunk/includes/translation-listener.php#L69-L70) to something reusable, as I noticed we need that in several places.
> 
> If we would do that, we would automatically get caching for the query that retrieves active events, as that can be an expensive query.

psrpinto:
This is now fixed since the introduction of the event repository.

PR #145, opened 2024-03-01
Proposal: Event repository

psrpinto:
> Please read this description before reviewing the PR :) It provides important context that I think is essential to consider while reviewing.
>
> I would recommend reviewing the final result instead of individual commits.

Recently, we identified the need to refactor the code that retrieves active events, as documented in #117. While looking into that issue, I realised that there was opportunity for a more extensive refactor, that would address #117 but also improve other parts of the codebase.

I took the liberty to implement such a refactor, and I'm proposing it in this PR. If you agree that this proposal is beneficial to the project, we can iterate on it here until we're happy with the result. If you don't agree, we can scrap this PR.

Note that the code introduced in here is not being called anywhere. Once/if this PR is merged, I will make it so that the codebase uses the code introduced here.

As always, I welcome any feedback you might have.
## The problem
Before going into the proposed solution, I think it's important to provide context on what problems this PR solves. What follows are some examples I identified that are addressed, or at least improved by this PR.
### Duplication of code
An example of this is retrieving active events (#117), where we have the same logic with slight variations in multiple places:

1. When storing stats
2. When displaying the active events of a user
3. When displaying the currently active events in the "homepage" (`/events`)
### Duplication of validation and/or incomplete validation
An example of this is creating an instance of an event. Whenever we need to create an instance of an event, we need to perform the same checks on the event. For example, checking that the `$end` is after the `$start`, or that the `$title` is not empty.

There are also instances where we are not fully validating the event data, for example, making sure that the title is not empty.
### Duplication of other checks
An example of this is validating that the `post_type` is `event`. This needs to be done every time an event is retrieved from the database. It should not be possible for the developer to retrieve an event from the database that is not of type `event`.
### Side-effects not guaranteed
An example of this is making sure that the active events cache is invalidated when events are created of modified. The developer must remember to invalidate the cache when required.
### Data access is hard to cover by unit tests
Because data access is spread throughout the codebase, it's difficult to cover with unit tests. We would only be able to test data access indirectly (by testing other features), but ensuring the validity of the stored data is essential, so it should be relatively easy for the developer to cover data access with tests.
### Error prone
IMHO, the combination of the above issues, results in error-prone code, because it's easy for the developer to make mistakes, because they need to consider a large number of things when making changes.

For example, when a developer is writing business logic (currently in the `Route` class and main plugin file), I think they should not need to consider details of data access like validating that the post has type `event`. 

IMHO all the little things the developer must consider add up, and make for a codebase that is hard to understand and maintain.
## Proposed solution
To address some of the issues mentioned above, this PR introduces a *Data access layer* for events. This layer can be thought of as a black box that abstracts details of how events are stored, through a higher-level API for manipulating and retrieving events. 

This API is specified by the `Event_Repository_Interface`, which I believe contains all the operations we currently need to perform on events:

```php
interface Event_Repository_Interface {  
    public function create_event( Event $event ): void;  
    public function update_event( Event $event ): void;  
    public function get_event( int $id ): Event;  
    public function get_current_events( int $page = -1, int $page_size = -1 ): Events_Query_Result;  
    public function get_current_events_for_user( int $user_id, int $page = -1, int $page_size = -1 ): Events_Query_Result;  
    public function get_past_events_for_user( int $user_id, int $page = -1, int $page_size = -1 ): Events_Query_Result;  
    public function get_events_created_by_user( int $user_id, int $page = -1, int $page_size = -1 ): Events_Query_Result;  
}
```

`Event` represents an event, and contains all the data of the event, e.g. `$start`, `$end`, `$timezone`, `$title`, etc. It is not possible for there to be an instance of `Event` that contains invalid data. The `Event` constructor and setters perform validation, which ensures we're validating as soon as possible, before data gets stored in the database.

Operations that retrieve a list of events (e.g. `get_current_events()` or `get_events_created_by_user()`) return an instance of `Events_Query_Result`, which carries the retrieved events themselves, and information needed for pagination (e.g. total number of results).

There are two implementations of `Event_Repository_Interface`:

1. `Event_Repository`: this is the "normal" repository
2. `Event_Repository_Cached`: extends `Event_Repository` and overrides methods that require caching.

The reasoning for having an interface and two different implementations is so that calling code does not depend on a specific implementation. The calling code should not need to care whether caching is enabled or not.

Finally, this "black box" is covered by some unit tests, so we can be reasonably confident that it will operate as expected.

### DIFF
--- a/includes/active-events-cache.php
+++ b/includes/active-events-cache.php
@@ -3,6 +3,7 @@
 namespace Wporg\TranslationEvents;
 
 use Exception;
+use Wporg\TranslationEvents\Event\Event;
 
 class Active_Events_Cache {
 	public const CACHE_DURATION = 60 * 60 * 24; // 24 hours.
--- a/includes/active-events-cache.php
+++ b/includes/active-events-cache.php
@@ -6,7 +6,7 @@ use Exception;
 use Wporg\TranslationEvents\Event\Event;
 
 class Active_Events_Cache {
-	public const CACHE_DURATION = 60 * 60 * 24; // 24 hours.
+	public const CACHE_DURATION = DAY_IN_SECONDS;
 	private const KEY           = 'translation-events-active-events';
 
 	/**
--- a/includes/event/event-repository-cached.php
+++ b/includes/event/event-repository-cached.php
@@ -20,8 +20,8 @@ class Event_Repository_Cached extends Event_Repository {
 		$this->invalidate_cache();
 	}
 
-	public function get_current_events( int $current_page = -1, int $page_size = -1 ): Events_Query_Result {
-		$this->assert_pagination_arguments( $current_page, $page_size );
+	public function get_current_events( int $page = -1, int $page_size = -1 ): Events_Query_Result {
+		$this->assert_pagination_arguments( $page, $page_size );
 
 		$cache_duration = self::CACHE_DURATION;
 		$now            = new DateTimeImmutable( 'now', new DateTimeZone( 'UTC' ) );
@@ -46,16 +46,16 @@ class Event_Repository_Cached extends Event_Repository {
 			)
 		);
 
-		if ( $current_page > 0 && $page_size > 0 ) {
+		if ( $page > 0 && $page_size > 0 ) {
 			// Convert from 1-indexed to 0-indexed.
-			--$current_page;
+			--$page;
 			$pages = array_chunk( $events, $page_size );
 		} else {
-			$current_page = 0;
-			$pages        = array( $events );
+			$page  = 0;
+			$pages = array( $events );
 		}
 
-		return new Events_Query_Result( $pages[ $current_page ], count( $pages ) );
+		return new Events_Query_Result( $pages[ $page ], count( $pages ) );
 	}
 
 	private function invalidate_cache(): void {
--- a/includes/event/event-repository-cached.php
+++ b/includes/event/event-repository-cached.php
@@ -46,7 +46,7 @@ class Event_Repository_Cached extends Event_Repository {
 
 		// TODO: paginate results.
 
-		return new Event_Query_Result( $events );
+		return new Event_Query_Result( $events, 1 );
 	}
 
 	private function invalidate_cache(): void {
--- a/includes/event/event-repository-cached.php
+++ b/includes/event/event-repository-cached.php
@@ -20,7 +20,7 @@ class Event_Repository_Cached extends Event_Repository {
 		$this->invalidate_cache();
 	}
 
-	public function get_current_events(): Event_Query_Result {
+	public function get_current_events( int $current_page = -1, int $page_size = -1 ): Event_Query_Result {
 		$cache_duration = self::CACHE_DURATION;
 		$now            = new DateTimeImmutable( 'now', new DateTimeZone( 'UTC' ) );
 		$boundary_start = $now;
@@ -43,6 +43,8 @@ class Event_Repository_Cached extends Event_Repository {
 				}
 			)
 		);
+
+		// TODO: paginate results.
 
 		return new Event_Query_Result( $events );
 	}
--- a/includes/event/event-repository-cached.php
+++ b/includes/event/event-repository-cached.php
@@ -20,7 +20,7 @@ class Event_Repository_Cached extends Event_Repository {
 		$this->invalidate_cache();
 	}
 
-	public function get_current_events( int $current_page = -1, int $page_size = -1 ): Event_Query_Result {
+	public function get_current_events( int $current_page = -1, int $page_size = -1 ): Events_Query_Result {
 		$this->assert_pagination_arguments( $current_page, $page_size );
 
 		$cache_duration = self::CACHE_DURATION;
@@ -55,7 +55,7 @@ class Event_Repository_Cached extends Event_Repository {
 			$pages        = array( $events );
 		}
 
-		return new Event_Query_Result( $pages[ $current_page ], count( $pages ) );
+		return new Events_Query_Result( $pages[ $current_page ], count( $pages ) );
 	}
 
 	private function invalidate_cache(): void {
--- a/includes/event/event-repository-cached.php
+++ b/includes/event/event-repository-cached.php
@@ -21,9 +21,14 @@ class Event_Repository_Cached extends Event_Repository {
 		return $event_id_or_error;
 	}
 
-	public function update_event( Event $event ): void {
-		parent::update_event( $event );
+	public function update_event( Event $event ) {
+		$event_id_or_error = parent::update_event( $event );
+		if ( $event_id_or_error instanceof WP_Error ) {
+			return $event_id_or_error;
+		}
+
 		$this->invalidate_cache();
+		return $event_id_or_error;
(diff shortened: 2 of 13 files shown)


includes/event/event.php
```diff
@@ -0,0 +1,198 @@
...
+		if ( ! isset( $meta['_event_start'][0] ) || ! isset( $meta['_event_end'][0] ) || ! isset( $meta['_event_timezone'][0] ) ) {
+			throw new Exception( 'Invalid event meta' );
+		}
+
+		return new Event(
+			$id,
+			DateTimeImmutable::createFromFormat( 'Y-m-d H:i:s', $meta['_event_start'][0], new DateTimeZone( 'UTC' ) ),
+			DateTimeImmutable::createFromFormat( 'Y-m-d H:i:s', $meta['_event_end'][0], new DateTimeZone( 'UTC' ) ),
+			new DateTimeZone( $meta['_event_timezone'][0] ),
+			'foo-slug',  // TODO: this function will be removed, this is here so tests pass.
```
psrpinto, line 53: This will be removed in upcoming PRs, if/when this one is merged.

akirk:
@psrpinto the tests are failing, can you take a look?

psrpinto:
Tests are now passing, it was due to the change of the post type from `event` to `translation_event`.

includes/active-events-cache.php
```diff
@@ -3,6 +3,7 @@
 namespace Wporg\TranslationEvents;
 
 use Exception;
+use Wporg\TranslationEvents\Event\Event;
 
 class Active_Events_Cache {
 	public const CACHE_DURATION = 60 * 60 * 24; // 24 hours.
```
amieiro:
```suggestion
	public const CACHE_DURATION = DAY_IN_SECONDS;
```

includes/event/event-repository-cached.php
```diff
@@ -0,0 +1,64 @@
+<?php
+
+namespace Wporg\TranslationEvents\Event;
+
+use DateTimeImmutable;
+use DateTimeZone;
+use Exception;
+
+class Event_Repository_Cached extends Event_Repository {
+	private const CACHE_DURATION    = 60 * 60 * 24; // 24 hours.
```
amieiro:
```suggestion
	private const CACHE_DURATION    = DAY_IN_SECONDS;
```

includes/event/event-repository-interface.php
```diff
@@ -0,0 +1,68 @@
...
+}
+
+class CreateEventFailed extends Exception {}
+class UpdateEventFailed extends Exception {}
+
+interface Event_Repository_Interface {
+	/**
+	 * @throws CreateEventFailed
+	 */
+	public function create_event( Event $event ): void;
```
amieiro: Could we use the same approach that we use in with the posts: `insert_event( Event $event )` to create and update the event and `update_event( Event $event )` to update the event. 

I think it will be interesting that both methods will return `int|WP_Error`, where `int` is the event ID.

And I think we need a method to delete the events: `delete_event( int $eventid, bool $force_delete = false ): Event|false`: 
- Parameters:
  - $eventid: the id of the event to delete.
  - $force_delete: delete the event if it has translations. false by default.
- Returned value: Event|false Event data on success, false on failure.

[wp_insert_post()](https://developer.wordpress.org/reference/functions/wp_insert_post/) and [wp_update_post()](https://developer.wordpress.org/reference/functions/wp_update_post/) documentation.

psrpinto:
Great feedback @amieiro, thank you! I will implement your suggested changes and fix conflicts with trunk.

includes/event/event-repository-interface.php
psrpinto: I have now implemented these suggestions.

Concerning `delete_event()`, for now the function doesn't take the `$force_delete` argument. I didn't do it here because I'm not sure it's something that belongs in the repository as it feels like it's more a business logic thing, so it would be the calling code that needs to check if the event has stats before calling `delete_event()`. That's already the case today, [here](https://github.com/WordPress/wporg-gp-translation-events/blob/trunk/wporg-gp-translation-events.php#L317-L326).

Currently (in trunk) we're not yet deleting stats when the event is deleted, and I would prefer to not add any features in this PR, so that would be another reason to not add the `$force_delete` flag yet. 

If this makes sense to you, I can create a new issue for deleting stats when the event is deleted.

includes/event/event-repository-interface.php
```diff
@@ -0,0 +1,77 @@
...
+use WP_Error;
+
+class EventNotFound extends Exception {
+	public function __construct( Throwable $previous = null ) {
+		parent::__construct( 'Event not found', 0, $previous );
+	}
+}
+
+interface Event_Repository_Interface {
+	/**
```
amieiro, line 9: I suggest adding the explanation of what each function does (not only in this one), as the [wp_insert_post](https://github.com/WordPress/WordPress/blob/9ed35adffa36c706ed5e2642fd91e86c5d011f83/wp-includes/post.php#L4082) function the WordPress core does. Something simple as
```suggestion
	/**
	* Inserts or update an event.
```

includes/event/event-repository-interface.php
```diff
@@ -0,0 +1,77 @@
+<?php
+
+namespace Wporg\TranslationEvents\Event;
+
+use Exception;
+use Throwable;
+use WP_Error;
+
+class EventNotFound extends Exception {
```
amieiro: If you execute the PHPCS with the WordPress standard for this file

`phpcs includes/event/event-repository-interface.php --standard=WordPress`

you will see some errors, like each class should be in its own file. I know it is some extreme to split these small classes in different files, but this is what the WPCS suggests. You can see [here](https://github.com/WordPress/WordPress-Coding-Standards/pull/1802) the PR for this:

> Best practice suggestion: Declare only one class/interface/trait in a file.

psrpinto: There are quite a few [rules that we are disabling](https://github.com/WordPress/wporg-gp-translation-events/blob/trunk/phpcs.xml#L15-L25). I would prefer to not make changes to phpcs in this PR, as that will require changing code that is not introduced by this PR, and in this PR the intention was just to add the repository without making any changes to the existing code.

If we want to change the phpcs rules, my suggestion would be to do that in a PR that only does that, as it will likely result in many changes across the codebase.

psrpinto:
I've now changed `get_event()` so that it returns `null` instead of throwing an `EventNotFound` exception, as returning `null` is more in line with how `get_post()` behaves.

akirk (approved):
Let's go ahead with it, thank you! If we find more we can always iterate some more.

includes/event/event-repository.php
```diff
@@ -0,0 +1,315 @@
...
+			$query_args['post__in'] = $filter_by_ids;
+		}
+
+		return $this->execute_events_query( $query_args );
+	}
+
+	/**
+	 * @throws Exception
+	 */
+	protected function assert_pagination_arguments( int $page, int $page_size ) {
```
amieiro, line 227: As this function is specific for this class (it is not implementing a method from the interface), it will be interesting to add the description to this method and to the other methods that are specific for only the current class.

amieiro (approved):

PR #148, opened 2024-03-04
Fix issue when retrieveing the number of current events the user is attending

### LINKED ISSUE #147: Glopress shows a "contributing to x events" event if user is not attending any events

psrpinto:
Fixes #147 

The cause of the issue is that we're setting `post__in` in the query to an empty array, which results in all events being returned. We have previously fixed a similar issue [here](https://github.com/WordPress/wporg-gp-translation-events/blob/trunk/includes/route.php#L130) (note the `array( 0 )` at the end of the line).

If I may plug my proposal at #145 :), #147 would not have happened if we had a single location in the code that retrieves the events of a user (because we would have fixed it globally when we changed it to `array( 0 )`. Event attendance is a core concept of this plugin, so I think it would make sense to enforce that concept through code, instead of relying on low-level WordPress functions in every part of the codebase that requires dealing with attendance.

### DIFF
--- a/wporg-gp-translation-events.php
+++ b/wporg-gp-translation-events.php
@@ -461,7 +461,11 @@ class Translation_Events {
 	 * @return void
 	 */
 	public function add_active_events_current_user(): void {
-		$user_attending_events      = get_user_meta( get_current_user_id(), Route::USER_META_KEY_ATTENDING, true ) ?: array();
+		$user_attending_events = get_user_meta( get_current_user_id(), Route::USER_META_KEY_ATTENDING, true ) ?: array();
+		if ( empty( $user_attending_events ) ) {
+			return;
+		}
+
 		$current_datetime_utc       = ( new DateTime( 'now', new DateTimeZone( 'UTC' ) ) )->format( 'Y-m-d H:i:s' );
 		$user_attending_events_args = array(
 			'post_type'   => self::CPT,


wporg-gp-translation-events.php
```diff
@@ -461,7 +461,11 @@ public function generate_event_slug( array $data, array $postarr ): array {
 	 * @return void
 	 */
 	public function add_active_events_current_user(): void {
-		$user_attending_events      = get_user_meta( get_current_user_id(), Route::USER_META_KEY_ATTENDING, true ) ?: array();
+		$user_attending_events = get_user_meta( get_current_user_id(), Route::USER_META_KEY_ATTENDING, true ) ?: array();
```
amieiro, line 464: I was reviewing this problem and I saw we solved it with this approach [here](https://github.com/WordPress/wporg-gp-translation-events/blob/9e950955b446931ef738b0dfe6bf435e42b96935/includes/route.php#L130). I think it is simpler. 
```suggestion
		$user_attending_events = get_user_meta( get_current_user_id(), Route::USER_META_KEY_ATTENDING, true ) ?: array( 0 );

```

akirk: I like the solution in this PR better because it is much more evident what is happening. I was already wondering what the `array( 0 )` is doing, it doesn't explain what's going on.

akirk:
Thanks, @psrpinto! I agree that we should proceed with #145, or something like it. On first look it looks quite complex. I'll take a look.

wporg-gp-translation-events.php
psrpinto: I agree with the solution in this PR being more explicit, so I would opt for it instead or setting `array( 0 )`.

amieiro: The `arrray( 0 )` solution is returning an array with and invalid ID (0), because if you pass an empty array to the query, it returns a post :/ 
Let's merge this PR.

amieiro (approved):

PR #151, opened 2024-03-05
Use gp_table_prefix instead of base_prefix

akirk:
GlotPress doesn’t use the WordPress table namespace and has its own prefix. As per @dd32's advice, this switches it over to using that so that on WordPress.org we use a `translate_` table prefix.

In order to get the new table created, you need to deactivate and reactivate the plugin.

### DIFF
--- a/wporg-gp-translation-events.php
+++ b/wporg-gp-translation-events.php
@@ -26,7 +26,7 @@ use WP_Post;
 use WP_Query;
 
 class Translation_Events {
-	const CPT      = 'translation_event';
+	const CPT = 'translation_event';
 
 	public static function get_instance() {
 		static $instance = null;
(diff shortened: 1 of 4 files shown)


trymebytes (approved):
Tested and works on my local machine

wporg-gp-translation-events.php
```diff
@@ -26,7 +26,8 @@
 use WP_Query;
 
 class Translation_Events {
-	const CPT = 'translation_event';
+	const CPT      = 'translation_event';
+	const DB_TABLE = 'translate_event_actions';
```
psrpinto: The `DB_TABLE` const seems to be unused.

```suggestion
```

psrpinto (approved):

wporg-gp-translation-events.php
amieiro: I looked for this `DB_TABLE` constant in the plugin, and it is unused, so I am going to approve this suggestion.

amieiro (approved):
I have tested in my local machine after remove the DB_TABLE constant, and works fine.

PR #153, opened 2024-03-05
Split routes into separate files

psrpinto:
This PR splits routes into separate files, instead of having a single `route.php` with all the routes. The logic hasn't been modified in any way, I just cut-pasted code from one file to another.

Screenshot 2024-03-05 at 15 19 13


akirk (approved):
Nice change! Tested it and it worked well

PR #154, opened 2024-03-06
Handle exceptions

### LINKED ISSUE #129: Review Code regarding catchable Exceptions
Let's do a code reivew at where we throw exceptions or where they could be thrown (see #71) and that we handle them.
- [ ] [wporg-gp-translation-events.php](https://github.com/WordPress/wporg-gp-translation-events/blob/trunk/wporg-gp-translation-events.php)
- [ ] ...

### LINKED ISSUE #130: Review Code regarding notices and warnings
We want to make sure that our code doesn't throw notices and warnings. Please review your local logfile to see if there is a notice or warning that occurred that can be fixed.

psrpinto:
Fixes #129 #130 

This PR handles all exceptions. I fixed some other things along the way because I was using the IDE to look for unhandled exceptions, and the IDE showed other things that could be fixed, so I took the opportunity to fix them as well.

I also tested everything and made sure there were no notices/warnings in the log, so #130 can also be considered fixed.

### DIFF
--- a/includes/routes/user/my-events.php
+++ b/includes/routes/user/my-events.php
@@ -4,6 +4,7 @@ namespace Wporg\TranslationEvents\Routes\User;
 
 use DateTime;
 use DateTimeZone;
+use Exception;
 use WP_Query;
 use Wporg\TranslationEvents\Routes\Route;
 use Wporg\TranslationEvents\Translation_Events;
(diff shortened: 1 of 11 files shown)


includes/routes/user/my-events.php
```diff
@@ -38,11 +39,23 @@ public function handle(): void {
...
-		$args                 = array(
+		$user_id = get_current_user_id();
+		$events  = get_user_meta( $user_id, Translation_Events::USER_META_KEY_ATTENDING, true ) ?: array();
+		$events  = array_keys( $events );
+
+		$current_datetime_utc = '';
+		try {
+			$current_datetime_utc = ( new DateTime( 'now', new DateTimeZone( 'UTC' ) ) )->format( 'Y-m-d H:i:s' );
+		} catch ( Exception $e ) {
+			// This can't happen because the DateTime constructor takes 'now' as argument so can't fail, neither can
```
akirk, line 50: Then let's remove this try catch?

akirk:
Thanks for this! I started #157 this morning and I think that maybe it could handle those exceptions on DateTime better, what do you think?

psrpinto:
Yep, #157 makes a lot of sense. I'm putting this back in draft and will revisit once #157 is merged.

psrpinto:
Closing this one since too many things have changed since it was opened, and it's probably not relevant any more after #157.

PR #157, opened 2024-03-07
Improve event dates

akirk:
This adds a new `Event_Date` class that tries to make handling dates more unified. It also adds relative JavaScript time formatting:

Screenshot 2024-03-07 at 11 08 06

Screenshot 2024-03-07 at 11 08 06

### DIFF
--- a/includes/event-date.php
+++ b/includes/event-date.php
@@ -62,7 +62,6 @@ abstract class Event_Date extends DateTimeImmutable {
 			return sprintf( 'starts %s', $this->format( 'l, F j, Y' ) );
 		}
 
-
 		$interval       = $this->diff( $current_date_time );
 		$hours_left     = ( $interval->d * 24 ) + $interval->h;
 		$hours_in_a_day = 24;
--- a/includes/event-date.php
+++ b/includes/event-date.php
@@ -15,8 +15,13 @@ abstract class Event_Date extends DateTimeImmutable {
 			$timezone = new DateTimeZone( 'UTC' );
 		}
 
-		$utc_date = new DateTime( $date, new DateTimeZone( 'UTC' ) );
-		$utc_date->setTimezone( $timezone );
+		try {
+			$utc_date = new DateTime( $date, new DateTimeZone( 'UTC' ) );
+			$utc_date->setTimezone( $timezone );
+		} catch ( Exception $e ) {
+			$utc_date = new DateTime( 'now', new DateTimeZone( 'UTC' ) );
+		}
+
 		parent::__construct( $utc_date->format( 'Y-m-d H:i:s' ), $timezone );
 		$this->timezone = $timezone;
 	}
--- a/includes/event-date.php
+++ b/includes/event-date.php
@@ -7,9 +7,15 @@ use DateTimeImmutable;
 use DateTimeZone;
 use Exception;
 
+/**
+ * class Event_Date
+ *
+ * The event date is in local time, get the UTC time via the utc() method.
+ *
+ * @package Wporg\TranslationEvents
+ */
 abstract class Event_Date extends DateTimeImmutable {
-	abstract public function is_end_date(): bool;
-	public $timezone;
+	protected $event_timezone;
 	public function __construct( string $date, DateTimeZone $timezone = null ) {
 		if ( ! $timezone ) {
 			$timezone = new DateTimeZone( 'UTC' );
@@ -23,7 +29,11 @@ abstract class Event_Date extends DateTimeImmutable {
 		}
 
 		parent::__construct( $utc_date->format( 'Y-m-d H:i:s' ), $timezone );
-		$this->timezone = $timezone;
+		$this->event_timezone = $timezone;
+	}
+
+	public function timezone() {
+		return $this->event_timezone;
 	}
 
 	/**
@@ -32,20 +42,20 @@ abstract class Event_Date extends DateTimeImmutable {
 	 * @return string The date text.
 	 */
 	public function __toString(): string {
-		return $this->utc( 'Y-m-d H:i:s' );
+		return $this->utc()->format( 'Y-m-d H:i:s' );
 	}
 	/**
 	 * Get the local formatted text for the date in UTC.
 	 *
-	 * @return string The date text.
+	 * @return DateTimeImmutable The date text.
 	 */
-	public function utc( $format ): string {
-		return $this->setTimeZone( new DateTimeZone( 'UTC' ) )->format( $format );
+	public function utc(): DateTimeImmutable {
+		return $this->setTimeZone( new DateTimeZone( 'UTC' ) );
 	}
 
 	public function is_in_the_past() {
 		$current_date_time = new DateTimeImmutable( 'now', new DateTimeZone( 'UTC' ) );
-		return $this < $current_date_time;
+		return $this->utc() < $current_date_time;
 	}
 
 	/**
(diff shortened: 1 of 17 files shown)


includes/event-date.php
```diff
@@ -0,0 +1,78 @@
...
+	public function utc( $format ): string {
+		return $this->setTimeZone( new DateTimeZone( 'UTC' ) )->format( $format );
+	}
+
+	/**
+	 * Generate variable text depending on when the event starts or ends.
+	 *
+	 * @return string The end date text.
+	 */
+	public static function get_variable_text(): string {
```
psrpinto: This function doesn't seem to be called anywhere. I'm guessing maybe this isn't needed anymore?

includes/routes/event/edit.php
```diff
@@ -71,6 +73,5 @@ public function handle( int $event_id ): void {
 	 * @throws Exception When date is invalid.
 	 */
 	private static function convertToTimezone( string $date_time, string $time_zone ): string {
```
psrpinto: Should this function be removed? It doesn't have a body and doesn't seem to be called anywhere.

psrpinto (commented):
This is great! Representing start and end as objects makes a lot of sense 👍 

I noticed some things though. I'm wondering if maybe this PR was meant to be draft and I'm reviewing too early :)

Also took the liberty of fixing formatting issues that were triggering phpcs.

akirk:
@psrpinto thanks for the review! You are right, this should have still been a draft. I'll set it back to that and will continue tomorrow morning.

includes/event-date.php
akirk: It is now used

includes/routes/event/edit.php
akirk: I removed it, thanks!

trymebytes (approved):
Tested and works for me!

I want to suggest that we make the date and time formats consistent, see screenshot below;
Screenshot 2024-03-14 at 07 07 27

We could do this in a follow up PR.

includes/event-date.php
```diff
@@ -0,0 +1,89 @@
...
+namespace Wporg\TranslationEvents;
+
+use DateTime;
+use DateTimeImmutable;
+use DateTimeZone;
+use Exception;
+
+abstract class Event_Date extends DateTimeImmutable {
+	abstract public function is_end_date(): bool;
+	public $timezone;
```
psrpinto: I think it would be best to make this private to prevent it from being modified from outside this class. Currently it's only being called in tests, so we could add a getter that the test would call.

includes/event-date.php
```diff
@@ -0,0 +1,89 @@
...
+	public function utc( $format ): string {
+		return $this->setTimeZone( new DateTimeZone( 'UTC' ) )->format( $format );
+	}
+
+	/**
+	 * Generate variable text depending on when the event starts or ends.
+	 *
+	 * @return string The end date text.
+	 */
+	public function get_variable_text(): string {
```
psrpinto: Since this is about presentation (displaying to the user), I think that this function could be moved out of this class into `templates/`, either directly in `events-list.php` (as it's currently only being called there), or if we expect it to be called from elsewhere in the future, in `helper-functions.php`.

Doing this would remove the need to have the `is_end_date()` method in the `Event_Date` class. It feels a bit weird that a start date would know whether it's an end date (it's a start date, so it should not be possible for it to be an end date), and moving the `get_variable_text()` out of this class would remove the need for the `is_end_date()` function, I think.

This would also probably mean that there is no longer the need to have both an `Event_Start_Date` and a `Event_End_Date` classes, there would only be an `Event_Date`.

To clarify, when we call `get_variable_text()`, we know whether we're calling it on an end date or not (`$event_end->get_variable_text()`), so `get_variable_text()` could take a boolean as argument that tells whether it's an end date or not: `get_variable_text( bool $is_end_date )`.

psrpinto (approved):
This is great!

I found a couple of things that could be changed, but this is already better than it was, so ok from me to merge :)

psrpinto:
@akirk If you want I can pick this one up and resolve conflicts, so we can get it merged.

akirk:
I'm on it.

akirk:
This is now ready for review again. I replaced all displayed time strings with a `print_(relative_)time_html` which output the time in a `<time>` HTML that gets updated via JS every 10 seconds. When it flips over from a future time to a past time, it reloads the page. To illustrate, here is a GIF of this happening:
auto-reload

wporg-gp-translation-events.php
```diff
@@ -53,6 +53,8 @@ public function __construct() {
 
 	public function gp_init() {
 		require_once __DIR__ . '/templates/helper-functions.php';
+		require_once __DIR__ . '/includes/active-events-cache.php';
```
akirk: I am not sure whether this was a rebase issue or why this is added here.

akirk:
Sorry about the CSS formatting changes, that was my editor since I added a rule to the CSS.

wporg-gp-translation-events.php
psrpinto: It's being included below, can be removed. I guess it was moved to below on trunk since this branch was created.

akirk: Ah indeed, thanks!

templates/events-my-events.php
```diff
@@ -33,26 +33,24 @@
...
+			$event_start                   = new Event_Start_Date( get_post_meta( get_the_ID(), '_event_start', true ) );
+			$event_end                     = new Event_End_Date( get_post_meta( get_the_ID(), '_event_end', true ) );
 			$stats_calculator              = new Stats_Calculator();
 			$has_stats                     = $stats_calculator->event_has_stats( get_post() );
 
 			?>
 			<li class="event-list-item">
 				<a class="event-link-<?php echo esc_attr( $event_status ); ?>" href="<?php echo esc_url( $event_url ); ?>"><?php the_title(); ?></a>
-				<?php if ( $event_end_utc > $now && ! $has_stats ) : ?>
+				<?php if ( $event_end->is_in_the_past() && ! $has_stats ) : ?>
```
amieiro: The event can't be a past event to be able to edit it.
```suggestion
				<?php if ( ! $event_end->is_in_the_past() && ! $has_stats ) : ?>
```

includes/event-date.php
```diff
@@ -0,0 +1,173 @@
+<?php
+
+namespace Wporg\TranslationEvents;
```
psrpinto, line 3: We could move this to `includes/event` since this is related to an event (namespace would then be `Wporg\TranslationEvents\Event`)

includes/event-date.php
```diff
@@ -0,0 +1,173 @@
...
+	public function __construct( string $date, DateTimeZone $timezone = null ) {
+		if ( ! $timezone ) {
+			$timezone = new DateTimeZone( 'UTC' );
+		}
+
+		try {
+			$utc_date = new DateTime( $date, new DateTimeZone( 'UTC' ) );
+			$utc_date->setTimezone( $timezone );
+		} catch ( Exception $e ) {
+			$utc_date = new DateTime( 'now', new DateTimeZone( 'UTC' ) );
```
psrpinto, line 28: I think maybe we should let it throw, otherwise this could result in an event being created/updated with a date that was not what the user supplied.

amieiro:
The event creation is broken 

image

You need to add this lines in the `includes/routes/create.php` file:

In the header:

```
use DateTimeZone;
use Wporg\TranslationEvents\Event_End_Date;
use Wporg\TranslationEvents\Event_Start_Date;
```

Before calling the template (`$this->tmpl( 'events-form', get_defined_vars() );`):

```
		$event_timezone           = wp_timezone() ?: new DateTimeZone( 'UTC' );
		$event_start              = new Event_Start_Date( date_i18n( 'Y-m-d H:i' ) );
		$event_end                = new Event_End_Date( date_i18n( 'Y-m-d H:i' ) );
```

psrpinto (approved):
Code looks good, well done 👌 

I also had the issue Jesús reported above.

includes/event-date.php
akirk: Indeed but if we do it, let's do it in a follow up. Would be nice to get this one over the finish line.

includes/event-date.php
akirk: Also follow up material, I think.

templates/events-my-events.php
akirk: Nice thanks! I think I might had disabled it during debug. I think it's easier on the eye than a comparison with `$now`.

akirk:
Thanks everyone for the reviews, I fixed this and merged it.

PR #159, opened 2024-03-07
Display language names instead of locale slugs

akirk:
Instead of the list of locales:
Screenshot 2024-03-07 at 13 28 16
show the language names:
Screenshot 2024-03-07 at 13 28 27

Also in the text:
Screenshot 2024-03-07 at 13 33 06

### DIFF
--- a/templates/event.php
+++ b/templates/event.php
@@ -102,11 +102,8 @@ gp_tmpl_load( 'events-header', get_defined_vars(), __DIR__ );
 				)
 			);
 			?>
-									</p>
+			</p>
 	</details>
-	<div class="event-contributors">
-		<h2><?php esc_html_e( 'Contributors', 'gp-translation-events' ); ?></h2>
-	</div>
 	<?php endif ?>
 	</div>
 	<div class="event-details-right">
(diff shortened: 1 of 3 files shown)


templates/event.php
```diff
@@ -58,9 +58,9 @@
 			</thead>
 			<tbody>
 			<?php /** @var $row Stats_Row */ ?>
-			<?php foreach ( $event_stats->rows() as $locale_ => $row ) : ?>
+			<?php foreach ( $event_stats->rows() as $_locale => $row ) : ?>
 			<tr>
-				<td><?php echo esc_html( $locale_ ); ?></td>
+				<td title="<?php echo esc_html( $_locale ); ?> "><a href="<?php echo esc_url( gp_url( '/locale/' . $row->language->slug ) ); ?>"><?php echo esc_html( $row->language->english_name ); ?></a></td>
```
trymebytes: Local GlotPress uses `'/languages/'` else a 404 is returned when the locale link is clicked.

trymebytes (approved):
Tested and works for me!

PR #164, opened 2024-03-07
Show the projects that were contributed to

akirk:
Screenshot 2024-03-07 at 15 41 39

### DIFF
--- a/includes/stats-calculator.php
+++ b/includes/stats-calculator.php
@@ -148,7 +148,7 @@ class Stats_Calculator {
 	/**
 	 * Get contributors for an event.
 	 */
-	public function get_contributors( WP_Post $event ): array {
+	public function get_contributors( int $event_id ): array {
 		global $wpdb, $gp_table_prefix;
 
 		// phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
@@ -165,7 +165,7 @@ class Stats_Calculator {
 				group by user_id
 			",
 				array(
-					$event->ID,
+					$event_id,
 				)
 			)
 		);
@@ -191,7 +191,7 @@ class Stats_Calculator {
 	/**
 	 * Get projects for an event.
 	 */
-	public function get_projects( WP_Post $event ): array {
+	public function get_projects( int $event_id ): array {
 		global $wpdb, $gp_table_prefix;
 
 		// phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
@@ -213,7 +213,7 @@ class Stats_Calculator {
 				group by o.project_id
 			",
 				array(
-					$event->ID,
+					$event_id,
 				)
 			)
 		);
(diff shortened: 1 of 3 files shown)


includes/stats-calculator.php
```diff
@@ -187,6 +188,55 @@ function ( $a, $b ) {
 		return $users;
 	}
 
+	/**
+	 * Get projects for an event.
+	 */
+	public function get_projects( WP_Post $event ): array {
```
psrpinto: Instead of passing the whole object here, we could pass just the id as the function only needs the id.

templates/event.php
```diff
@@ -79,13 +79,30 @@
 		<h2><?php esc_html_e( 'Contributors', 'gp-translation-events' ); ?></h2>
 		<ul>
 			<?php foreach ( $contributors as $contributor ) : ?>
-			<li class="event-contributor" title="<?php echo esc_html( implode( ', ', $contributor->locales ) ); ?>"
+			<li class="event-contributor" title="<?php echo esc_html( implode( ', ', $contributor->locales ) ); ?>">
```
psrpinto, line 82: Nice bugfix :) I'm surprised it was working correctly due to the HTML being malformed.

psrpinto (approved):
Tested and works nicely 👍

templates/event.php
```diff
@@ -79,13 +79,30 @@
...
+			<?php foreach ( $projects as $project_name => $row ) : ?>
+			<li class="event-project" title="<?php echo esc_html( str_replace( ',', ', ', $row->locales ) ); ?>">
+				<a href="<?php echo esc_url( gp_url_project( $row->project ) ); ?>"><?php echo esc_html( $project_name ); ?></a> <small> to
+				<?php
+				foreach ( explode( ',', $row->locales ) as $_locale ) {
+					$_locale = \GP_Locales::by_slug( $_locale );
+					?>
+					<a href="<?php echo esc_url( gp_url_project_locale( $row->project, $_locale, 'default' ) ); ?>"><?php echo esc_html( $_locale->english_name ); ?></a>
+				<?php } ?>
+				by <?php echo esc_html( $row->users ); ?> contributors</small>
```
trymebytes, line 101:
```suggestion
				by
				<?php
				// translators: %d: Number of contributors.
				printf( _n( '%d contributor', '%d contributors', $row->users, 'gp-translation-events' ), esc_html( $row->users ) );
				?>
```

trymebytes:
```suggestion
				by 
				<?php
				// translators: %d: Number of contributors.
				printf( _n( '%d contributor', '%d contributors', $row->users, 'gp-translation-events' ), esc_html( $row->users ) );
				?>
```

trymebytes: Just a tiny suggestion for singular and plurals.

trymebytes (commented):
Tested and works as expected!

includes/stats-calculator.php
akirk: Thanks

templates/event.php
akirk: Thanks for the suggestion, we should just use the `esc_html()` on the outside and `sprintf` in the inside to make sure all of it is escaped.

Issue #167, opened 2024-03-08
Store attendance for an event


akirk:
We currently don't know who attends an event since we store it only on the user side.

PR #171, opened 2024-03-11
Prepare for storing attendees of an event

psrpinto:
This PR is the first step to tackle #167.

It makes it so that the fact that attendance is stored as user meta is not exposed to the codebase. This is achieved through an `Attendee_Repository`.

This will make it possible to change how attendance is stored without impacting the rest of the codebase. Only the internals of `Attendee_Repository` will need to change. I will be doing that in a follow-up PR.

### DIFF
--- a/includes/attendee-repository.php
+++ b/includes/attendee-repository.php
@@ -1,0 +1,7 @@
+<?php
+
+namespace Wporg\TranslationEvents;
+
+class Attendee_Repository {
+
+}
--- a/includes/attendee-repository.php
+++ b/includes/attendee-repository.php
@@ -2,11 +2,28 @@
 
 namespace Wporg\TranslationEvents;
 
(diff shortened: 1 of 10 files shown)


includes/attendee-repository.php
```diff
@@ -0,0 +1,86 @@
...
+		}
+
+		return isset( $event_ids[ $event_id ] );
+	}
+
+	/**
+	 * @return int[] User ids.
+	 */
+	public function get_attendees( int $event_id ): array { // phpcs:ignore Generic.CodeAnalysis.UnusedFunctionParameter.Found
+		// TODO.
```
psrpinto, line 65: This will be done in a follow-up PR. It's not currently needed.

trymebytes (approved):
Tested and works as expected

PR #172, opened 2024-03-12
Store attendance in an attendees table

### LINKED ISSUE #167: Store attendance for an event

psrpinto:
Fixes #167 

I would recommend reviewing the final result instead of individual commits, as some iteration was required.

## Summary of changes

- Extract all `require_once` calls to an `autoload.php` file
    - And load it early instead of in `gp_init()`, to guarantee required files are always loaded
- Extract database upgrade logic to a separate file
- Move `Attendee_Repository` to an `Attendee` namespace
- Create `Attendee` class which represents a user attending an event
- Adapt `Attendee_Repository` so it uses `Attendee` class
- Add missing methods in `Attendee_Repository` (e.g. `get_attendee()`)
- Create `attendees` table, with the following fields
    - `translate_event_attendees_id` (primary key)
    - `event_id`: `int(10)`
    - `user_id`: `int(10)`
    - `is_host`: `tinyint(1)` (bool)

### DIFF
--- a/includes/attendee-repository.php
+++ b/includes/attendee-repository.php
@@ -104,11 +104,30 @@ class Attendee_Repository {
 	 * @return int[] Event ids.
 	 */
 	public function get_events_for_user( int $user_id ): array {
-		$event_ids = get_user_meta( $user_id, self::USER_META_KEY, true );
-		if ( ! $event_ids ) {
-			$event_ids = array();
-		}
+		global $wpdb, $gp_table_prefix;
 
-		return array_keys( $event_ids );
+		// phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.NoCaching
+		$rows = $wpdb->get_results(
+			$wpdb->prepare(
+				"
+				select event_id
+				from {$gp_table_prefix}event_attendees
+				where user_id = %d
+			",
+				array(
+					$user_id,
+				)
+			)
+		);
+		// phpcs:enable
+
+		return array_map(
+			function ( object $row ) {
+				return intval( $row->event_id );
+			},
+			$rows
+		);
 	}
 }
--- a/includes/attendee-repository.php
+++ b/includes/attendee-repository.php
@@ -69,7 +69,7 @@ class Attendee_Repository {
 	/**
 	 * @deprecated
 	 * TODO: This method should be moved out of this class because it's not about attendance,
-	 *       it returns events that match a condition (belong to a user), so it belongs in an event repository.
+	 *       it returns events that match a condition (have a user as attendee), so it belongs in an event repository.
 	 *       However, since we don't have an event repository yet, the method is placed here for now.
 	 *       When the method is moved to an event repository, it should return Event instances instead of event ids.
 	 *
--- a/includes/attendee-repository.php
+++ b/includes/attendee-repository.php
@@ -5,8 +5,6 @@ namespace Wporg\TranslationEvents;
 use Exception;
 
 class Attendee_Repository {
-	private const USER_META_KEY = 'translation-events-attending';
-
 	/**
 	 * @throws Exception
 	 */
--- a/includes/attendee-repository.php
+++ b/includes/attendee-repository.php
@@ -45,16 +45,20 @@ class Attendee_Repository {
 			throw new Exception( 'invalid user id' );
 		}
 
-		$event_ids = get_user_meta( $user_id, self::USER_META_KEY, true );
-		if ( ! $event_ids ) {
-			$event_ids = array();
-		}
-
-		if ( isset( $event_ids[ $event_id ] ) ) {
-			unset( $event_ids[ $event_id ] );
-		}
-
-		update_user_meta( $user_id, self::USER_META_KEY, $event_ids );
+		global $wpdb, $gp_table_prefix;
+		// phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.NoCaching
+		$wpdb->query(
+			$wpdb->prepare(
+				"delete from {$gp_table_prefix}event_attendees where event_id = %d and user_id = %d",
+				array(
+					'event_id' => $event_id,
+					'user_id'  => $user_id,
+				),
+			),
+		);
+		// phpcs:enable
 	}
 
 	public function is_attending( int $event_id, int $user_id ): bool {
--- a/includes/attendee-repository.php
+++ b/includes/attendee-repository.php
@@ -50,12 +50,28 @@ class Attendee_Repository {
 	}
 
 	public function is_attending( int $event_id, int $user_id ): bool {
-		$event_ids = get_user_meta( $user_id, self::USER_META_KEY, true );
-		if ( ! $event_ids ) {
-			$event_ids = array();
-		}
+		global $wpdb, $gp_table_prefix;
 
-		return isset( $event_ids[ $event_id ] );
+		// phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.NoCaching
+		$row = $wpdb->get_row(
+			$wpdb->prepare(
+				"
+				select count(*) as cnt
+				from {$gp_table_prefix}event_attendees
+				where event_id = %d
+				  and user_id = %d
+			",
+				array(
+					$event_id,
+					$user_id,
+				)
+			)
+		);
+		// phpcs:enable
+
+		return 1 === intval( $row->cnt );
 	}
 
 	/**
--- a/includes/attendee-repository.php
+++ b/includes/attendee-repository.php
@@ -44,16 +44,17 @@ class Attendee_Repository {
 		}
 
 		global $wpdb, $gp_table_prefix;
-		// phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
 		// phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery
 		// phpcs:disable WordPress.DB.DirectDatabaseQuery.NoCaching
-		$wpdb->query(
-			$wpdb->prepare(
-				"delete from {$gp_table_prefix}event_attendees where event_id = %d and user_id = %d",
-				array(
-					'event_id' => $event_id,
-					'user_id'  => $user_id,
-				),
+		$wpdb->delete(
+			"{$gp_table_prefix}event_attendees",
+			array(
+				'event_id' => $event_id,
+				'user_id'  => $user_id,
+			),
+			array(
+				'%d',
+				'%d',
 			),
 		);
 		// phpcs:enable
--- a/includes/attendee-repository.php
+++ b/includes/attendee-repository.php
@@ -18,12 +18,20 @@ class Attendee_Repository {
 			throw new Exception( 'invalid user id' );
 		}
 
-		$event_ids = get_user_meta( $user_id, self::USER_META_KEY, true );
-		if ( ! $event_ids ) {
-			$event_ids = array();
-		}
-		$event_ids[ $event_id ] = true;
-		update_user_meta( $user_id, self::USER_META_KEY, $event_ids );
+		global $wpdb, $gp_table_prefix;
+		// phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.NoCaching
+		$wpdb->query(
+			$wpdb->prepare(
+				"insert ignore into {$gp_table_prefix}event_attendees (event_id, user_id) values (%d, %d)",
+				array(
+					'event_id' => $event_id,
+					'user_id'  => $user_id,
+				),
+			),
+		);
+		// phpcs:enable
 	}
 
 	/**
--- a/includes/database.php
+++ b/includes/database.php
@@ -75,7 +75,7 @@ class Database {
 		foreach ( $events as $event ) {
 			foreach ( $stats_calculator->get_contributors( $event->ID ) as $user ) {
 				$attendee = new Attendee( $event->ID, $user->ID );
-				$attendee_repository->add_attendee( $attendee );
+				$attendee_repository->insert_attendee( $attendee );
 			}
 		}
 	}
--- a/includes/database.php
+++ b/includes/database.php
@@ -74,7 +74,6 @@ class Database {
 		foreach ( $events as $event ) {
 			foreach ( $stats_calculator->get_contributors( $event->ID ) as $user ) {
 				$attendee = new Attendee( $event->ID, $user->ID );
-				$attendee->mark_as_host();
 				$attendee_repository->insert_attendee( $attendee );
 			}
 		}
--- a/includes/database.php
+++ b/includes/database.php
@@ -74,6 +74,7 @@ class Database {
 		foreach ( $events as $event ) {
 			foreach ( $stats_calculator->get_contributors( $event->ID ) as $user ) {
 				$attendee = new Attendee( $event->ID, $user->ID );
+				$attendee->mark_as_host();
 				$attendee_repository->insert_attendee( $attendee );
 			}
 		}
--- a/includes/database.php
+++ b/includes/database.php
@@ -32,6 +32,8 @@ class Database {
 			`translate_event_attendees_id` int(11) unsigned NOT NULL AUTO_INCREMENT,
 			`event_id` int(10) NOT NULL COMMENT 'Post_ID of the translation_event post in the wp_posts table',
 			`user_id` int(10) NOT NULL COMMENT 'ID of the user who is attending the event',
+			`is_host` tinyint(1) default 0 not null comment 'Whether the user is a host of the event',
+			`is_contributor` tinyint(1) default 0 not null comment 'Whether the user created or reviewed translations during the event',
 		PRIMARY KEY (`translate_event_attendees_id`),
 		UNIQUE KEY `event_per_user` (`event_id`,`user_id`),
 		INDEX `user` (`user_id`)
--- a/includes/database.php
+++ b/includes/database.php
@@ -1,0 +1,80 @@
+<?php
+
+namespace Wporg\TranslationEvents;
+
+use Exception;
+use WP_Query;
+use Wporg\TranslationEvents\Attendee\Attendee;
+
+class Database {
+	public static function upgrade(): void {
+		require_once ABSPATH . 'wp-admin/includes/upgrade.php';
+		global $gp_table_prefix;
+
+		// Create actions table.
+		$create_actions_table = "
+		CREATE TABLE `{$gp_table_prefix}event_actions` (
+			`translate_event_actions_id` int(11) unsigned NOT NULL AUTO_INCREMENT,
+			`event_id` int(10) NOT NULL COMMENT 'Post_ID of the translation_event post in the wp_posts table',
+			`original_id` int(10) NOT NULL COMMENT 'ID of the translation',
+			`user_id` int(10) NOT NULL COMMENT 'ID of the user who made the action',
+			`action` enum('approve','create','reject','request_changes') NOT NULL COMMENT 'The action that the user made (create, reject, etc)',
+			`locale` varchar(10) CHARACTER SET latin1 COLLATE latin1_swedish_ci NOT NULL COMMENT 'Locale of the translation',
+			`happened_at` datetime NOT NULL COMMENT 'When the action happened, in UTC',
+		PRIMARY KEY (`translate_event_actions_id`),
+		UNIQUE KEY `event_per_translated_original_per_user` (`event_id`,`locale`,`original_id`,`user_id`)
+		) COMMENT='Tracks translation actions that happened during a translation event'";
+		dbDelta( $create_actions_table );
+
+		// Create attendees table.
+		$create_attendees_table = "
+		CREATE TABLE `{$gp_table_prefix}event_attendees` (
+			`translate_event_attendees_id` int(11) unsigned NOT NULL AUTO_INCREMENT,
+			`event_id` int(10) NOT NULL COMMENT 'Post_ID of the translation_event post in the wp_posts table',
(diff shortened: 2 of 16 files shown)


wporg-gp-translation-events.php
```diff
@@ -97,6 +97,18 @@ public function activate(): void {
...
+
+		// Create attendees table.
+		$create_attendees_table = "
+		CREATE TABLE `{$gp_table_prefix}event_attendees` (
+			`translate_event_attendees_id` int(11) unsigned NOT NULL AUTO_INCREMENT,
+			`event_id` int(10) NOT NULL COMMENT 'Post_ID of the translation_event post in the wp_posts table',
+			`user_id` int(10) NOT NULL COMMENT 'ID of the user who is attending the event',
+		PRIMARY KEY (`translate_event_attendees_id`),
+		UNIQUE KEY `event_per_user` (`event_id`,`user_id`),
+		INDEX `user` (`user_id`)
```
psrpinto: Should the indexes be namesapaced to translation-events, or is this fine?

wporg-gp-translation-events.php
```diff
@@ -97,6 +97,18 @@ public function activate(): void {
 		UNIQUE KEY `event_per_translated_original_per_user` (`event_id`,`locale`,`original_id`,`user_id`)
 		) COMMENT='Tracks translation actions that happened during a translation event'";
 		dbDelta( $create_actions_table );
+
+		// Create attendees table.
+		$create_attendees_table = "
+		CREATE TABLE `{$gp_table_prefix}event_attendees` (
+			`translate_event_attendees_id` int(11) unsigned NOT NULL AUTO_INCREMENT,
```
psrpinto: The primary key is technically not needed but since the actions table above has it (and it's also not technically needed), I guessed I should probably add it here too.

wporg-gp-translation-events.php
```diff
@@ -109,6 +109,60 @@ public function activate(): void {
...
+
+		// Delete legacy attendance information.
+		global $wpdb, $wp_table_prefix;
+		// phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.NoCaching
+		// phpcs:disable WordPress.DB.SlowDBQuery.slow_db_query_meta_key
+		$wpdb->query(
+			$wpdb->prepare(
+				"delete from {$wp_table_prefix}usermeta where meta_key = %s",
```
psrpinto: This doesn't seem to be working somehow. Before spending time debugging it, I figured I should first confirm that we want to do this. We can always leave the meta behind, it won't have any impacts.

Let me know how to proceed.

wporg-gp-translation-events.php
akirk: Index names are just informative, I guess you'd only the name when you'd use `USE INDEX` which is a bit of a bad sign of too much table complexity.

akirk:
Since we need a database upgrade, we need to be careful with errors. What if we added upgrade functionality like GlotPress with a version number? https://github.com/GlotPress/GlotPress/blob/develop/gp-settings.php#L137-L142

Otherwise what is your recommendation for the upgrade? Deactivate the plugin, update the code, activate it again?

includes/attendee-repository.php
```diff
@@ -37,25 +43,45 @@ public function remove_attendee( int $event_id, int $user_id ): void {
...
-			unset( $event_ids[ $event_id ] );
-		}
-
-		update_user_meta( $user_id, self::USER_META_KEY, $event_ids );
+		global $wpdb, $gp_table_prefix;
+		// phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.NoCaching
+		$wpdb->query(
+			$wpdb->prepare(
```
akirk: Could we use `$wpdb->delete()`? Then we don't need all the above disables.

includes/attendee-repository.php
```diff
@@ -18,12 +16,20 @@ public function add_attendee( int $event_id, int $user_id ): void {
...
-		}
-		$event_ids[ $event_id ] = true;
-		update_user_meta( $user_id, self::USER_META_KEY, $event_ids );
+		global $wpdb, $gp_table_prefix;
+		// phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.NoCaching
+		$wpdb->query(
+			$wpdb->prepare(
+				"insert ignore into {$gp_table_prefix}event_attendees (event_id, user_id) values (%d, %d)",
```
akirk: Why don't we use `$wpdb->insert()` here? Then we wouldn't need the disables from above.

includes/attendee-repository.php
```diff
@@ -37,25 +43,45 @@ public function remove_attendee( int $event_id, int $user_id ): void {
...
 	public function is_attending( int $event_id, int $user_id ): bool {
-		$event_ids = get_user_meta( $user_id, self::USER_META_KEY, true );
-		if ( ! $event_ids ) {
-			$event_ids = array();
-		}
+		global $wpdb, $gp_table_prefix;
+
+		// phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.NoCaching
```
akirk: Could you check again if these are needed?

includes/attendee-repository.php
```diff
@@ -69,18 +95,37 @@ public function get_attendees( int $event_id ): array { // phpcs:ignore Generic.
...
 	public function get_events_for_user( int $user_id ): array {
-		$event_ids = get_user_meta( $user_id, self::USER_META_KEY, true );
-		if ( ! $event_ids ) {
-			$event_ids = array();
-		}
+		global $wpdb, $gp_table_prefix;
+
+		// phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.NoCaching
```
akirk: Also here I am not sure why we need these, does phpcs really complain?

tests/attendee-repository.php
```diff
@@ -91,4 +88,13 @@ public function test_get_event_ids() {
 		$this->assertEquals( array( $event3_id ), $this->repository->get_events_for_user( $another_user ) );
 		$this->assertEmpty( $this->repository->get_events_for_user( $another_user + 1 ) );
 	}
+
+	private function all_table_rows(): array {
+		global $wpdb, $gp_table_prefix;
+		// phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.NoCaching
```
akirk: Check these also please.

wporg-gp-translation-events.php
akirk: Let's leave the meta behind and remove this.

akirk (commented):
Functionally this works well. I wonder about all of the phpcs:disables and whether we really need them. I'd try to avoid them if possible and use `wpdb` functions.

akirk:
I've been thinking about this. Should we rename the table also since our rename plan from attendees to RSVP? #176

psrpinto:
I agree with all your remarks/comments, will update accordingly.

> Otherwise what is your recommendation for the upgrade? Deactivate the plugin, update the code, activate it again?
> What if we added upgrade functionality like GlotPress with a version number? 

Indeed I guess we would need to do something like that. I will look into the GlotPress's upgrade functionality with version number as you suggested.

includes/attendee-repository.php
psrpinto: We're now using `$wpdb->delete()`. It allows for removing `WordPress.DB.PreparedSQL.InterpolatedNotPrepared` but not the other ones.

includes/attendee-repository.php
psrpinto: Since we're doing an `insert ignore`, I looked for a way to do insert ignore with `$wpdb->insert()` but it doesn't seem to be possible. Should I leave as is, or should we first delete and then `$wpdb->insert()`?

psrpinto: Or we could throw an exception if we're trying to insert an attendee that already exists.

akirk:
I don't recall where I read @amieiro saying this but I agree that we should not rename the table to rsvp. But I do think we should introduce the status fields as an enum with `rsvp`, `contributed`, `cohost` and we'll keep it at `rsvp` in this iteration.

amieiro:
@akirk I commented this [here](https://github.com/WordPress/wporg-gp-translation-events/issues/176#issuecomment-2011783729).

includes/attendee-repository.php
psrpinto: I have confirmed that all these disables are needed, also the ones you mentioned in other comments.

The `InterpolatedNotPrepared` is because of the table name interpolation.

includes/attendee-repository.php
akirk: An `INSERT IGNORE` is just an insert that doesn't return an error if it couldn't insert a row because it already existed. So if we just ignore the return value then just using `$wpdb->insert()` should be fine?

wporg-gp-translation-events.php
```diff
@@ -96,8 +99,59 @@ public function activate() {
...
+		PRIMARY KEY (`translate_event_attendees_id`),
+		UNIQUE KEY `event_per_user` (`event_id`,`user_id`),
+		INDEX `user` (`user_id`)
+		) COMMENT='Attendees of events'";
+		dbDelta( $create_attendees_table );
+
+		// TODO: Remove this once it has been run in production.
+		try {
+			// Don't run it during tests.
+			if ( 'true' === getenv( 'TRANSLATION_EVENTS_NO_IMPORT_LEGACY_ATTENDEES' ) ) {
```
amieiro: If you want to run this once in production and to avoid running on the tests, I think you need to invert the check:

```suggestion
			if ( 'true' !== getenv( 'TRANSLATION_EVENTS_NO_IMPORT_LEGACY_ATTENDEES' ) ) {
```

wporg-gp-translation-events.php
```diff
@@ -96,8 +99,59 @@ public function activate() {
...
+			array(
+				'post_type' => self::CPT,
+			)
+		);
+
+		$events              = $query->get_posts();
+		$stats_calculator    = new Stats_Calculator();
+		$attendee_repository = new Attendee_Repository();
+		foreach ( $events as $event ) {
+			foreach ( $stats_calculator->get_contributors( $event ) as $user ) {
```
amieiro: The `get_contributors` method receive and `event id`, so you need to update the parameter passed.

```suggestion
			foreach ( $stats_calculator->get_contributors( $event->ID ) as $users ) {
```
This method returns a list of users, so I think you need to loop though the users and add all to the new table.

amieiro (changes requested):

wporg-gp-translation-events.php
```diff
@@ -96,8 +99,59 @@ public function activate() {
...
+
+		// Create attendees table.
+		$create_attendees_table = "
+		CREATE TABLE `{$gp_table_prefix}event_attendees` (
+			`translate_event_attendees_id` int(11) unsigned NOT NULL AUTO_INCREMENT,
+			`event_id` int(10) NOT NULL COMMENT 'Post_ID of the translation_event post in the wp_posts table',
+			`user_id` int(10) NOT NULL COMMENT 'ID of the user who is attending the event',
+		PRIMARY KEY (`translate_event_attendees_id`),
+		UNIQUE KEY `event_per_user` (`event_id`,`user_id`),
+		INDEX `user` (`user_id`)
```
amieiro: Are you going to add here the new field for the user status in this event: attendee (default value), co-host,...?

psrpinto:
Thanks for reviewing @amieiro, I'm picking up this PR again and will address your remarks.

psrpinto:
I've rebased this with #191 so this PR now depends on that one. No other changes have been done, only rebased. New commits will be added after this comment, prior commits can be considered reviewed.

wporg-gp-translation-events.php
psrpinto: > The `get_contributors` method receive and event id, so you need to update the parameter passed.

Great catch, thank you! Fixed.

psrpinto: > This method returns a list of users, so I think you need to loop though the users and add all to the new table.

The `foreach` iterates over the result of the call to `get_contributors()`, and assigns the value at that index to the `$user` variable. I'm guessing you missed the `foreach`? Or I might be misunderstanding something.

wporg-gp-translation-events.php
psrpinto: Good point, I've added this to the TODO in the PR description.

wporg-gp-translation-events.php
psrpinto: Nice catch, fixed

includes/attendee-repository.php
psrpinto: I tried it and indeed it works, though warnings are written to the log, which also happens during tests, so it pollutes the test output. An example of these warnings:

> WordPress database error Duplicate entry '35-17' for key 'event_per_user'

Would it be ok to leave it as is?

psrpinto:
This is now ready for re-review. I've updated the PR description with a summary of all changes introduced here.

Concerning tracking who is host/contributor, finally I went with two booleans instead of an enum, I believe that is an appropriate solution here. Let me know if you have some remarks about that, or about anything else.

I also opened #195 and #196 which I would tackle as follow-ups to this PR.

includes/database.php
```diff
@@ -0,0 +1,82 @@
...
+		) COMMENT='Tracks translation actions that happened during a translation event'";
+		dbDelta( $create_actions_table );
+
+		// Create attendees table.
+		$create_attendees_table = "
+		CREATE TABLE `{$gp_table_prefix}event_attendees` (
+			`translate_event_attendees_id` int(11) unsigned NOT NULL AUTO_INCREMENT,
+			`event_id` int(10) NOT NULL COMMENT 'Post_ID of the translation_event post in the wp_posts table',
+			`user_id` int(10) NOT NULL COMMENT 'ID of the user who is attending the event',
+			`is_host` tinyint(1) default 0 not null comment 'Whether the user is a host of the event',
```
amieiro, line 35: The main host is stored in the `wp_posts` table as `post_author`, so maybe it is better to call the other host `cohosts`, so we can differentiate the main creator and the other hosts.

```suggestion
			`is_cohost` tinyint(1) default 0 not null comment 'Whether the user is a host of the event',
```

includes/database.php
```diff
@@ -0,0 +1,82 @@
...
+			`is_contributor` tinyint(1) default 0 not null comment 'Whether the user created or reviewed translations during the event',
```
amieiro: Why we need this field, as we store all the translations or suggestions in the `wp_gp_event_actions` table, so we don't need this field? In fact, we can have conflicting information if we store it in both places.

includes/database.php
```diff
@@ -0,0 +1,82 @@
...
+	/**
+	 * Previously, event attendance was tracked through user_meta.
+	 * This function imports this legacy attendance information into the attendees table.
+	 *
+	 * Instead of looping through all users, we consider only users who have contributed to an event.
+	 *
+	 * @deprecated TODO: Remove this function once this has been run in production.
+	 * @throws Exception
+	 */
+	private static function import_legacy_attendees(): void {
```
amieiro, line 64: This method is not importing the data for the `is_host` and `is_contributor` fields.

includes/database.php
```diff
@@ -0,0 +1,82 @@
...
+			`happened_at` datetime NOT NULL COMMENT 'When the action happened, in UTC',
+		PRIMARY KEY (`translate_event_actions_id`),
+		UNIQUE KEY `event_per_translated_original_per_user` (`event_id`,`locale`,`original_id`,`user_id`)
+		) COMMENT='Tracks translation actions that happened during a translation event'";
+		dbDelta( $create_actions_table );
+
+		// Create attendees table.
+		$create_attendees_table = "
+		CREATE TABLE `{$gp_table_prefix}event_attendees` (
+			`translate_event_attendees_id` int(11) unsigned NOT NULL AUTO_INCREMENT,
```
amieiro, line 32: Why we are using the table name in the field name and not only the `id` word? If we change it, we should change the `PRIMARY KEY` field.

```suggestion
			`id` int(11) unsigned NOT NULL AUTO_INCREMENT,
```

includes/database.php
psrpinto: I guess we could indeed calculate it from the `actions` table, though I don't know if that could be an expensive query, given the `actions` table will likely become a very large table with time.

For example, when displaying the event details page, we would need to calculate it for each attendee of the event.

Concerning conflicting information, I think that would not happen, because you can't go back to not being a contributor. I mean, if you contributed to an event, it's not possible to "undo" that contribution, right?

So once `is_contributor` is set to `true`, it will always be `true`, I think there is no scenario where it would become `false` again.

Personally I like that the data of attendees would be decoupled from the `actions`. If you think we should change it, I can do it.

includes/database.php
psrpinto: ~I pushed a commit and the method is now setting `is_host`. Concerning, `is_contributor`, I would take care of setting it when I tackle #196. Would that be ok?~

includes/database.php
psrpinto: This was done for consistency with the `event_actions` table, where the field is named `translate_event_actions_id`. I believe we named the primary key like that after feedback we received from wordpress.org maintainers, so I figured it would make sense to follow the same naming pattern here.

In fact, the primary key is technically not needed, but since the `event_actions` table has it (and it's also not technically needed), I guessed I should probably add it here too.

includes/database.php
psrpinto: The idea would be that the post author is no longer used to determine whether a user is a host. Instead, the first `host` is set to be the post author when the event is published.

I think the advantage of doing it like this is that there's a single way to determine whether a user is a host. An added advantage, is that author and host become decoupled, so it's possible for an event to have a single host that is not the post author. This makes it possible to "transfer" host if, for example, the user who created the event can no longer attend.

I realise this isn't a feature we have now, just using it as an example of an advantage of decoupling author from host.

includes/database.php
amieiro: My main concerns here are:
- If we have a bug in the code, we could have inconsistent information in the tables. E.g., translations for a user in the `event_actions` table and this user with `is_contributor=0` in the `event_ateendees` table. I could assume that this is unlikely.
- If we store this information in the `event_ateendees` table, each time a user makes a translation, we are going to need to query this table to know if this user is contributing to this event, so we need to do a lot of extra queries when the users translate strings. If we have this information in the `event_actions` table, we will only query it when we need to show info for an event, and the query is not expensive, because it only uses one table (will maybe a lot of registers). We could cache this query in the future. And as Alex suggested [here](https://github.com/WordPress/wporg-gp-translation-events/issues/196#issuecomment-2043171940), we could compute it once a day, with a cron, for the events that have finished the last day.

What do you think about?

psrpinto: These are fair points 👍

I also thought further about this and another problem with having `is_contributor` in the attendees table is that if the user would un-attend the event, the `is_contributor` would be lost. So we would need to prevent attendees who contributed from un-attending, which is not something I'm certain we want to do.

I will change it to remove `is_contributor` from the attendees table.

akirk: What about we call it `was_contributor` so that it's clear that this is supposed to be calculated after the event is over?

akirk: The idea is that we can use the `event_attendees` table to generate stats without having to query the `event_actions` table.

psrpinto: Given that these stats are not a feature we have right now, could we delay this decision until we implement those stats? At that point we can decide to introduce the `was_contributor` field, or to calculate those stats from both the `attendees` and `actions` tables.

akirk: Ok! Better than delaying this more.

amieiro: Ok, we can remove the `is_contributor` and move on with this PR to approve it.

psrpinto: I'm on this.

psrpinto:
I believe this is now ready. #195 and #199 will be follow-ups to this PR.

psrpinto:
~Don't merge yet, I accidentally pushed a commit to another branch~

psrpinto:
Now it's ready :)

akirk:
Sorry, I am not sure I understand 7e5dd652909e216cf92615cd6aaa5cf611d9fe9a. Why do contributors of previous projects get added as host?

psrpinto:
> Sorry, I am not sure I understand https://github.com/WordPress/wporg-gp-translation-events/commit/7e5dd652909e216cf92615cd6aaa5cf611d9fe9a. Why do contributors of previous projects get added as host?

Thanks for catching that, I had a brain freeze there, it was not what I intended to do, I got confused between contributor and host. I will revert that commit. 

The intention was to add the post_author of the event as an attendee with `host` set to `true`. I will tackle that in #195 (which I'm working on now), if that's ok with you.

includes/database.php
psrpinto: I will tackle setting `is_host` for legacy attendees in #195 .

amieiro (approved):

wporg-gp-translation-events.php
```diff
@@ -240,10 +211,15 @@ public function event_status_transition( string $new_status, string $old_status,
 			return;
 		}
 		if ( 'publish' === $new_status && ( 'new' === $old_status || 'draft' === $old_status ) ) {
-			$attendee_repository = new Attendee_Repository();
-			$current_user_id     = get_current_user_id();
-			if ( ! $attendee_repository->is_attending( $post->ID, $current_user_id ) ) {
-				$attendee_repository->add_attendee( $post->ID, $current_user_id );
+			$event_id            = $post->ID;
+			$user_id             = get_current_user_id();
```
akirk, line 215: @psrpinto and I talked about this and that it should probably rather be `$user_id = $post->post_author`

amieiro: So, if:
- User1 creates a draft event.
- User2 publishes this event.
- The user1 will be set as attendee and as host.

It makes sense to me. I will update it in https://github.com/WordPress/wporg-gp-translation-events/pull/200.

akirk: @amieiro actually made exactly that change in https://github.com/WordPress/wporg-gp-translation-events/pull/200/files/ca6c64390702d6e46f4302aa5cd835e202b6b010..5bd799f4757f54dfd871ad80aa98b1a8ae412512

akirk: Thanks, @amieiro!

PR #179, opened 2024-03-19
Small cleanup of handling ajax submit

psrpinto:
This PR cleans up the `submit_event_ajax()` function:

1. Move validation to the top of the function
2. Remove duplicated validation checks
3. Only parse the event id from the `$_POST` once
4. Only call `get_post()` once

This is in preparation for using the event repository to insert, update and delete an event.


psrpinto:
Closing since I'm implementing this in a different way in a new PR.

PR #180, opened 2024-03-19
Fix permission checks

psrpinto:
**#179 must be merged first**

Previously, we were doing:

```php
if ( ! ( $can_crud_event || current_user_can( 'edit_post', $event_id ) ) ) {
```

If `$can_crud_event` is true, the conditions after are not evaluated.

This PR fixes this issue.

### DIFF
--- a/wporg-gp-translation-events.php
+++ b/wporg-gp-translation-events.php
@@ -250,7 +250,8 @@ class Translation_Events {
 			if ( ! $event || self::CPT !== $event->post_type ) {
 				wp_send_json_error( esc_html__( 'Event does not exist.', 'gp-translation-events' ), 404 );
 			}
-			if ( ! ( $can_crud_event || current_user_can( 'edit_post', $event_id ) || intval( $event->post_author ) === get_current_user_id() ) ) {
+			$can_edit = $can_crud_event && current_user_can( 'edit_post', $event_id ) && intval( $event->post_author ) === get_current_user_id();
+			if ( ! $can_edit ) {
 				wp_send_json_error( esc_html__( 'The user does not have permission to edit or delete the event.', 'gp-translation-events' ), 403 );
 			}
 		}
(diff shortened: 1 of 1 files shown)


wporg-gp-translation-events.php
```diff
@@ -250,7 +250,8 @@ public function submit_event_ajax() {
 			if ( ! $event || self::CPT !== $event->post_type ) {
 				wp_send_json_error( esc_html__( 'Event does not exist.', 'gp-translation-events' ), 404 );
 			}
-			if ( ! ( $can_crud_event || current_user_can( 'edit_post', $event_id ) || intval( $event->post_author ) === get_current_user_id() ) ) {
+			$can_edit = $can_crud_event && current_user_can( 'edit_post', $event_id ) && intval( $event->post_author ) === get_current_user_id();
```
akirk, line 253: I think with this suggestion the post author can no longer edit the post when they are the post author but don't have elevated permissions otherwise.

wporg-gp-translation-events.php
```diff
@@ -259,7 +260,8 @@ public function submit_event_ajax() {
 			if ( ! $event || self::CPT !== $event->post_type ) {
 				wp_send_json_error( esc_html__( 'Event does not exist.', 'gp-translation-events' ), 404 );
 			}
-			if ( ! ( $can_crud_event || current_user_can( 'delete_post', $event->ID ) || get_current_user_id() === $event->post_author ) ) {
+			$can_delete = $can_crud_event && current_user_can( 'delete_post', $event->ID ) && get_current_user_id() === $event->post_author;
```
akirk, line 263: Same as above,

wporg-gp-translation-events.php
psrpinto: Great catch, ~will do the author check first.~

psrpinto:
Closing this since I think I misinterpreted what should happen, I guess we do want only one of the conditions to be true.

That said, currently any user who can edit posts can edit any event, not sure that's what we want, but it was not the scope of this PR to fix that.

PR #181, opened 2024-03-19
Block the creation, editing or deletion of an event with a past end date

### LINKED ISSUE #72: Prevent editing past events
I think we should remove the "Update Event" button for past events. Otherwise, users can easily change the dates to capture more contributions in the stats.

amieiro:
This PR blocks the creation, editing or deletion of an event with a past end date.

Only shows the "Edit" button in the current of future events.

image

The user can't create a past event.

image

If the user tries to edit a finished event, using the id, she will get an error:

image

Fixes https://github.com/WordPress/wporg-gp-translation-events/issues/72

### DIFF
--- a/includes/routes/event/edit.php
+++ b/includes/routes/event/edit.php
@@ -47,21 +47,21 @@ class Edit_Route extends Route {
 			$event_end     = self::convertToTimezone( get_post_meta( $event_id, '_event_end', true ), $event_timezone );
 			$now           = self::convertToTimezone( 'now', $event_timezone );
 			$event_end_utc = new DateTime( get_post_meta( $event_id, '_event_end', true ), new DateTimeZone( 'UTC' ) );
-			$now           = new DateTime( 'now', new DateTimeZone( 'UTC' ) );
+			$now_utc       = new DateTime( 'now', new DateTimeZone( 'UTC' ) );
 		} catch ( Exception $e ) {
 			// phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
 			error_log( $e );
 			$this->die_with_error( esc_html__( 'Something is wrong.', 'gp-translation-events' ) );
 		}
 
-		if ( $now > $event_end_utc ) {
+		if ( $now_utc > $event_end_utc ) {
 			$this->die_with_error( esc_html__( 'You cannot edit a finished event.', 'gp-translation-events' ), 403 );
 		}
 
 		$stats_calculator = new Stats_Calculator();
 		if ( ! $stats_calculator->event_has_stats( $event ) ) {
 			$current_user = wp_get_current_user();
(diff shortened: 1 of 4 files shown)


templates/events-my-events.php
```diff
@@ -34,10 +35,16 @@
 			$event_status                  = get_post_status( $event_id );
 			$event_start                   = ( new DateTime( get_post_meta( get_the_ID(), '_event_start', true ) ) )->format( 'M j, Y' );
 			$event_end                     = ( new DateTime( get_post_meta( get_the_ID(), '_event_end', true ) ) )->format( 'M j, Y' );
+			$envent_end_utc                = new DateTime( get_post_meta( get_the_ID(), '_event_end', true ), new DateTimeZone( 'UTC' ) );
```
psrpinto: There's a typo in the variable name, should be `$event_end_utc` :)

templates/events-my-events.php
```diff
@@ -34,10 +35,16 @@
...
+			$now                           = new DateTime( 'now', new DateTimeZone( 'UTC' ) );
+
+			$now > $envent_end_utc ? $past_event = true : $past_event = false;
```
psrpinto: I was confused by this initially, I thought it had no purpose as the result of the comparison wasn't being used, but then I noticed variables are being assigned in the condition. Personally, I think it would be more readable to do something like:

```php
$past_event = false;
if ( $now > $event_end_utc ) {
    $past_event = true;
}
```

psrpinto (commented):
Tested and it works 👍 

Will approve once typo in the variable name is fixed :)

includes/routes/event/edit.php
```diff
@@ -41,23 +42,30 @@ public function handle( int $event_id ): void {
...
+			$event_end_utc = new DateTime( get_post_meta( $event_id, '_event_end', true ), new DateTimeZone( 'UTC' ) );
+			$now_utc       = new DateTime( 'now', new DateTimeZone( 'UTC' ) );
 		} catch ( Exception $e ) {
 			// phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
 			error_log( $e );
 			$this->die_with_error( esc_html__( 'Something is wrong.', 'gp-translation-events' ) );
 		}
 
+		if ( $now_utc > $event_end_utc ) {
+			$this->die_with_error( esc_html__( 'You cannot edit a finished event.', 'gp-translation-events' ), 403 );
```
psrpinto: I think "past" would be more idiomatic than "finished".

```suggestion
			$this->die_with_error( esc_html__( 'You cannot edit a past event.', 'gp-translation-events' ), 403 );
```

templates/events-my-events.php
amieiro: As the `$past_event` is only used in one check, I removed it and put the comparison directly in the check.

psrpinto (approved):

akirk:
@amieiro I changed this a little in #157 where I introduced a `$event_end->is_in_the_past()`, could you take a look if I got it right there?

PR #182, opened 2024-03-20
Disable edit event with stats

### LINKED ISSUE #115: Disable edit event if event has stats
I think organizers should not be able to modify an event once there has been at least one contribution. We may decide on what can be modified, for example; event end date may be modified to extend the event but not to shorten it.

amieiro:
This PR depends on https://github.com/WordPress/wporg-gp-translation-events/pull/181, so we need to merge it before.

This PR disable the edit action in the `Edit_Route` if the event has translations and disable the `Edit` buttons in the:

**Show view for an event**

image

**"My events" template**

image

Fixes https://github.com/WordPress/wporg-gp-translation-events/issues/115

### DIFF
--- a/templates/events-header.php
+++ b/templates/events-header.php
@@ -33,7 +33,7 @@ use GP;
 	<?php if ( isset( $event_id ) && ! isset( $event_form_name ) ) : ?>
 	<p class="event-sub-head">
 			<span class="event-host">Host: <a href="<?php echo esc_attr( get_author_posts_url( $event->post_author ) ); ?>"><?php echo esc_html( get_the_author_meta( 'display_name', $event->post_author ) ); ?></a></span>
-			<?php if ( current_user_can( 'edit_post', $event_id ) && ( true === $is_editable_event ) ) : ?>
+			<?php if ( current_user_can( 'edit_post', $event_id ) && ( $is_editable_event ) ) : ?>
 				<a class="event-page-edit-link" href="<?php echo esc_url( gp_url( 'events/edit/' . $event_id ) ); ?>"><span class="dashicons dashicons-edit"></span>Edit event</a>
 			<?php endif ?>
 		</p>
--- a/templates/events-header.php
+++ b/templates/events-header.php
@@ -1,6 +1,9 @@
 <?php
 namespace Wporg\TranslationEvents;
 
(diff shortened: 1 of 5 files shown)


templates/events-header.php
```diff
@@ -30,7 +33,7 @@
 	<?php if ( isset( $event_id ) && ! isset( $event_form_name ) ) : ?>
 	<p class="event-sub-head">
 			<span class="event-host">Host: <a href="<?php echo esc_attr( get_author_posts_url( $event->post_author ) ); ?>"><?php echo esc_html( get_the_author_meta( 'display_name', $event->post_author ) ); ?></a></span>
-			<?php if ( current_user_can( 'edit_post', $event_id ) ) : ?>
+			<?php if ( current_user_can( 'edit_post', $event_id ) && ( true === $is_editable_event ) ) : ?>
```
psrpinto:
```suggestion
			<?php if ( current_user_can( 'edit_post', $event_id ) && $is_editable_event ) : ?>
```

templates/events-my-events.php
```diff
@@ -34,10 +35,18 @@
 			$event_status                  = get_post_status( $event_id );
 			$event_start                   = ( new DateTime( get_post_meta( get_the_ID(), '_event_start', true ) ) )->format( 'M j, Y' );
 			$event_end                     = ( new DateTime( get_post_meta( get_the_ID(), '_event_end', true ) ) )->format( 'M j, Y' );
+			$envent_end_utc                = new DateTime( get_post_meta( get_the_ID(), '_event_end', true ), new DateTimeZone( 'UTC' ) );
+			$now                           = new DateTime( 'now', new DateTimeZone( 'UTC' ) );
+			$stats_calculator              = new Stats_Calculator();
+			$has_translations              = $stats_calculator->event_has_stats( get_post() );
```
psrpinto: I think it would be clearer to name this `$has_stats` since we're using the term "stats" elsewhere in the code, and also because has translations makes me think translations were created during the event, but stats can also be about reviewed translations.

psrpinto (approved):
Looks good!

When a PR depends on another one, I think it's best to make its base branch the branch of the other PR instead of trunk, otherwise it's harder to review because it shows the diff of the commits in the other branch too.

PR #185, opened 2024-03-20
Add form handler

psrpinto:
> Should be easy to review commit by commit

This PR extracts the form handling logic out of the main plugin file into a separate class.

The only changes in behaviour concern validation, we're now checking that the title is set, and that the time zone is valid. No other changes were made.

## Summary of changes

- Extract form handling code out of the main plugin and into an `Event_Form_Handler` class.
- Extract form parsing/validation out of the `handle()` function into a dedicated private function that takes the form data as argument and returns an instance of `Event`. The goal is to have a valid instance of `Event` as early as possible, so that validation happens early, and so that the `Event` instance can be used in the rest of the code. (If you're interested in some theory about this see [Parse, don't validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/)).
- Use the `Event` instance instead of declaring a variable for each event property.

## Screen captures

Screenshot 2024-03-21 at 16 33 40

## Test instructions
Test creating, editing and deleting events. The behaviour should be the same as before, except that you now see an error if the title is empty.

### DIFF
--- a/assets/js/translation-events.js
+++ b/assets/js/translation-events.js
@@ -38,6 +38,10 @@
 		 * @param isDraft	  Whether the current event status is a draft or not
 		 */
 		function handleSubmit( eventStatus, isDraft ) {
+			if ( '' === $( '#event-title' ).val() ) {
+				$gp.notices.error( 'Event title must be set.' );
+				return;
+			}
 			if ( '' === $( '#event-start' ).val() ) {
 				$gp.notices.error( 'Event start date and time must be set.' );
 				return;
--- a/assets/js/translation-events.js
+++ b/assets/js/translation-events.js
@@ -38,6 +38,11 @@
 		 * @param isDraft	  Whether the current event status is a draft or not
 		 */
 		function handleSubmit( eventStatus, isDraft ) {
+			const $form = $( '.translation-event-form' );
+			if ( ! $form[0].reportValidity() ) {
+				return;
+			}
+
 			if ( '' === $( '#event-title' ).val() ) {
 				$gp.notices.error( 'Event title must be set.' );
 				return;
@@ -61,7 +66,6 @@
 				}
 			}
 			$( '#event-form-action' ).val( eventStatus );
-			const $form        = $( '.translation-event-form' );
 			const $is_creation = $( '#form-name' ).val() === 'create_event';
 
 			$.ajax(
--- a/includes/event/event-form-handler.php
+++ b/includes/event/event-form-handler.php
@@ -15,7 +15,7 @@ class Event_Form_Handler {
 		if ( ! is_user_logged_in() ) {
 			wp_send_json_error( esc_html__( 'The user must be logged in.', 'gp-translation-events' ), 403 );
 		}
-		$action           = isset( $_POST['form_name'] ) ? sanitize_text_field( wp_unslash( $_POST['form_name'] ) ) : '';
+		$action           = isset( $form_data['form_name'] ) ? sanitize_text_field( wp_unslash( $form_data['form_name'] ) ) : '';
 		$event_id         = null;
 		$event            = null;
 		$response_message = '';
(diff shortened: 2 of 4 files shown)


akirk:
> This PR extracts the form handling logic out of the main plugin file into a separate class.

Just to be sure, this is a janitorial PR and doesn't change functionality? How can it be tested? I suppose submitting the form with wrong parameters?

amieiro (dismissed):
I am getting an error trying to create a new draft event with these parameters:

image

This is the debug.log file output:

```
[21-Mar-2024 09:23:58 UTC] PHP Fatal error:  Uncaught Error: Call to a member function setTimezone() on bool in wp-content/plugins/wporg-gp-translation-events/includes/event/event-form-handler.php:227
Stack trace:
#0 wp-content/plugins/wporg-gp-translation-events/includes/event/event-form-handler.php(61): Wporg\TranslationEvents\Event\Event_Form_Handler->parse_form_data(Array)
#1 wp-content/plugins/wporg-gp-translation-events/wporg-gp-translation-events.php(189): Wporg\TranslationEvents\Event\Event_Form_Handler->handle(Array)
#2 wp-includes/class-wp-hook.php(324): Wporg\TranslationEvents\Translation_Events->submit_event_ajax('')
#3 wp-includes/class-wp-hook.php(348): WP_Hook->apply_filters('', Array)
#4 wp-includes/plugin.php(517): WP_Hook->do_action(Array)
#5 wp-admin/admin-ajax.php(192): do_action('wp_ajax_submit_...')
#6 /Users/xxxxx/.composer/vendor/laravel/valet/server.php(250): require('/Users/xxxxx/...')
#7 {main}
  thrown in wp-content/plugins/wporg-gp-translation-events/includes/event/event-form-handler.php on line 227

```
If I dump the parameters used just before calling the `return new Event`, these are the problematic values:


```
$event_id: ""
$start: false
$end: false
$timezone: DateTimeZone {
                        "timezone_type": 3
                        "timezone": "Europe/Madrid"
                      }
$title: "Future event"
$event_status: ""
$title: "Future event"
$description: "The description"
```

akirk:
@amieiro that's probably because this needs to be rebased after #157

psrpinto:
Yes indeed, this is a janitorial PR, doesn't modify behaviour. Sorry, I should've made that clear in the description.

I'm taking care of resolving conflicts.

psrpinto:
I've now resolved conflicts and fixed some issues. I tested locally and it worked correctly.

I updated the PR description since there's some changes since the PR was first created.

akirk:
Creating an event works but when I try to delete it I get a 422 and no error message, it's as if nothing happens. There are JS errors:

Screenshot 2024-03-21 at 18 06 34

akirk:
Could we add a `required` attribute to the title input so that we also get some client validation?

psrpinto:
Good catch, it didn't occur to me to test the delete.

It's now fixed, by moving the delete code into an `if`, and create/update into the `else`. This is because for deleting there's no need to parse the event form into an event.

psrpinto:
> Could we add a `required` attribute to the title input so that we also get some client validation?

It actually [already has the `required`](https://github.com/WordPress/wporg-gp-translation-events/blob/trunk/templates/events-form.php#L33) but it doesn't seem to result in client-side validation. Maybe because we're submitting through AJAX?

akirk:
Indeed. I think it can be triggered calling this on the form: `form.checkValidity()`

psrpinto:
I looked into `form.checkValidity()` but AFAICT it just returns false, it doesn't seem to trigger any error message on the form input itself. So for now I just added another check, like we're doing for start and end.

akirk:
Did you try[ `reportValidity()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/reportValidity)?

psrpinto:
That works, thanks!

Screenshot 2024-03-22 at 11 29 32

akirk:
The default time seems to be "now" in UTC. I am in UTC+1 so the time is already in the past. It's not too bad since you are supposed to change the dates anyway but could we also just leave the date fields empty?


Screenshot 2024-03-22 at 12 44 05

psrpinto:
@akirk This issue seems to already be on trunk, so I would rather fix it in another PR, is that ok? I opened #187 to track it.

I also think we should do a review of the whole flow (user time -> utc -> database -> user time) to make sure time zones are being handled correctly, I'm not confident there are no issues there. I mentioned this in #187 too.

akirk (approved):
Everything that I tested works, so let's move on with this!

amieiro:
We are having some problems with the timezone conversion. Currently, I am reviewing some problems I found today in `trunk`.

I have created an event for July:
- Starts: 2024-07-01 00:01
- Ends: 2024-08-31 23:59
- Timezone: Europe/Madrid (CEST in July, UTC+2)

screencapture-glotpress-test-glotpress-events-new-2024-03-22-12_46_48

The information is stored correctly in the database

image

But when I try to see the event, it is making a bad conversion:

image

psrpinto:
@amieiro I added your comment in #187. Ok with you if we merge this one?

amieiro:
> @amieiro I added your comment in #187. Ok with you if we merge this one?

Yes, of course, it is ok for me to merge this PR.

PR #186, opened 2024-03-22
Add  text snippets

### LINKED ISSUE #96: Add Text Snippets to the Edit Event page
You might want to reuse the same introductory snippets in every post, things like:

> If you have never translated before, check out the <a href="https://make.wordpress.org/polyglots/handbook/" rel="nofollow">https://make.wordpress.org/polyglots/handbook/</a>

They could be added through text links on the side (or blocks if we do #95 )

trymebytes:
Fixes #96 

**Screen recording**
eventsnippets

**Testing Instructions**

Add the code block below to this [line](https://github.com/WordPress/wporg-gp-translation-events/blob/6c5218259f8ae0ddd05c95ea216fd257da2976fd/wporg-gp-translation-events.php#L73). Or you could add it to another plugin installed in the same environment. 

```		
add_filter( 'wporg_translation_events_snippets', function () {
			return array(
				array(
					'title'   => '1st snippet',
					'snippet' => 'This is the 1st snippet',
				),
				array(
					'title'   => '2nd snippet',
					'snippet' => 'This is the 2nd snippet',
				),
				array(
					'title'   => '3rd snippet',
					'snippet' => 'This is the 3rd snippet, with a <a href="https://example.com">link</a> on it.',
				),
			);
		} );
```

### DIFF
--- a/assets/js/translation-events.js
+++ b/assets/js/translation-events.js
@@ -33,7 +33,9 @@
 					'click',
 					function ( e ) {
 						e.preventDefault();
-						$( this ).closest( 'div' ).find( 'textarea' ).append( $( this ).data( 'snippet' ) );
+						var textArea = $( this ).closest( 'div' ).find( 'textarea' );
+						var textAreaContent = textArea.val();
+						textArea.val( textAreaContent + $( this ).data( 'snippet' ) );
 					}
 				);
 			}
--- a/assets/js/translation-events.js
+++ b/assets/js/translation-events.js
@@ -33,7 +33,7 @@
 					'click',
 					function ( e ) {
 						e.preventDefault();
-						$( this ).closest( 'div').find( 'textarea' ).append( $( this ).data( 'snippet' ) );
+						$( this ).closest( 'div' ).find( 'textarea' ).append( $( this ).data( 'snippet' ) );
 					}
 				);
 			}
--- a/assets/js/translation-events.js
+++ b/assets/js/translation-events.js
@@ -33,7 +33,7 @@
 					'click',
 					function ( e ) {
 						e.preventDefault();
-						var textArea = $( this ).closest( 'div' ).find( 'textarea' );
+						var textArea        = $( this ).closest( 'div' ).find( 'textarea' );
 						var textAreaContent = textArea.val();
 						textArea.val( textAreaContent + $( this ).data( 'snippet' ) );
 					}
--- a/assets/js/translation-events.js
+++ b/assets/js/translation-events.js
@@ -28,6 +28,14 @@
 						handleDelete()
 					}
 				);
+
+				$( '.text-snippet' ).on(
+					'click',
+					function ( e ) {
+						e.preventDefault();
+						$( this ).closest( 'div').find( 'textarea' ).append( $( this ).data( 'snippet' ) );
+					}
+				);
 			}
 		);
 
--- a/includes/event-text-snippet.php
+++ b/includes/event-text-snippet.php
@@ -13,7 +13,7 @@ class Events_Text_Snippet {
 	public static function generate_snippet_links( $snippets ) : string {
 		$links = '<ul class="text-snippets">';
 		foreach ( $snippets as $snippet ) {
-			$links .= sprintf( '<li><a href="#" data-snippet="%s">%s</a></li>', $snippet['snippet'], $snippet['title'] );
+			$links .= sprintf( '<li><a href="#" class="text-snippet" data-snippet="%s">%s</a></li>', $snippet['snippet'], $snippet['title'] );
 		}
 		$links .= '</ul>';
 		return $links;
--- a/includes/event-text-snippet.php
+++ b/includes/event-text-snippet.php
@@ -7,33 +7,15 @@ class Events_Text_Snippet {
 	/**
 	 * Generate links for text snippets.
 	 *
-	 * @param array $snippets The array of snippets.
-	 * @return string The generated snippet links.
+	 * @return string The snippet links in a list.
 	 */
-	public static function generate_snippet_links( $snippets ): string {
-		$links = '<ul class="text-snippets">';
+	public static function get_snippet_links(): string {
+		$snippets           = apply_filters( 'wporg_translation_events_snippets', array() );
+		$snippets_link_list = '<ul class="text-snippets">';
 		foreach ( $snippets as $snippet ) {
-			$links .= sprintf( '<li><a href="#" class="text-snippet" data-snippet="%s">%s</a></li>', $snippet['snippet'], $snippet['title'] );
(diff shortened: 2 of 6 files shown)


includes/event-text-snippet.php
```diff
@@ -0,0 +1,39 @@
...
+	 */
+	public static function get_default_snippets(): array {
+		return array(
+			array(
+				'title'   => 'General Instructions',
+				'snippet' => 'These are general instructions for anyone intending to contribute',
+			),
+			array(
+				'title'   => 'New Contributor Guidance',
+				'snippet' => 'Here is the snippet for new contributor guidance',
```
psrpinto: Are these intended as placeholder snippets that we would replace later, or are they the actual snippets we want to have?

I'm asking because when creating the event, I found the snippets to be a bit confusing, as they feel more like instructions to the user creating/editing the event, but I was expecting them to be instructions to the viewer of the event.

includes/event-text-snippet.php
```diff
@@ -0,0 +1,39 @@
...
+	/**
+	 * Generate links for text snippets.
+	 *
+	 * @param array $snippets The array of snippets.
+	 * @return string The generated snippet links.
+	 */
+	public static function generate_snippet_links( $snippets ): string {
+		$links = '<ul class="text-snippets">';
+		foreach ( $snippets as $snippet ) {
+			$links .= sprintf( '<li><a href="#" class="text-snippet" data-snippet="%s">%s</a></li>', $snippet['snippet'], $snippet['title'] );
```
psrpinto: I'm not sure why but the `<li>`s are being rendered on the same line.

Screenshot 2024-04-02 at 14 15 14

assets/js/translation-events.js
```diff
@@ -28,6 +28,14 @@
 						handleDelete()
 					}
 				);
+
+				$( '.text-snippet' ).on(
+					'click',
+					function ( e ) {
+						e.preventDefault();
+						$( this ).closest( 'div' ).find( 'textarea' ).append( $( this ).data( 'snippet' ) );
```
psrpinto: I noticed that the snippet buttons don't work if the text area content is modified after the page renders. I'm not sure if this is intended, but to me it felt like something was wrong. I typed a description, then clicked a snippet, but nothing happened.

amieiro (changes requested):
I would suggest putting the different snippets in an unordered list, changing from:

image

to

image

You only need to remove 

```
ul.text-snippets li {
    display: inline;
}
```

and updating this CSS:
```
ul.text-snippets {
    /* list-style-type: none; */
    padding: 0;
    margin-left: 160px;
}
```

Related with the link content, I know we are going to overwrite them at translate.w.org, but I think they should have useful content without overwriting them.

The snippets are added to the textarea only if I have not changed the empty content from it. I have tested it with these tests:

1. Reload the page.
2. Add some text to the text area and then try to add one snippet.
3. Reload the page.
4. Add one snippet, delete it and then try to add the other snippet.
5. Reload the page.
6. Add one snippet, add some text to the textarea, and then try to add the other snippet.

includes/event-text-snippet.php
trymebytes: Yes, they are placeholders.

assets/js/translation-events.js
trymebytes: Thanks for catching this

includes/event-text-snippet.php
trymebytes: Thank you for catching this, this has been fixed.

includes/event-text-snippet.php
trymebytes: I think we could remove the placeholders in this PR until we define what should snippets should be.

templates/events-form.php
```diff
@@ -43,8 +43,21 @@
 	<div>
 		<label for="event-description">Event Description</label>
 		<textarea id="event-description" name="event_description" rows="4" required><?php echo esc_html( $event_description ); ?></textarea>
-	</div>
-	<div>
+		<?php
+		echo wp_kses(
+			apply_filters( 'wporg_translation_events_snippets', Events_Text_Snippet::get_default_snippets() ),
```
akirk: Could we add some more parameters to allow filters to determine their snippets more specifically? For example (don't use this):
```suggestion
			apply_filters( 'wporg_translation_events_snippets', Events_Text_Snippet::get_default_snippets(), $event_title ),
```

trymebytes: Yes, I agree as it would allow for more customization of the snippets for specific situations or based on certain conditions. I would suggest we do this in a follow-up PR

amieiro (changes requested):
To test this PR, I have added this code in the `__construct()` method of the `Translation_Events` class (wporg-gp-translation-events.php).

```php
		\add_filter( 'wporg_translation_events_snippets', function () {
			return array(
				array(
					'title'   => '1st snippet',
					'snippet' => 'This is the 1st snippet',
				),
				array(
					'title'   => '2nd snippet',
					'snippet' => 'This is the 2nd snippet',
				),
				array(
					'title'   => '3rd snippet',
					'snippet' => 'This is the 3rd snippet, with a <a href="https://example.com">link</a> on it.',
				),
			);
		} );
```

image

The third element is not working with links, and I think it is a needed element, because in a snippet you are going to need to link to an external resource.

I will recommend adding the bullets as an unordered list, because I think the UI is clearer with them. This is the CSS you need to update:

```css
ul.text-snippets {
    padding: 0;
    margin-left: 160px;
}
```

image

amieiro (changes requested):
I have tested it with a different array:

```php
		add_filter( 'wporg_translation_events_snippets', function () {
			return array(
				array(
					'title'   => '1st snippet',
					'snippet' => 'This is the 1st snippet',
				),
				array(
					'title'   => '2nd snippet',
					'snippet' => 'This is the 2nd snippet',
				),
				array(
					'title'   => '3rd snippet',
					'snippet' => '<ul><li>This is the 3rd snippet, <tag>abc</tag>with a <a href="https://example.com">link</a> on it.</li><li><strong>Another text</strong></li></ul>',
				),
			);
		} );
```

And the third element doesn't filter some tags like `<tag></tag>` or `<strong></strong>`. Are we going to filter these tags, allowing only `<a>`, `<ul>` and `<li>`?

image

trymebytes:
Using the same array and using the 3rd snippet, here is how it's published;
image
The HTML seems to work has expected or am I mistaken?

amieiro:
What I was thinking was to filter the `data-snippet` content, so we won't have any unwanted HTML string in the textarea. But, of course, we should filter the textarea content just before storing in the database, because a user can add it manually.

> Using the same array and using the 3rd snippet, here is how it's published; image The HTML seems to work has expected or am I mistaken?

amieiro (approved):

PR #188, opened 2024-03-22
Use event repository for creating/updating/deleting events

psrpinto:
> Please don't be scared by the number of commits, they're small and should be self-explanatory with the commit message.

In previous PRs, we added an `Event_Repository` that acts as a layer between the database and the "business logic", but it wasn't being called anywhere yet.

In this PR, we change the form handling code so that creating/updating/deleting events uses the `Event_Repository`. There are no functional changes here, the behaviour should be the same as before.

I will tackle the reading side of things (calling the `get_*` methods of the repository) in another PR.

## Testing instructions
Test creating, editing and deleting events. If should behave as it did before.

### DIFF
--- a/includes/active-events-cache.php
+++ b/includes/active-events-cache.php
@@ -40,15 +40,4 @@ class Active_Events_Cache {
 
 		return $result;
 	}
-
-	/**
-	 * Invalidates the active events cache.
-	 *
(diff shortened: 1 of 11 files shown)


akirk (approved):
I've functionally tested this successfully and I think we can move forward with this.

PR #189, opened 2024-03-28
Highlight first time contributors

### LINKED ISSUE #166: Highlight first time contributors in the stats summary
In the stats summary text, we could decide to highlight the number of first time contributors.

trymebytes:
Fixes #166 

**Screenshot**
Screenshot 2024-04-04 at 16 51 01

### DIFF
--- a/includes/event/event-date.php
+++ b/includes/event/event-date.php
@@ -58,6 +58,15 @@ abstract class Event_Date extends DateTimeImmutable {
 		return $this->utc() < $current_date_time;
 	}
 
+	/**
+	 * Get the date in seconds.
+	 *
+	 * @return int The date in seconds.
+	 */
+	public function __toSeconds(): int {
+		return $this->utc()->getTimestamp();
+	}
+
 	public function print_relative_time_html( $format = false ) {
 		echo wp_kses(
 			'<time
--- a/includes/event/event-date.php
+++ b/includes/event/event-date.php
@@ -58,15 +58,6 @@ abstract class Event_Date extends DateTimeImmutable {
 		return $this->utc() < $current_date_time;
 	}
 
-	/**
-	 * Get the date in seconds.
-	 *
-	 * @return int The date in seconds.
-	 */
-	public function __toSeconds(): int {
-		return $this->utc()->getTimestamp();
-	}
-
 	public function print_relative_time_html( $format = false ) {
 		echo wp_kses(
 			'<time
--- a/includes/stats-calculator.php
+++ b/includes/stats-calculator.php
@@ -269,7 +269,7 @@ class Stats_Calculator {
 		// phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery
 		// phpcs:disable WordPress.DB.DirectDatabaseQuery.NoCaching
 		// phpcs:disable WordPress.DB.DirectDatabaseQuery.SchemaChange
-		$users_first_translation_date  = $wpdb->get_var(
+		$users_first_translation_date = $wpdb->get_var(
 			$wpdb->prepare(
 				"
 			select min(date_added) from {$gp_table_prefix}translations where user_id = %d
--- a/includes/stats-calculator.php
+++ b/includes/stats-calculator.php
@@ -262,7 +262,7 @@ class Stats_Calculator {
 	 *
 	 * @return bool True if the user is a first time contributor, false otherwise.
 	 */
-	public function is_first_time_contributors( $event_start, $user_id ) {
+	public function is_first_time_contributor( $event_start, $user_id ) {
 		global $wpdb, $gp_table_prefix;
 
 		// phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
--- a/includes/stats-calculator.php
+++ b/includes/stats-calculator.php
@@ -8,6 +8,8 @@ use WP_User;
 use GP;
 use GP_Locale;
 use GP_Locales;
+use DateTimeImmutable;
+use DateTimeZone;
 
 class Stats_Row {
 	public int $created;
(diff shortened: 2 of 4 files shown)


trymebytes:
@psrpinto suggests that we let the attendees table handle only the attendees and not contributors and it makes sense to me. I wonder if we can first use `wp_cache` till we decide if we want to have another table for contributors or leave it this way. 
Cc @akirk

akirk:
My thought was to do some sort of wrap up after the event to mark contributors as such when they won't change any longer, so that we can easily do stats about the contributor / attendee ratio.

trymebytes:
This sounds good, so we only fetch the first-time contributors after event has ended.

amieiro (changes requested):
I have tested, and you can see the `user2` is marked as a new contributor in the **stats summary**, but not in the **contributors list**.

image

To resolve this problem, you need to update this code:

```php
				<ul>
					<?php foreach ( $contributors as $contributor ) : ?>
						<li class="event-contributor" title="<?php echo esc_html( implode( ', ', $contributor->locales ) ); ?>">
							<a href="<?php echo esc_url( get_author_posts_url( $contributor->ID ) ); ?>"><?php echo get_avatar( $contributor->ID, 48 ); ?></a>
						</li>
					<?php endforeach; ?>
				</ul>
```

to 

```php
				<ul>
					<?php foreach ( $contributors as $contributor ) : ?>
						<li class="event-contributor" title="<?php echo esc_html( implode( ', ', $contributor->locales ) ); ?>">
							<a href="<?php echo esc_url( get_author_posts_url( $contributor->ID ) ); ?>"><?php echo get_avatar( $contributor->ID, 48 ); ?></a>
							<?php if ( $stats_calculator->is_first_time_contributors( $event_start, $contributor->ID ) ) : ?>
								<span class="first-time-contributor-tada"></span>
							<?php endif; ?>
						</li>
					<?php endforeach; ?>
				</ul>
```

With this change, the new contributor is marked in the contributors list.

image

includes/stats-calculator.php
```diff
@@ -253,4 +253,39 @@ public function event_has_stats( int $event_id ): bool {
...
+
+	/**
+	 * Check if a user is a first time contributor.
+	 *
+	 * @param Event_Start_Date $event_start The event start date.
+	 * @param int              $user_id      The user ID.
+	 *
+	 * @return bool True if the user is a first time contributor, false otherwise.
+	 */
+	public function is_first_time_contributors( $event_start, $user_id ) {
```
psrpinto:
```suggestion
	public function is_first_time_contributor( $event_start, $user_id ) {
```

psrpinto:
```suggestion
	public function is_first_time_contributors( $event_start, $user_id ): bool {
```

includes/stats-calculator.php
```diff
@@ -253,4 +253,39 @@ public function event_has_stats( int $event_id ): bool {
...
+			)
+		);
+
+		if ( ! $users_first_translation_date ) {
+			return true;
+		}
+		$first_translation_date_in_sec = ( new \DateTime( $users_first_translation_date ) )->getTimestamp();
+		$event_start_date_in_sec       = $event_start->__toSeconds();
+		$days_in_sec                   = 60 * 60 * 24;
+		return ( $event_start_date_in_sec - $first_translation_date_in_sec ) < $days_in_sec;
```
psrpinto: We could simplify this logic with something like:

```php
$now                    = new DateTimeImmutable( 'now', new DateTimeZone( 'UTC' ) );
$first_translation_date = new DateTimeImmutable( $users_first_translation_date, new DateTimeZone( 'UTC' ) );

return $now->diff( first_translation_date )->days <= 1;
```

This would also remove the need for the `__toSeconds()` method.

includes/stats-calculator.php
```diff
@@ -253,4 +255,40 @@ public function event_has_stats( int $event_id ): bool {
...
+		);
+
+		if ( get_userdata( $user_id ) && ! $users_first_translation_date ) {
+			return true;
+		}
+		$event_start_date_time  = new DateTimeImmutable( $event_start->__toString(), new DateTimeZone( 'UTC' ) );
+		$first_translation_date = new DateTimeImmutable( $users_first_translation_date, new DateTimeZone( 'UTC' ) );
+		$multiplier             = $users_first_translation_date > $event_start ? 1 : -1;
+		$date_diff              = ( $event_start_date_time->diff( $first_translation_date )->days ) * $multiplier;
+		return $date_diff >= -1;
```
amieiro: You can compare `DateTimeImmutable` objects in PHP using standard comparison operators, so I think this code is simpler:

```suggestion
		if ( $event_start_date_time <= $first_translation_date ) {
			return true;
		}
		return false;
```

includes/event/event-date.php
```diff
@@ -58,6 +58,15 @@ public function is_in_the_past() {
...
 
+	/**
+	 * Get the date in seconds.
+	 *
+	 * @return int The date in seconds.
+	 */
+	public function __toSeconds(): int {
+		return $this->utc()->getTimestamp();
+	}
+
```
amieiro: I think we are not using this method. Can we remove it, or are you planning to use it in another place?

```suggestion
```

amieiro (changes requested):

includes/stats-calculator.php
trymebytes: Thank you for the comment, I am also trying to check if the user's first translation was made in at most1 day **before** the event start date. I realised that the `diff()` function returns absolutes regardless if it's before or after the specific date so I needed a way to check that. Therefore, if the date is before the start date, it should return a negative value and if after positive. Perhaps I could change the name of the variable `multiplier`.

includes/event/event-date.php
trymebytes: Thanks good catch!

includes/stats-calculator.php
akirk: That's a great explanation, can you add it as a code comment (maybe rewrite it a little to be less prose)?

amieiro: In this situation, I would recommend you to subtract 1 day using the `modify` method, explaining in a comment why we do this, because it is not trivial to understand the reason (I like it).

```suggestion
		// We subtract one day from the first translation date, so we set the translator as a "first time contributor" if she has started translating the previous day to the event starts.
		$first_translation_date = $first_translation_date->modify('-1 day');
		if ( $event_start_date_time <= $first_translation_date ) {
			return true;
		}
		return false;
```

akirk: @amieiro I like that solution better than a code comment. Nice!

amieiro: @akirk the code comment is only to explain the reason why we subtract one day, because if I see this code without any context, it will be hard to understand for me. That is the only reason to add it, nothing else.

akirk: That's what I was trying to say: the code change you suggested is the way to go and is better than the explanation that @trymebytes provided.

I think your code comment would be even better if it just explained what we consider as a first time contributor, not what the code does at it speaks for itself.

Something like:
```
// A first time contributor is someone who has never submitted a translation, or only within 1 day before the event.
// Someone who has submitted their first translation before that is not considered a first time contributor.
```

trymebytes: Thank you for the idea @amieiro. I just tested and I think it should be written this way;
```suggestion
		$event_start_date_time = $event_start_date_time->modify( '-1 day' );
		if ( $event_start_date_time <= $first_translation_date ) {
		 	return true;
		}
		return false;
```

amieiro (approved):
Works fine!

akirk:
@trymebytes Could you please add a unit test for this in a follow up PR?

PR #190, opened 2024-04-02
Use event repository for retrieving events

psrpinto:
> Apologies in advance for the high number of commits. If you review commit-by-commit, it should be relatively easy to review., a lot of the commits are very simple.

This PR is the conclusion of the event repository effort. In #188 we made it so that creating/updating/deleting events happens through the event repository. In this PR, we make it so that retrieving events also happens through the event repository.

With this PR, all access to events now happens through the event repository.

## Summary of changes

- Use `get_*()` methods from `Event_Repository` everywhere, instead of `get_posts()` or `WP_Query()`
- Add missing `get_*()` methods in `Event_Repository`
  - Not all needed methods were present yet
- Fix issues in some of the `get_*()` methods in `Event_Repository`
  - `Event_Repository` is now doing the same queries that were being done prior to this PR, so that this PR behaves the same as the previous code.
- Make it so that pagination handling is done only in the `execute_events_query()` method.
    - No need to call `assert_pagination_arguments()` in individual methods any more.

## Testing instructions

Test creating, editing, deleting and viewing events.

PR #191, opened 2024-04-03
Fix tests

psrpinto:
Requires #190 

This PR makes tests pass again (they're failing on trunk).

## Summary of changes

-  Make it so that `Event` uses `Event_Start_Date` and `Event_End_Date` internally, instead of using `DateTimeImmutable`
- Don't show seconds in event create/edit form
- Fix tests
- Move `event-date.php` to `Event` namespace

## Testing instructions

There should be no functional changes, everything should work the same as before, except that seconds aren't shown in the event edit form.

### DIFF
--- a/includes/event/event-date.php
+++ b/includes/event/event-date.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace Wporg\TranslationEvents;
+namespace Wporg\TranslationEvents\Event;
 
 use DateTime;
 use DateTimeImmutable;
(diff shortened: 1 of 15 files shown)


akirk (approved):
Thank you!

psrpinto:
This PR was based on #190, so the branch was merged into there. Is this what you intended to do @amieiro?

I think it might be best to not merge one PR into another, because that results in the diff of the first PR also containing changes of the second, which makes reviewing harder.

Typically when working with stacked PRs, I think it's more common to wait for the first PR to be merged, then merge the second. What do you think, should we do this going forward?

PR #197, opened 2024-04-08
Database upgrade

### LINKED ISSUE #195: Implement database upgrade system
[Like GlotPress](https://github.com/GlotPress/GlotPress/blob/develop/gp-settings.php#L137-L142) with a version number.

- Version 1 is the version prior to the introduction of the attendees table
- Version 2 is the version that introduces the attendees table

The following data migration steps are needed when migrating to version 2:

- [x] Create an attendee for the owner of each event, with `is_host` set to `true`
- [x] Import legacy attendees (from usermeta) to the attendees table

psrpinto:
Fixes #195

### DIFF
--- a/includes/upgrade.php
+++ b/includes/upgrade.php
@@ -7,21 +7,39 @@ use WP_Query;
 use Wporg\TranslationEvents\Attendee\Attendee;
 
 class Upgrade {
+	private const VERSION        = 2;
+	private const VERSION_OPTION = 'wporg_gp_translations_events_version';
+
 	public static function upgrade(): void {
+		$previous_version = get_option( self::VERSION_OPTION );
+
+		// If previous version is not set yet, set it to version 1.
+		if ( false === $previous_version ) {
+			$previous_version = 1;
+			update_option( self::VERSION_OPTION, $previous_version );
+		}
+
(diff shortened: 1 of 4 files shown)


includes/upgrade.php
```diff
@@ -0,0 +1,106 @@
...
+	private const VERSION        = 2;
+	private const VERSION_OPTION = 'wporg_gp_translations_events_version';
+
+	public static function upgrade(): void {
+		$previous_version = get_option( self::VERSION_OPTION );
+
+		// If previous version is not set yet, set it to version 1.
+		if ( false === $previous_version ) {
+			$previous_version = 1;
+			update_option( self::VERSION_OPTION, $previous_version );
```
akirk: We could just drop this line
```suggestion
```

akirk:
> Should we keep it as is, or should we do it on every request to wp-admin? Or another alternative? Please let me know.

I'd prefer it to go the GlotPress way and do it upon every `is_admin()` request.

I've tested the PR and it works well upon plugin activation.

psrpinto:
This should be good to go now

psrpinto:
This should be good to go now.

akirk (approved):
Great! Thank you!

PR #198, opened 2024-04-09
Show the strings of a translation event

### LINKED ISSUE #169: See the unreviewed strings of a translation event
It would be very rewarding for participants of a translation events to make sure that their translations can be approved, so we should make it easy to discover them.

amieiro:
Fixes https://github.com/WordPress/wporg-gp-translation-events/issues/169. Needs https://github.com/GlotPress/gp-translation-helpers/pull/222.

This adds links to the Translations table:
Screenshot 2024-04-17 at 16 43 05

For example for the WordCamp Asia contributor day, this now adds new URLs like:
- https://translate.wordpress.org/events/wordcamp-asia-2024-translations/ceb/
- https://translate.wordpress.org/events/wordcamp-asia-2024-translations/ceb/waiting/

and that page then lists all the strings that were modified in the project:
Screenshot 2024-04-17 at 17-37-29 Translation Events – WordCamp Asia 2024 Contributor Day WordPress org

Here you can see how this also takes into account individual project permissions. I am a PTE for German in the Friends plugin, but not a GTE for German, thus cannot approve in Core:

Screenshot 2024-04-17 at 17-36-34 akirk dev ord Translation Events – WordCamp Asia 2024 Contributor Day WordPress org


includes/routes/event/translations.php
```diff
@@ -0,0 +1,153 @@
...
+			$filters = array();
+			$sort = array();
+			$glossary = GP::$glossary->get( $project->id, $locale );
+			$page = 1;
+			$per_page = 10000;
+			$total_translations_count = 0;
+			$text_direction = 'ltr';
+			foreach ( $translations as $translation ) {
+//				$args['trasnslations'][] = GP::$translation->get( $translation->id );
+				$args['trasnslations'][] = new \Translation_Entry( $translation );
```
akirk:
```suggestion
				$args['translations'][] = new \Translation_Entry( $translation );
```

includes/stats-calculator.php
```diff
@@ -253,4 +253,41 @@ public function event_has_stats( int $event_id ): bool {
 
 		return ! empty( $stats->rows() );
 	}
+
+	public function unreview_translations( int $event_id ) {
```
psrpinto: I assume this is a typo

```suggestion
	public function unreviewed_translations( int $event_id ) {
```

templates/event-translations.php
```diff
@@ -0,0 +1,145 @@
+<?php
```
psrpinto: Would there be a way to to not duplicate the code for generating the translation table, and instead call the existing GlotPress code? This would mean that this plugin would automatically get any improvements or fixes that are made to the translations table in GlotPress.

I took a quick look at the GlotPress code and I'm guessing it is currently not possible to do such a thing (I could be mistaken), but we could consider making changes to GlotPress that make it possible.

akirk: We started using the `translations` template but it also includes header code. Also we need to do some tricks to get multiple of the tables working on the same page (e.g. the translations table has an `id="translations"` of which there cannot be multiple on a page).

It might be worth to propose to add the ability to have multiple translation tables on the same page in GlotPress but this is a) hard to argue, and b) not a quick PR. So I suggest to go this route for now.

includes/stats-calculator.php
akirk: I removed this while rebasing, its no longer needed.

wporg-gp-translation-events.php
```diff
@@ -84,6 +84,7 @@ public function gp_init() {
 		GP::$router->add( '/events/attend/(\d+)', array( 'Wporg\TranslationEvents\Routes\User\Attend_Event_Route', 'handle' ), 'post' );
 		GP::$router->add( '/events/host/(\d+)/(\d+)', array( 'Wporg\TranslationEvents\Routes\User\Host_Event_Route', 'handle' ), 'post' );
 		GP::$router->add( '/events/my-events', array( 'Wporg\TranslationEvents\Routes\User\My_Events_Route', 'handle' ) );
+		GP::$router->add( '/events/([a-z0-9_-]+)/([a-z0-9_-]+)(?:/([a-z0-9_-]+))?', array( 'Wporg\TranslationEvents\Routes\Event\Translations_Route', 'handle' ) );
```
psrpinto: Could we "namespace" this route? Something like `/events/$slug/translations/$locale`. Would be consistent with other routes, and would make explicit that this is the translations page.

psrpinto (changes requested):
I'm getting a 404 when trying to access the translations page, by clicking in an entry in the stats table. The URL is:

http://localhost:8888/events/test-hosts/pt/

Not sure why it's 404ing.

includes/routes/event/translations.php
```diff
@@ -0,0 +1,186 @@
...
+
+namespace Wporg\TranslationEvents\Routes\Event;
+
+use Exception;
+use GP;
+use GP_Locales;
+use GP_Original;
+use Translation_Entry;
+use Wporg\TranslationEvents\Attendee_Repository;
+use Wporg\TranslationEvents\Event\Event_Repository_Interface;
```
psrpinto: These are unused.

```suggestion
```

includes/routes/event/translations.php
```diff
@@ -0,0 +1,186 @@
...
+use Wporg\TranslationEvents\Routes\Route;
+use Wporg\TranslationEvents\Stats\Stats_Calculator;
+use Wporg\TranslationEvents\Translation_Events;
+
+/**
+ * Displays the event details page.
+ */
+class Translations_Route extends Route {
+	private $event_repository;
```
psrpinto:
```suggestion
	private Event_Repository_Interface $event_repository;
```

(Also requires a `use` statement)

includes/routes/event/translations.php
```diff
@@ -0,0 +1,186 @@
+<?php
+
+namespace Wporg\TranslationEvents\Routes\Event;
+
+use Exception;
```
psrpinto:
```suggestion
```

includes/routes/event/translations.php
```diff
@@ -0,0 +1,186 @@
...
+ */
+class Translations_Route extends Route {
+	private $event_repository;
+
+	public function __construct() {
+		parent::__construct();
+		$this->event_repository = Translation_Events::get_event_repository();
+	}
+
+	public function handle( string $event_slug, string $locale, string $status = '%' ): void {
```
psrpinto: It feels a bit weird to have the SQL wildcard here, initially it was confusing to me. Maybe this could be `all` instead?

akirk: Understandable, I changed it to `any`.

wporg-gp-translation-events.php
akirk: Agreed, thanks for the suggestion! I've added variables.

akirk:
@psrpinto could you try again with the latest regex updates?

wporg-gp-translation-events.php
psrpinto: What I meant in the suggestion above was to add the literal string `translations` (as an example, can be another string) because the URL is currently simply `/events/wordcamp-bilbao/pt/`. Instead the URL would become `/events/wordcamp-bilbao/translations/pt/`.

I think this would make it easier to add routes in the future, without them clashing with the translations route. For example, if we decide to add a separate route for the list of attendees of an event, its URL would then be `/events/wordcamp-bilbao/attendees/`.

psrpinto:
The 404 is still there but I now realised why it's happening: it lacks the `/glotpress` prefix, it's `/events/$event-slug/translations/pt/` but it should be `/glotpress/events/$event-slug/translations/pt/`.

templates/event.php
```diff
@@ -134,14 +135,16 @@
 					<?php foreach ( $event_stats->rows() as $_locale => $row ) : ?>
 					<tr>
 						<td title="<?php echo esc_html( $_locale ); ?> "><a href="<?php echo esc_url( gp_url_join( gp_url( '/languages' ), $row->language->slug ) ); ?>"><?php echo esc_html( $row->language->english_name ); ?></a></td>
-						<td><?php echo esc_html( $row->created ); ?></td>
+						<td><a href="<?php echo esc_url( wp_make_link_relative( gp_url_join( get_the_permalink( $event_id ), 'translations', $row->language->slug ) ) ); ?>"><?php echo esc_html( $row->created ); ?></a></td>
+						<td><a href="<?php echo esc_url( wp_make_link_relative( gp_url_join( get_the_permalink( $event_id ), 'translations', $row->language->slug, 'waiting' ) ) ); ?>"><?php echo esc_html( $row->waiting ); ?></a></td>
```
amieiro: Currently, in these URLs we don't have the `glotpress` text just after the domain. `https://glotpress.test/events/{event-slug}/translations/{language-slug}/`. This update adds it `https://glotpress.test/glotpress/events/{event-slug}/translations/{language-slug}/`

```suggestion
						<td><a href="<?php echo esc_url( gp_url_join( gp_url( wp_make_link_relative( get_the_permalink( $event_id ) ) ), 'translations', $row->language->slug ) ); ?>"><?php echo esc_html( $row->created ); ?></a></td>
						<td><a href="<?php echo esc_url( gp_url_join( gp_url( wp_make_link_relative( get_the_permalink( $event_id ) ) ), 'translations', $row->language->slug, 'waiting' ) ); ?>"><?php echo esc_html( $row->waiting ); ?></a></td>
```

templates/event-translations-header.php
```diff
@@ -0,0 +1,63 @@
...
+namespace Wporg\TranslationEvents;
+
+use GP;
+use Wporg\TranslationEvents\Event\Event;
+
+/** @var Event  $event */
+
+/* translators: %s: Event title. */
+gp_title( sprintf( __( 'Translation Events - %s' ), esc_html( $event->title() ) ) );
+gp_breadcrumb_translation_events( array( '<a href="' . esc_attr( gp_url( wp_make_link_relative( get_the_permalink( $event->id() ) ) ) ) . '">' . esc_html( $event->title() ) . '</a>', __( 'Translations', 'glotpress' ), $locale ) );
```
amieiro: Currently, we add the `english name` and the `locale name`.

image

This update uses only the `english name`.

image

```suggestion
gp_breadcrumb_translation_events( array( '<a href="' . esc_attr( gp_url( wp_make_link_relative( get_the_permalink( $event->id() ) ) ) ) . '">' . esc_html( $event->title() ) . '</a>', __( 'Translations', 'glotpress' ), $locale->english_name ) );
```

templates/event-translations-header.php
```diff
@@ -0,0 +1,63 @@
...
+	</ul>
+
+	<?php
+	if ( trim( $status, '/' ) !== 'waiting' ) {
+		?>
+		<a href="<?php echo esc_url( gp_url_join( gp_url(), wp_make_link_relative( get_the_permalink( $event->id() ) ), $locale->slug, 'waiting' ) ); ?>"><?php esc_html_e( 'Show only waiting translations', 'glotpress' ); ?></a>
+		<?php
+	} else {
+		?>
+		<a href="<?php echo esc_url( gp_url_join( gp_url(), wp_make_link_relative( get_the_permalink( $event->id() ) ), $locale->slug ) ); ?>"><?php esc_html_e( 'Show all contributed translations', 'glotpress' ); ?></a>
```
amieiro: Currently, in these URLs we don't have the `translations` in the URL. `https://glotpress.test/glotpress/events/{event-slug}/{language-slug}/waiting/`. This update adds it 
`https://glotpress.test/glotpress/events/{event-slug}/translations/{language-slug}/`

```suggestion
		<a href="<?php echo esc_url( gp_url_join( gp_url( wp_make_link_relative( get_the_permalink( $event->id() ) ) ), 'translations', $locale->slug, 'waiting' ) ); ?>"><?php esc_html_e( 'Show only waiting translations', 'glotpress' ); ?></a>
		<?php
	} else {
		?>
		<a href="<?php echo esc_url( gp_url_join( gp_url( wp_make_link_relative( get_the_permalink( $event->id() ) ) ), 'translations', $locale->slug ) ); ?>"><?php esc_html_e( 'Show all contributed translations', 'glotpress' ); ?></a>
```

templates/event.php
akirk: @psrpinto running out of time today, could you help add a new `Urls` method for this? Thanks!

psrpinto (approved):
Tested and seems to work well. Great feature!

PR #200, opened 2024-04-09
Allow to have co-hosts

### LINKED ISSUE #178: Allow to have co-hosts
Allow main host to add other users as co-hosts so that the co-hosts can also manage the event.

amieiro:
This PR adds: 
- A new list of attendees in the event's view, showing these users without contributions to the event.
- 2 buttons, available only for the admins and for the host, to add or remove the host permissions to each contributor or attendee.
- Some checks in the event view:
  - Doesn't show the contributors' section if the event doesn't have any contributor.
  - Doesn't show the attendees' section if the event doesn't have any attendee.
  - Doesn't show the stats, projects and stats summary sections if the event doesn't have any contribution.
- Some checks in the edit and delete actions for an event. Now, the host can edit and delete the event.

### Current event with contributors and attendees

image

### Future event with attendees but without contributors yet

image

### Same views with a user who is neither administrator nor host

image

image

Fixes https://github.com/WordPress/wporg-gp-translation-events/issues/178


assets/css/translation-events.css
```diff
@@ -83,19 +83,23 @@
 	font-weight: bold;
 }
 
-.event-contributors li {
+.event-contributors li, .event-atendees li {
```
akirk: Typo:
```suggestion
.event-contributors li, .event-attendees li {
```

templates/event.php
```diff
@@ -39,10 +41,48 @@
...
+								endif;
+								echo '</form>';
+							endif;
+							?>
 						</li>
 					<?php endforeach; ?>
 				</ul>
 			</div>
+	<div class="event-atendees">
+		<h2><?php esc_html_e( 'Atendees', 'gp-translation-events' ); ?></h2>
```
akirk:
```suggestion
		<h2><?php esc_html_e( 'Attendees', 'gp-translation-events' ); ?></h2>
```

templates/event.php
```diff
@@ -39,10 +41,48 @@
...
+		<small><?php esc_html_e( 'Users without contributions', 'gp-translation-events' ); ?></small>
+		<ul>
+			<?php foreach ( $attendees as $_user ) : ?>
+				<li class="event-atendee">
```
akirk:
```suggestion
				<li class="event-attendee">
```

assets/css/translation-events.css
```diff
@@ -83,19 +83,23 @@
...
+.event-contributors li, .event-attendees li {
 	display: inline-block;
 	list-style-type: none;
 	margin-right: 1em;
 	margin-bottom: .5em;
 	width: 15em;
 }
 
-.event-contributors li .avatar {
+.event-contributors li .avatar, .event-atendees li .avatar {
```
akirk:
```suggestion
.event-contributors li .avatar, .event-attendees li .avatar {
```

assets/css/translation-events.css
```diff
@@ -83,19 +83,23 @@
...
 	border-radius: 50%;
 	vertical-align: middle;
 }
 
+.event-atendees h2 {
```
akirk:
```suggestion
.event-attendees h2 {
```

templates/event.php
```diff
@@ -39,10 +41,48 @@
...
+									echo '<input type="submit" class="button is-secondary convert-to-host" value="Convert to host"/>';
+								endif;
+								echo '</form>';
+							endif;
+							?>
 						</li>
 					<?php endforeach; ?>
 				</ul>
 			</div>
+	<div class="event-atendees">
```
akirk:
```suggestion
	<div class="event-attendees">
```

assets/css/translation-events.css
akirk:
```suggestion
.event-attendees h2 {
```

includes/attendee/attendee-repository.php
```diff
@@ -128,4 +128,46 @@ function ( object $row ) {
...
 		);
 	}
+
+	/**
+	 * Add a user as host of an event.
+	 *
+	 * @param Attendee $attendee The attendee to add as host.
+	 * @return void
+	 */
+	public function add_host( Attendee $attendee ): void {
```
psrpinto: Instead of having this function, I think we could use the existing `insert_attendee()` (it already handles the `is_host` flag). Personally I see the host as an attendee who has a special status, instead of the host being a separate entity.

The calling code is then responsible for deciding whether the attendee being created is a host or not, something like:

```php
$attendee = new Attendee( 1, 2 );
$attendee->mark_as_host();
$attendee_repository->insert_attendee( $attendee );
```

includes/attendee/attendee-repository.php
```diff
@@ -128,4 +128,46 @@ function ( object $row ) {
...
+		$this->update_host( $attendee, 1 );
+	}
+
+	/**
+	 * Remove a user as host of an event.
+	 *
+	 * @param Attendee $attendee The attendee to remove as host.
+	 * @return void
+	 */
+	public function remove_host( Attendee $attendee ): void {
```
psrpinto: Similarly to my other comment about `add_host()`, I think we could use the existing `remove_attendee()` method instead of adding this one.

includes/attendee/attendee-repository.php
```diff
@@ -128,4 +128,46 @@ function ( object $row ) {
...
+	}
+
+	/**
+	 * Update the host status of an attendee.
+	 *
+	 * @param Attendee $attendee The attendee to update.
+	 * @param int      $is_host  Whether the attendee is a host.
+	 * @return void
+	 */
+	private function update_host( Attendee $attendee, int $is_host ) {
```
psrpinto: Similarly to my other comment about `add_host()`, I think we should have a generic `update_attendee( Attendee $attendee )` method that updates attendees, regardless of the attendee being a host or not.

includes/routes/user/host-event.php
```diff
@@ -0,0 +1,64 @@
...
+	 */
+	public function handle( int $event_id, int $user_id ): void {
+		$current_user = wp_get_current_user();
+		if ( ! $current_user->exists() ) {
+			$this->die_with_error( esc_html__( 'Only logged-in users can manage event\'s hosts.', 'gp-translation-events' ), 403 );
+		}
+
+		$current_user_attendee = $this->attendee_repository->get_attendee( $event_id, $current_user->ID );
+		if ( ! current_user_can( 'manage_options' ) && ! $current_user_attendee->is_host() ) {
+			$this->die_with_error( esc_html__( 'This use does not have permissions to manage event\'s hosts.', 'gp-translation-events' ), 403 );
```
psrpinto:
```suggestion
			$this->die_with_error( esc_html__( "This user does not have permissions to manage event's hosts.", 'gp-translation-events' ), 403 );
```

includes/routes/user/host-event.php
```diff
@@ -0,0 +1,64 @@
...
+
+		$event = $this->event_repository->get_event( $event_id );
+		if ( ! $event ) {
+			$this->die_with_404();
+		}
+
+		$affected_attendee = $this->attendee_repository->get_attendee( $event_id, $user_id );
+		if ( $affected_attendee instanceof Attendee ) {
+			if ( $affected_attendee->is_host() ) {
+				$this->attendee_repository->remove_host( $affected_attendee );
```
psrpinto: Instead of removing/adding I think we could just update the `is_host` property, something like:

```php
$affected_attendee = $this->attendee_repository->get_attendee( $event_id, $user_id );
if ( ! $affected_attendee ) {
	$affected_attendee = new Attendee( $event_id, $user_id );
        $this->attendee_repository->insert_attendee( $affected_attendee );
}

if ( $affected_attendee->is_host() ) {
	$affected_attendee->mark_as_non_host();
} else {
	$affected_attendee->mark_as_host();
}

$this->attendee_repository->update_attendee( $affected_attendee );
```

psrpinto (commented):
I realise this PR is still in draft but I was taking a look and figured I would provide some early feedback.

includes/attendee/attendee-repository.php
amieiro: The user is attending to the event, but it is not marked as host. If I do this, we will have two rows with the same event_id and user_id.

includes/attendee/attendee-repository.php
amieiro: The `remove_attendee()`removes the user from attending to the event, and what I want to do is to remove her as host, maintaining her assistance to the event.

includes/attendee/attendee-repository.php
amieiro: Maybe the best approach is to have:
- a `add_host( Attendee $attendee )` method.
- a `remove_host( Attendee $attendee )` method.
- add a new `update_attendee( Attendee $attendee )`, used by the previous methods, removing the `update_attendee( Attendee $attendee )` method.

What do you think about?

includes/routes/user/host-event.php
amieiro: I have implemented this approach. Thank you for the feedback.

includes/attendee/attendee-repository.php
amieiro: Mark as resolved, because I have implemented [this suggestion](https://github.com/WordPress/wporg-gp-translation-events/pull/200#discussion_r1559642033).

templates/event.php
```diff
@@ -31,122 +33,163 @@
...
 							<a href="<?php echo esc_url( get_author_posts_url( $contributor->ID ) ); ?>"><?php echo esc_html( get_the_author_meta( 'display_name', $contributor->ID ) ); ?></a>
+							<?php
+							if ( ( $user_is_attending && $attendee->is_host() ) || current_user_can( 'manage_options' ) ) :
+								$_attendee_repo = new Attendee_Repository();
+								$_attendee      = $_attendee_repo->get_attendee( $event_id, $contributor->ID );
+								echo '<form class="add-remove-user-as-host" method="post" action="' . esc_url( gp_url( "/events/host/$event_id/$contributor->ID" ) ) . '">';
+								if ( $_attendee->is_host() ) :
+									echo '<input type="submit" class="button is-primary remove-as-host" value="Remove as host"/>';
+								else :
+									echo '<input type="submit" class="button is-secondary convert-to-host" value="Convert to host"/>';
```
akirk:
```suggestion
									echo '<input type="submit" class="button is-secondary convert-to-host" value="Make co-host"/>';
```

templates/event.php
```diff
@@ -31,122 +33,163 @@
...
-				<?php
-				foreach ( explode( ',', $row->locales ) as $_locale ) {
-					$_locale = \GP_Locales::by_slug( $_locale );
-					?>
-					<a href="<?php echo esc_url( gp_url_project_locale( $row->project, $_locale, 'default' ) ); ?>"><?php echo esc_html( $_locale->english_name ); ?></a>
+		<?php endif; ?>
+		<?php if ( ! empty( $attendees ) ) : ?>
+			<div class="event-attendees">
+				<h2><?php esc_html_e( 'Attendees', 'gp-translation-events' ); ?></h2>
+				<small><?php esc_html_e( 'Users without contributions', 'gp-translation-events' ); ?></small>
```
akirk: Can we just remove this? It visually doesn't look nice and I am not sure we need this explanation.
```suggestion
```

akirk (commented):
I tested promotion and demotion of hosts and that worked well. Also how non-hosts can't see the button worked. I have several observations, though:
- You shouldn't be able to modify yourself. Right now, as not the event author, I can remove myself from being a Co-host and then I lose editing ability.
- That should prevent the scenario when there are no co-hosts at all. Right now I can remove all co-hosts and then there is no co-host anymore.
- We need to decide if the post author can be removed as a host by anyone or if we should just hide the button for them (and prevent it server-side too).
- We should reveal the hosts here:
cohostdisplay
- On the "My Events" page, we show a "Events I have created" and when you are a host, you don't see the event there. What about we renamed it to "Events I host" and show the event if you are a co-host.

Lots of small details that I am also fine if we added them in a follow-up, but it's some necessary polish. Great job!

templates/event.php
```diff
@@ -31,122 +33,163 @@
...
-			<?php foreach ( $projects as $project_name => $row ) : ?>
-			<li class="event-project" title="<?php echo esc_html( str_replace( ',', ', ', $row->locales ) ); ?>">
-				<a href="<?php echo esc_url( gp_url_project( $row->project ) ); ?>"><?php echo esc_html( $project_name ); ?></a> <small> to
-				<?php
-				foreach ( explode( ',', $row->locales ) as $_locale ) {
-					$_locale = \GP_Locales::by_slug( $_locale );
-					?>
-					<a href="<?php echo esc_url( gp_url_project_locale( $row->project, $_locale, 'default' ) ); ?>"><?php echo esc_html( $_locale->english_name ); ?></a>
+		<?php endif; ?>
+		<?php if ( ! empty( $attendees ) ) : ?>
```
akirk: We should only show it to hosts.

templates/event.php
```diff
@@ -31,122 +33,163 @@
...
+			<div class="event-attendees">
+				<h2><?php esc_html_e( 'Attendees', 'gp-translation-events' ); ?></h2>
+				<small><?php esc_html_e( 'Users without contributions', 'gp-translation-events' ); ?></small>
+				<ul>
+					<?php foreach ( $attendees as $_user ) : ?>
+						<li class="event-attendee">
+							<a href="<?php echo esc_url( get_author_posts_url( $_user->ID ) ); ?>"><?php echo get_avatar( $_user->ID, 48 ); ?></a>
```
akirk, line 90: @trymebytes when this PR is merged, please highlight first-time contributors here. Can we use 🎉 as a avatar overlay (like in the bottom right corner?). Let's also change it in the contributors list so that it's consistent.

akirk (dismissed):
I think this still needs to be changed before the PR can be merged:
- Attendees should only be visible to hosts. Clicking attend on an event should not give a user exposure, only contribution.

templates/event.php
amieiro: I am not sure about this. Another sites, like meetup.com, show all attendees in public. Take a look at this [Elementor meetup](https://www.meetup.com/es-ES/elementor-espana-meetup-group/events/299627457/).

It is ok for me to show the attendees' list in public, and it is interesting when you look for information about the event if you are going to be alone or with more people in the event.

image

akirk: I understand your point. I really want to encourage contribution and only show people on the page when they have contributed, not only when they have made a free pledge.

Maybe we could show the list of attendees only before the event starts and then hide it?

amieiro: This is a better approach. What about showing them just until the event finishes?
Imagine a contributor day, starting at 9:00 am and finishing the same day at 4:00 pm. If a user arrives at 9:30, it will be interesting to show her the attendees. And after 4:00 pm, we only show them the contributors. In fact, it could be interesting to ping some users who have forgotten to show up at the event.

akirk: Yeah but for that wouldn't it be enough for the host to see the attendees? Probably worth testing out with real life events as it's easy enough to change (even during the event).

I think we should try to avoid that some people just always attend all events and they will always appear on all event attendees lists. That's my reasoning behind hiding the attendees list.

amieiro: Yes, it would be enough for the host to see the attendees. But sometimes you decide to go to an event based on who is there, or if there are a few or a lot of people, so I think until the event is finalized, we should show this information to everyone. Then hide the attendees who have not contributed, and that's it.

It is easy to change it, so we can test it like this and if we have any complaints, we can change it according to the new requirements.

psrpinto: I tend to agree that only showing the user on the page if they contribute could encourage more contribution.

At the same time, I think there's also value in displaying the list of attendees. Showing that an event has attendees could be encouragement to other people to also attend.

One solution that would address both these needs would be to display the list of attendees in another page (we could also consider introducing tabs on the event details page). This page/tab could then be linked from the main event page, for example below the "Attend" button.

includes/routes/user/host-event.php
```diff
@@ -0,0 +1,67 @@
...
+		// The user is attending to the event, so if I don't find the attendee, I won't create it.
+		if ( $affected_attendee instanceof Attendee ) {
+			if ( $affected_attendee->is_host() ) {
+				$affected_attendee->mark_as_non_host();
+			} else {
+				$affected_attendee->mark_as_host();
+			}
+
+			$this->attendee_repository->update_attendee( $affected_attendee );
+		}
```
psrpinto, line 65: Currently, if the user is not an attendee of the event, the request will succeed but the user was not made a host of the event. Maybe we could throw an error if the user is not attending the event. Or we could make it so that an attendee is created if this route is called for a user who is not yet an attendee.

psrpinto (approved):

includes/attendee/attendee-repository.php
```diff
@@ -26,6 +26,27 @@ public function insert_attendee( Attendee $attendee ): void {
...
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.NoCaching
+		$wpdb->update(
+			"{$gp_table_prefix}event_attendees",
+			array( 'is_host' => $attendee->is_host() ? 1 : 0 ),
+			array(
+				'event_id' => $attendee->event_id(),
+				'user_id'  => $attendee->user_id(),
+			)
+		);
```
psrpinto, line 48: We need to re-enable the rules we disabled above.


```suggestion
		);
		// phpcs:enable
```

psrpinto (changes requested):

templates/event.php
trymebytes: While considering a public list of attendees, there should be an option for attendees to choose not to be listed on the public list.

templates/event.php
trymebytes: Sure, I'll come up with a mockup

trymebytes:
In a follow up PR, it would be interesting to add revision history to track changes made by any of the co-hosts.

includes/attendee/attendee-repository.php
amieiro: Done. Thank you!

psrpinto (approved):

includes/routes/user/host-event.php
amieiro: I think if the user is not attending the event, the $affected_attendee won't an instance of Attendee, so I think this code won't be executed. Can you double-check it, but I think this is the main reason of this `if` check?

includes/stats-calculator.php
```diff
@@ -233,6 +233,43 @@ public function get_attendees_not_contributing( int $event_id ): array {
 		return $attendees_not_contributing;
 	}
 
+	/**
+	 * Get the hosts' users for an event.
+	 *
+	 * @param int $event_id The id of the event.
+	 * @return array[WP_User] The hosts of the event.
+	 */
+	public function get_hosts( int $event_id ): array {
```
psrpinto: I think we should place this method in the attendee repository, something like:

```php
/**
 * @return Attendee[] Attendees of the event who are hosts.
 */
public function get_hosts( int $event_id ): array;
```

There's already a `get_attendees()` method in the repository (though it's currently not implemented), so I think it would make sense to put the `get_hosts()` method there too.

amieiro:
Some changes.

The current user, if she is a host, can't modify her host status:

image

Another co-host can't modify the host status of the creator:

image

I added a list with all the host:

image

In the "My Events" page, I added a new list (Events I host) with the events the current user is hosting. I didn't delete the "Events I have created", because we could have situations where an administrator creates an event, makes another people co-host and stops attending the event. We have an edge-situation, without any host, but the creator always have permission to edit the event.

image

includes/stats-calculator.php
amieiro: Done. Thank you for the feedback!

includes/attendee/attendee-repository.php
```diff
@@ -114,6 +115,43 @@ public function get_attendees( int $event_id ): array { // phpcs:ignore Generic.
 		return array();
 	}
 
+	/**
+	 * Get the hosts' users for an event.
+	 *
+	 * @param int $event_id The id of the event.
+	 * @return array[WP_User] The hosts of the event.
+	 */
+	public function get_hosts( int $event_id ): array {
```
psrpinto, line 124: I think this method should return an array of `Attendee`. The idea with the repositories is that they always "deal with" a specific entity.

For example in the `Event_Repository`, methods that modify an event take an `Event` instance, and methods that return events return instances of `Event`.

I think it would make sense to continue to follow that pattern here, so this method would return an array of `Attendee`. Then you would call the method with something like:

```php
$host_attendees = $attendee_repository->get_hosts( $event_id );

// If you would need an instance of WP_user, you would:
$host = new WP_User( $host->attendeees[0]->user_id() );
```

includes/attendee/attendee-repository.php
```diff
@@ -114,6 +115,43 @@ public function get_attendees( int $event_id ): array { // phpcs:ignore Generic.
...
+			",
+				array(
+					$event_id,
+				)
+			)
+		);
+		// phpcs:enable
+
+		$event    = get_post( $event_id );
+		$host_ids = array_unique( array_merge( $host_ids, array( $event->post_author ) ) );
```
psrpinto: The post author is [automatically added as a host](https://github.com/WordPress/wporg-gp-translation-events/blob/trunk/wporg-gp-translation-events.php#L220) when the event is published, so I don't think we need to do this here.

(Actually, currently it's not the post_author who is being added as a host, but the user who publishes the event, but as discussed previously, we should fix that).

psrpinto (commented):
Looks good! Just a have a couple of remarks

includes/attendee/attendee-repository.php
amieiro: The problem with the author was resolved [here](https://github.com/WordPress/wporg-gp-translation-events/pull/200/files#diff-93ac378e918721e6698187d0123e47b161f8fe273064813e9cdef3433a9165a8L215).

If the event creator left the event, and we don't have another host, we will have a list of empty hosts. And if the `get_hosts` should return an array of `Attendee`, the creator won't appear there. So we need an array of `WP_User`, not of `Attendee`. What do you think the best place to put this method? Return it to the `Stats_Calculator`?

psrpinto: > The problem with the author was resolved [here](https://github.com/WordPress/wporg-gp-translation-events/pull/200/files#diff-93ac378e918721e6698187d0123e47b161f8fe273064813e9cdef3433a9165a8L215).

Nice! I had missed that.

psrpinto: The way I see it is that, now that we're tracking who is a host in the attendees table, the author of the event no longer has any relation to who is a host. The author is simply the user who created the event, and the hosts are the entries in the attendees table with `is_host` set to true.

To me this means that if there are no entries in the attendees table with `is_host` set to `true`, then the `get_hosts()` method should return an empty array.

However, as you mention, it should not be possible to have an event with no hosts. In the same way that we add the post_author as host when the event is published (to make sure the event has a host), we would also need to prevent the last host from un-attending the event.

This would mean that it's guaranteed that an event has at least one host, so the `get_hosts()` method will never return an empty array.

To me the solution would be to add a check in the attend/un-attend code that prevents the last host from un-attending. Do you think this would make sense?

amieiro: Yes, I like this approach. We can disable the "You're attending" button, with a small text saying "You are the unique host, so you can't leave the event. Please, add a new host before". I will work on this approach.

akirk: Isn't that also solved by the fact that the host cannot modify themselves, thus remove themselves? Thus hosts should not be allowed to unattend an event unless someone else releaves them from being a host.

akirk: I don't think we should make this just for the last host but for any host. If you're a host, you cannot leave the event.

amieiro: The host can't modify herself, but she can stop attending to the event, so she will remove the row in the `attendees` table.

I think we could have situations where one event creator or a host wants to leave the event, and I think we should all her to do this, except she is the last one. One situation:
- A user creates an event that she will not attend: a physical event, an event in a different locale,...
- She adds a new host.
- She left the event.

For me, the most flexible approach is to allow leaving the event, expect the user is the last host, and avoid using the event creator to determine the hosts.

amieiro:
I have made some improvements, and the current situation is:

- The event creator will always have edition permissions, so she may not be a host and could manage the event, including to add and remove hosts.
- The current user, if she is a host, can't modify her status as host.
- The current user, if she is the unique host, can't click on the `You're attending` button: I have disabled the button and check it in the function that manage the request.
- If we have only one host, the another user with permissions (the event creator or an administrator) can't click on the `Remove as host` button to avoid having an event without a host: I have disabled the button and check it in the function that manage the request.

includes/attendee/attendee.php
```diff
@@ -37,7 +37,20 @@ public function is_host(): bool {
 		return $this->is_host;
 	}
 
+	/**
+	 * Check if the user is the unique host of the event.
+	 *
+	 * @return bool
+	 */
+	public function is_unique_host(): bool {
+		return ( new Attendee_Repository() )->is_unique_host( $this->event_id, $this->user_id );
```
psrpinto: I think that when using the repository pattern, the "entity" classes (`Event`, `Attendee`, etc) don't typically call other repositories, they are only "containers" for data, and their methods operate solely on that data.

So, if you would want to have a `is_foo()` method, the value of `foo` would need to be passed to the constructor, and then that method would `return $this->foo;`.

However, in this case, I don't think `is_unique_host` should be passed to the constructor, because its value depends on data that does not "belong" to this attendee (it depends on other attendees leaving the event), so you could be in a situation where `is_unique_host()` returns false when it should in fact return `true`, because an attendee left after the instance of `Attendee` was created.

I think we could remove the `is_unique_host()` from the `Attendee`, and call it on the repository instead:

```php
// attend-event.php

if ( $this->attendee_repository->is_unique_host( $event_id, $user_id ) ) {
    // Can't remove last host
}
```

Alternatively, because what we want to know is whether the event has a single attendee, we could instead use the already existing `get_hosts()` method:

```php
if ( 1 === count( $this->attendee_repository->get_hosts( $event_id ) ) ) {
    // Can't remove last host
}
```

This would mean we could remove the `is_unique_host()` from the repository, and from `Attendee`.

psrpinto (commented):
This is looking great, well done @amieiro !

includes/attendee/attendee.php
amieiro: Thank you for the feedback. Done!

templates/event.php
amieiro: I added it to the contributors' list, because I needed to fix a conflict merge from trunk. After this, I added it to the attendees' list [here](https://github.com/WordPress/wporg-gp-translation-events/pull/200/commits/eb99f44120e3cfb6e4ac9266e99c9eaa508c90fe).

image

templates/event.php
```diff
@@ -42,119 +50,187 @@
...
+		<?php endif; ?>
+		<?php if ( ! empty( $attendees ) && ( ! $event->end()->is_in_the_past() || ( ( $attendee instanceof Attendee && $attendee->is_host() ) || current_user_can( 'manage_options' ) ) ) ) : ?>
+			<div class="event-attendees">
+				<h2><?php esc_html_e( 'Attendees', 'gp-translation-events' ); ?></h2>
+				<ul>
+					<?php foreach ( $attendees as $_user ) : ?>
+						<li class="event-attendee">
+							<a href="<?php echo esc_url( get_author_posts_url( $_user->ID ) ); ?>"><?php echo get_avatar( $_user->ID, 48 ); ?></a>
+							<a href="<?php echo esc_url( get_author_posts_url( $_user->ID ) ); ?>"><?php echo esc_html( get_the_author_meta( 'display_name', $_user->ID ) ); ?></a>
+							<?php if ( $stats_calculator->is_first_time_contributor( $event_start, $contributor->ID ) ) : ?>
```
trymebytes:
```suggestion
							<?php if ( $stats_calculator->is_first_time_contributor( $event_start, $_user->ID ) ) : ?>
```
This throws a notice `Trying to get property 'ID' of non-object in`

trymebytes (changes requested):

trymebytes (approved):
Tested and works well for me!

trymebytes:
A quick question, should an event creator be able to edit an event after they have been removed as a host?

amieiro:
> A quick question, should an event creator be able to edit an event after they have been removed as a host?

Yes, I think and event creator and an administrator should be able to edit the event even if they are not a host, to be able to manage last minute updates. But as it is easy to change, we can update this behavior in the future if we have information that makes to change our mind.

psrpinto (approved):
Tested and it works, well done!

One thing I found a bit weird is that another host can remove me as a host but I can't do it myself. Typically I would expect myself to have at least the same permissions as other users in my "group" (hosts). Maybe something we can fix in another PR.

akirk (approved):
Works well for me! I have a similar problem as @psrpinto which we already discussed in chat.

amieiro:
> Tested and it works, well done!
> 
> One thing I found a bit weird is that another host can remove me as a host but I can't do it myself. Typically I would expect myself to have at least the same permissions as other users in my "group" (hosts). Maybe something we can fix in another PR.

We discuss this behavior and this was we decided. But with the different changes we did in this PR, maybe in the future we can enable this button for yourself only if you are not the last host. And we will always have the creator or one administrator to update the hosts. 
Something similar was reported by @akirk (see image below), related with an event not being a host, not being able to remove the unique host as co-host. 

image

PR #202, opened 2024-04-10
Count contributions prior to user joining event

### LINKED ISSUE #161: When joining an event late, make the earlier contributions after event start count
It should not matter if you join an event late, all your contributions during the event time should count. This should only be true for events lasting 24hrs max.

Example for the SQL: 
```sql
INSERT IGNORE INTO gp_event_actions
( event_id, original_id, locale, user_id, happened_at, action )
SELECT 123, t.original_id, ts.slug, t.user_id, t.date_added, "created"
FROM gp_translations t, gp_translation_sets ts
WHERE t.user_id = 146 AND t.translation_set_id = ts.id AND date_added > "2024-03-06"
```

psrpinto:
Fix #161 

## Summary of changes

- Move stats-related code to a `Stats` namespace
- Add `is_active()` and `is_past()` methods to `Event`
- Add a `Stats_Importer` for importing stats
- Add `Stats_Importer::import_for_user_and_event( $user_id, $event_id )`
- When a user attends an event that is active at that moment, call `import_for_user_and_event()`
- Make sure it's not possible to attend/un-attend past events

## Testing instructions

1. Login as User A
2. Create a translation
1. Login as User B
3. Create an event and make it start now
4. Login as User A
6. Navigate to the page of the event you created above, and click the "Attend" button
7. Create a translation
8. You should now be listed as a contributor, and the stats table should show **a single translation** (it should not count the first translation, created above in 2., as the event hadn't started yet)

### DIFF
--- a/includes/event/event.php
+++ b/includes/event/event.php
@@ -2,6 +2,7 @@
 
 namespace Wporg\TranslationEvents\Event;
 
+use DateTimeImmutable;
 use DateTimeZone;
 use Exception;
 use Throwable;
@@ -84,6 +85,11 @@ class Event {
 
 	public function end(): Event_End_Date {
 		return $this->end;
+	}
+
+	public function is_active(): bool {
+		$now = new DateTimeImmutable( 'now', new DateTimeZone( 'UTC' ) );
+		return $now >= $this->start->utc() && $now < $this->end->utc();
 	}
 
 	public function timezone(): DateTimeZone {
--- a/includes/event/event.php
+++ b/includes/event/event.php
@@ -93,8 +93,7 @@ class Event {
 	}
 
 	public function is_past(): bool {
-		$now = new DateTimeImmutable( 'now', new DateTimeZone( 'UTC' ) );
-		return $this->end->utc() < $now;
+		return $this->end->is_in_the_past();
 	}
 
 	public function timezone(): DateTimeZone {
--- a/includes/event/event.php
+++ b/includes/event/event.php
@@ -92,6 +92,11 @@ class Event {
 		return $now >= $this->start->utc() && $now < $this->end->utc();
 	}
 
+	public function is_past(): bool {
+		$now = new DateTimeImmutable( 'now', new DateTimeZone( 'UTC' ) );
+		return $this->end->utc() < $now;
+	}
+
 	public function timezone(): DateTimeZone {
 		return $this->timezone;
 	}
--- a/includes/stats/stats-importer.php
+++ b/includes/stats/stats-importer.php
@@ -2,11 +2,40 @@
 
 namespace Wporg\TranslationEvents\Stats;
 
+use Wporg\TranslationEvents\Event\Event;
+
 class Stats_Importer {
 	/**
 	 * Imports the contributions a user made while a given event was active.
 	 */
-	public function import_for_user_and_event( int $user_id, int $event_id ): void {
-		// TODO.
+	public function import_for_user_and_event( int $user_id, Event $event ): void {
+		global $wpdb, $gp_table_prefix;
+
+		// phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.NoCaching
+		$wpdb->query(
+			$wpdb->prepare(
+				"
+				insert ignore into {$gp_table_prefix}event_actions
+				    (event_id, user_id, original_id, locale, action, happened_at)
+				select %d, t.user_id, t.original_id, ts.slug, %s, t.date_added
+				from {$gp_table_prefix}translations t,
+				     {$gp_table_prefix}translation_sets ts
+				where t.user_id = %d
+				  and t.translation_set_id = ts.id
+				  AND date_added >= %s
+				  AND date_added <= %s
+				",
+				array(
+					'event_id'          => $event->id(),
+					'action'            => Stats_Listener::ACTION_CREATE,
+					'user_id'           => $user_id,
+					'date_added_after'  => $event->start()->utc()->format( 'Y-m-d H:i:s' ),
+					'date_added_before' => $event->end()->utc()->format( 'Y-m-d H:i:s' ),
+				),
+			),
+		);
+		// phpcs:enable
 	}
 }
--- a/includes/stats/stats-importer.php
+++ b/includes/stats/stats-importer.php
@@ -24,7 +24,7 @@ class Stats_Importer {
 				     {$gp_table_prefix}translation_sets ts
 				where t.user_id = %d
 				  and t.translation_set_id = ts.id
-				  and t.status in ('current', 'waiting', 'changesrequested', 'fuzzy' )
+				  and t.status in ( 'current', 'waiting', 'changesrequested', 'fuzzy' )
 				  and date_added between %s and %s
 				",
 				array(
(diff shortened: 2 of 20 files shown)


includes/event/event.php
```diff
@@ -86,6 +87,16 @@ public function end(): Event_End_Date {
...
 	}
 
+	public function is_active(): bool {
+		$now = new DateTimeImmutable( 'now', new DateTimeZone( 'UTC' ) );
+		return $now >= $this->start->utc() && $now < $this->end->utc();
+	}
+
+	public function is_past(): bool {
+		$now = new DateTimeImmutable( 'now', new DateTimeZone( 'UTC' ) );
+		return $this->end->utc() < $now;
```
akirk:
```suggestion
		return $this->end->is_in_the_past();
```

templates/event.php
```diff
@@ -7,6 +7,8 @@
 
 use Wporg\TranslationEvents\Event\Event_End_Date;
 use Wporg\TranslationEvents\Event\Event_Start_Date;
+use Wporg\TranslationEvents\Stats\Event_Stats;
+use Wporg\TranslationEvents\Stats\Stats_Row;
```
akirk, line 15: What's the reason for these? Are they added by PhpStorm?

tests/stats/stats-importer.php
```diff
@@ -0,0 +1,74 @@
...
+
+		// Create a translation before the event starts, which should not be imported.
+		$this->translation_factory->create( $user_id );
+
+		// Wait for the event to start.
+		// It needs to be 2 seconds because the event starts at "now -1 second".
+		// TODO: This makes test runs take longer, we should find another way.
+		sleep( 2 );
+
+		$event_id = $this->event_factory->create_active();
```
akirk:
```suggestion
		$event_id = $this->event_factory->create_active( array(), new DateTimeImmutable( '5 minutes ago', new DateTimeZone( 'UTC' ) ) );
```

akirk (approved):
Works well! I took the liberty to add fb3ddfc as it verifies that translations added before the event are not added retroactively.

akirk:
Apologies, you already had a translation before the event started but it was made ineffective by my other suggestion to just create the event in the past. Fixed it now.

templates/event.php
psrpinto: PhpStorm complains indeed if these aren't there. I actually thought they were required but I just tested without and it works correctly. How can it even work without it?

Because we're using namespaces, without `use` directives, PHP should look for those types in the namespace of the current file, and not find them. I don't understand why it works without the `use` directives. 🤔

psrpinto:
Thanks for removing the `sleep()`! I couldn't figure out how to manipulate the `date_added` of the translation, so that's why I resorted to using the `sleep()`. Much better now.

includes/stats/stats-importer.php
```diff
@@ -0,0 +1,41 @@
...
+
+		// phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.NoCaching
+		$wpdb->query(
+			$wpdb->prepare(
+				"
+				insert ignore into {$gp_table_prefix}event_actions
+				    (event_id, user_id, original_id, locale, action, happened_at)
+				select %d, t.user_id, t.original_id, ts.slug, %s, t.date_added
```
trymebytes:
```suggestion
				select %d, t.user_id, t.original_id, ts.locale, %s, t.date_added
```
This throws a notice `Notice: Trying to get property 'english_name' of non-object`

trymebytes (dismissed):

includes/stats/stats-importer.php
psrpinto: Could you describe the steps you took so I can reproduce?

trymebytes: These were the steps I took;

1. Log in as User A in browser A and create an event
2. Log in as User B in browser B and translate a string
3. Visit the event page for event created in No. 1 above and click 'Attend'.

The error shows in the stats table and projects section

includes/stats/stats-importer.php
```diff
@@ -0,0 +1,41 @@
...
+			$wpdb->prepare(
+				"
+				insert ignore into {$gp_table_prefix}event_actions
+				    (event_id, user_id, original_id, locale, action, happened_at)
+				select %d, t.user_id, t.original_id, ts.locale, %s, t.date_added
+				from {$gp_table_prefix}translations t,
+				     {$gp_table_prefix}translation_sets ts
+				where t.user_id = %d
+				  and t.translation_set_id = ts.id
+				  AND date_added >= %s
```
amieiro: I think we only want to store the translation strings in `current`, `waiting,` `changesrequested` or `fuzzy` status. I think we should avoid storing the strings in `rejected` and `old` status. 
```suggestion
				  AND ( t.status = 'current' OR t.status = 'waiting' OR t.status = 'changesrequested' OR status = 'fuzzy' ) 
				  AND date_added >= %s
```

amieiro:
In the `translations` table (GlotPress plugin), if you have a row with `user_id` != `user_id_last_modified`, this means this string has been reviewed by a validator, so we should add this work to the `event_actions` table, only if the validator is assisting to the event.

includes/stats/stats-importer.php
```diff
@@ -0,0 +1,42 @@
...
+			$wpdb->prepare(
+				"
+				insert ignore into {$gp_table_prefix}event_actions
+				    (event_id, user_id, original_id, locale, action, happened_at)
+				select %d, t.user_id, t.original_id, ts.locale, %s, t.date_added
+				from {$gp_table_prefix}translations t,
+				     {$gp_table_prefix}translation_sets ts
+				where t.user_id = %d
+				  and t.translation_set_id = ts.id
+				  and ( t.status = 'current' or t.status = 'waiting' or t.status = 'changesrequested' or status = 'fuzzy' )
```
akirk: The last `status` is missing the `t.` but we can also write:
```suggestion
				  and t.status in ('current', 'waiting', 'changesrequested', 'fuzzy' )
```

includes/stats/stats-importer.php
```diff
@@ -0,0 +1,42 @@
...
+				  and date_added >= %s
+				  AND date_added <= %s
```
akirk: Here we've got one `and` and one `AND`. It can be also written as:
```suggestion
				  and date_added between %s and %s
```
https://dev.mysql.com/doc/refman/8.0/en/functions.html#operator_between
> this is equivalent to the expression (min <= expr AND expr <= max)

psrpinto: Great suggestions, thanks!

akirk (approved):

includes/stats/stats-importer.php
```diff
@@ -0,0 +1,41 @@
...
+			$wpdb->prepare(
+				"
+				insert ignore into {$gp_table_prefix}event_actions
+				    (event_id, user_id, original_id, locale, action, happened_at)
+				select %d, t.user_id, t.original_id, ts.locale, %s, t.date_added
+				from {$gp_table_prefix}translations t,
+				     {$gp_table_prefix}translation_sets ts
+				where t.user_id = %d
+				  and t.translation_set_id = ts.id
+				  and t.status in ('current', 'waiting', 'changesrequested', 'fuzzy' )
```
akirk: Nit:
```suggestion
				  and t.status in ( 'current', 'waiting', 'changesrequested', 'fuzzy' )
```

PR #206, opened 2024-04-15
Optimise retrieveing list of hosts

psrpinto:
Just a small optimisation so that we don't do n+1 queries when retrieving the list of hosts.


includes/attendee/attendee-repository.php
```diff
@@ -142,8 +144,10 @@ public function get_hosts( int $event_id ): array {
 		// phpcs:enable
 
 		$hosts = array();
-		foreach ( $host_ids as $host_id ) {
-			$hosts[] = $this->get_attendee( $event_id, $host_id );
+		foreach ( $rows as $row ) {
+			$host = new Attendee( $row->event_id, $row->user_id );
+			$host->mark_as_host();
```
akirk, line 149: What about we constructed the attendee with all fields (needs a change in the constructor)?

```suggestion
			$hosts[] = new Attendee( $row->event_id, $row->user_id, $row->is_host );
```

"Mark as host" sounds like it would invoke a db activity.

akirk (commented):
Nice, just an idea of how to construct the Attendees object.

includes/attendee/attendee-repository.php
psrpinto: I tend to agree and I think there could be some other improvements we can do to the Attendee/Host thing, so I opened #208. Just so we can get this one merged quickly :)

akirk (approved):

PR #211, opened 2024-04-15
Event permissions

psrpinto:
WIP


psrpinto:
Discussed with @akirk and will implement this using WordPress capabilities. Will do so in https://github.com/WordPress/wporg-gp-translation-events/pull/215.

PR #212, opened 2024-04-15
Cohosts: Fix some edge cases and simplify their management

amieiro:
We had an edge case where and WordPress `admin` was shown as `event creator` and not as an attendee.

image

This PR improve and simplify the different situations. As a WordPress admin and an event creator always can edit the event, adding/removing cohosts, this PR:
- Removes the restriction of having one mandatory cohost. If we remove them all, we will always have an administrator or the event creator to be able to add them back.
- Allows a cohost to remove herself from this role.
- Allows all users (attendee, contributor, host, administrator) to attend or stop attending an event at any time.
- The next users/roles will always be able to add/remove cohosts: a cohost, an administrator and the event creator.

This PR also add a title with the `This user is starting to translate.` text to the  🎉 element, so when a user hover this element, she will see this explanation.
 
image


trymebytes:
Seems like an event can exist without a host and the host field empty. I don't think we should allow this, perhaps if there are no co-hosts, the event creator should show as the host (in the host field below the event title)
Screenshot 2024-04-16 at 08 30 59

amieiro:
> * Removes the restriction of having one mandatory cohost. If we remove them all, we will always have an administrator or the event creator to be able to add them back.

It is a design decision: This PR removes the restriction of having one mandatory cohost. If we remove them all, we will always have an administrator or the event creator to be able to add them back.

trymebytes:
Yes, I agree but I think that host field should not be left empty. Maybe if there is no host, we could have something like this;
Screenshot 2024-04-16 at 10 21 42

templates/event.php
```diff
@@ -50,30 +50,22 @@
 							<a href="<?php echo esc_url( get_author_posts_url( $contributor->ID ) ); ?>"><?php echo get_avatar( $contributor->ID, 48 ); ?></a>
 							<a href="<?php echo esc_url( get_author_posts_url( $contributor->ID ) ); ?>"><?php echo esc_html( get_the_author_meta( 'display_name', $contributor->ID ) ); ?></a>
 							<?php if ( $stats_calculator->is_first_time_contributor( $event_start, $contributor->ID ) ) : ?>
-								<span class="first-time-contributor-tada"></span>
+								<span class="first-time-contributor-tada" title="<?php esc_html_e( 'This user is starting to translate.', 'gp-translation-events' ); ?>"></span>
```
akirk: What if we changed it to this?
```suggestion
								<span class="first-time-contributor-tada" title="<?php esc_html_e( 'New Translation Contributor', 'gp-translation-events' ); ?>"></span>
```

templates/event.php
```diff
@@ -50,30 +50,22 @@
 							<a href="<?php echo esc_url( get_author_posts_url( $contributor->ID ) ); ?>"><?php echo get_avatar( $contributor->ID, 48 ); ?></a>
```
akirk: Can we add classes to make it easier to style via CSS?
```suggestion
							<a href="<?php echo esc_url( get_author_posts_url( $contributor->ID ) ); ?>" class="avatar"><?php echo get_avatar( $contributor->ID, 48 ); ?></a>
```

templates/event.php
```diff
@@ -50,30 +50,22 @@
...
 							<a href="<?php echo esc_url( get_author_posts_url( $contributor->ID ) ); ?>"><?php echo esc_html( get_the_author_meta( 'display_name', $contributor->ID ) ); ?></a>
```
akirk:
```suggestion
							<a href="<?php echo esc_url( get_author_posts_url( $contributor->ID ) ); ?>" class="name"><?php echo esc_html( get_the_author_meta( 'display_name', $contributor->ID ) ); ?></a>
```

templates/events-header.php
```diff
@@ -39,20 +39,29 @@
...
+					?>
+					&nbsp;<a href="<?php echo esc_attr( get_author_posts_url( $user->ID ) ); ?>"><?php echo esc_html( get_the_author_meta( 'display_name', $user->ID ) ); ?></a>
+					<?php
 				endif;
 				?>
 				<?php foreach ( $hosts as $host ) : ?>
-					<?php $user = get_userdata( $host->user_id() ); ?>
-					&nbsp;<a href="<?php echo esc_attr( get_author_posts_url( $user->ID ) ); ?>"><?php echo esc_html( get_the_author_meta( 'display_name', $user->ID ) ); ?></a>
+					<?php $_user = get_userdata( $host->user_id() ); ?>
+					&nbsp;<a href="<?php echo esc_attr( get_author_posts_url( $_user->ID ) ); ?>"><?php echo esc_html( get_the_author_meta( 'display_name', $_user->ID ) ); ?></a>
```
psrpinto: I think the `$_user` variable is not needed

```suggestion
					&nbsp;<a href="<?php echo esc_attr( get_author_posts_url( $host->user_id() ) ); ?>"><?php echo esc_html( get_the_author_meta( 'display_name', $host->user_id() ) ); ?></a>
```

akirk (approved):
Everything I tested works. Feels much more predictable now, well done!

templates/events-header.php
amieiro: Good catch! Thank you.

PR #214, opened 2024-04-16
New translation contributor

### LINKED ISSUE #213: Determine new translation contributors by number of translations submitted.
In the first iteration of determining who a new translation contributor(previously known as first-time contributor) is, we used the following criteria;
- a user whose first translation was made in the 24 hours leading to the event start time or later.

However, we realised that there could be users who have made random translations weeks or even months before the event because they were trying out the translation system., this category of users shouldn't be discarded and they also need to be captured as new translation contributors.

trymebytes:
Fixes #213 
In this PR, first time contributors will now be known as "new translation contributors". We have also updated the logic for determining who a new translation contributor is by checking if they have submitted only 10 or fewer translations before the start time of the event.

**Test 1**
- Create a new user and join an event, the tada emoji should show beside their name on the attendee list on the event page.

**Test 2**
- Create a new user
- Create event and set the start time to a few minutes or hours later
- Create 11 or more translations before the start time of that event, then join the event.
- The tada emoji should not show beside their name on the attendee list on the event page.

### DIFF
--- a/includes/stats/stats-calculator.php
+++ b/includes/stats/stats-calculator.php
@@ -311,15 +311,16 @@ class Stats_Calculator {
 	 */
 	public function is_first_time_contributor( $event_start, $user_id ) {
 		global $wpdb, $gp_table_prefix;
+		$new_contributor_max_translation_count = 10;
 
 		// phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
 		// phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery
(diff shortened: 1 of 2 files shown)


akirk (changes requested):
I think we should not replace the criterium of the first translation comparison because they would lose their status as soon as they have translated more than 10 translations in the course of a translation event.

trymebytes:
We can check if they have at most 10 translations by the start of the event

psrpinto (approved):
LGTM 👍

akirk (approved):

PR #215, opened 2024-04-16
Event capabilities

psrpinto:
This PR introduces the following custom capabilities, and changes the whole codebase so that these capabilities are used, through calls to `current_user_can()`:

- `create_translation_event`
- `view_translation_event`
- `edit_translation_event`
- `delete_translation_event`

There should be no changes to how things work, everything should work the same as before.

## References

- [Meta capabilities for custom post types](https://justintadlock.com/archives/2010/07/10/meta-capabilities-for-custom-post-types)
- [`current_user_can()`](https://developer.wordpress.org/reference/functions/current_user_can/)
- [`map_meta_cap()`](https://developer.wordpress.org/reference/hooks/map_meta_cap/)

> This function also accepts an ID of an object to check against if the capability is a meta capability. Meta capabilities such as edit_post and edit_user are capabilities used by the map_meta_cap() function to map to primitive capabilities that a user or role has, such as edit_posts and edit_others_posts.

- [`friends/includes/class-access-control.php#L52`](https://github.com/akirk/friends/blob/5a9ee0d07c021ef1bd022790e4ce9cfb4b34d161/includes/class-access-control.php#L52)
- [`friends/includes/class-access-control.php#L302`](https://github.com/akirk/friends/blob/5a9ee0d07c021ef1bd022790e4ce9cfb4b34d161/includes/class-access-control.php#L302)
- [`global.wordpress.org/public_html/wp-content/mu-plugins/roles/cross-locale-pte.php#L386`](https://github.com/WordPress/wordpress.org/blame/trunk/global.wordpress.org/public_html/wp-content/mu-plugins/roles/cross-locale-pte.php#L386)

### DIFF
--- a/includes/event/event-capabilities.php
+++ b/includes/event/event-capabilities.php
@@ -130,7 +130,7 @@ class Event_Capabilities {
 			return false;
 		}
 
-		if ( user_can( $user->ID, 'manage_options' ) ) {
+		if ( user_can( $user->ID, 'delete_post', $event->id() ) ) {
 			return true;
 		}
 
--- a/includes/event/event-capabilities.php
+++ b/includes/event/event-capabilities.php
@@ -4,6 +4,8 @@ namespace Wporg\TranslationEvents\Event;
 
 use GP;
 use WP_User;
+use Wporg\TranslationEvents\Attendee\Attendee_Repository;
+use Wporg\TranslationEvents\Stats\Stats_Calculator;
 
 class Event_Capabilities {
 	private const CREATE = 'create_translation_event';
@@ -18,9 +20,17 @@ class Event_Capabilities {
 	);
 
 	private Event_Repository_Interface $event_repository;
+	private Attendee_Repository $attendee_repository;
+	private Stats_Calculator $stats_calculator;
 
-	public function __construct( Event_Repository_Interface $event_repository ) {
-		$this->event_repository = $event_repository;
+	public function __construct(
+		Event_Repository_Interface $event_repository,
+		Attendee_Repository $attendee_repository,
+		Stats_Calculator $stats_calculator
+	) {
+		$this->event_repository    = $event_repository;
+		$this->attendee_repository = $attendee_repository;
+		$this->stats_calculator    = $stats_calculator;
 	}
 
 	/**
--- a/includes/event/event-capabilities.php
+++ b/includes/event/event-capabilities.php
@@ -10,6 +10,7 @@ use Wporg\TranslationEvents\Stats\Stats_Calculator;
 
 class Event_Capabilities {
 	private const CREATE = 'create_translation_event';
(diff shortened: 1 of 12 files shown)


akirk (approved):
Great job, this vastly improves the readability and ensures consistency. Also thanks for the tests!

akirk (changes requested):
I have found one thing that no longer works: When an event doesn't have a host, the creator should be able to edit.

includes/event/event-capabilities.php
```diff
@@ -0,0 +1,186 @@
...
+					return false;
+				}
+				$event = $this->event_repository->get_event( $args[2] );
+				if ( ! $event ) {
+					return false;
+				}
+
+				if ( self::VIEW === $cap ) {
+					return $this->has_view( $user, $event );
+				} elseif ( self::EDIT === $cap ) {
```
akirk: Since we're returning, we can remove the `else`.
```suggestion
				}
				if ( self::EDIT === $cap ) {
```

includes/event/event-capabilities.php
```diff
@@ -0,0 +1,186 @@
...
+	 * @return bool
+	 */
+	private function has_edit( WP_User $user, Event $event ): bool {
+		if ( $event->end()->is_in_the_past() ) {
+			return false;
+		}
+
+		if ( $this->stats_calculator->event_has_stats( $event->id() ) ) {
+			return false;
+		}
```
akirk, line 118: Maybe this is too early and/or we need more fine grained controls (which is easy now!) But I think it should be possible to change at least the event description (and maybe the event end, if it should be extended) even when there are stats.

akirk (approved):
I'm going to re-approve this. I think this might not change the current behavior but we might need to discuss it wrt when editing an event should remain possible.

includes/event/event-capabilities.php
psrpinto: Indeed, more fine grained capabilities could make sense, e.g. `edit_translation_event_with_stats`. Also agree that it it should be possible to edit certain fields of a past event. I opened https://github.com/WordPress/wporg-gp-translation-events/issues/227 to track that.

psrpinto:
> I have found one thing that no longer works: When an event doesn't have a host, the creator should be able to edit.

I'm working on a fix for this

akirk:
@psrpinto I'm no longer sure this is true. I think I had tested with an event that already had contributions.

psrpinto:
There was indeed an issue when the event_id is a `string` and not an `int`. Now fixed.

psrpinto:
Merging since @akirk had approved, and after his approval there was only the bug fix in the last commit.

akirk:
Sorry for missing the re-review, thanks!

psrpinto:
No worries!

PR #228, opened 2024-04-19
Consolidate URL generation

psrpinto:
This PR makes it so that all URLs of the plugin are generated through a new `Urls` class, so whenever there's a need to generate a URL you can simply do something like:

```
echo Urls::event_details( $event_id );
```

This PR also fixes an issue where the URL was wrong in the event edit page (it would be correct when post is draft, and incorrect when post is published).

### DIFF
--- a/includes/routes/event/edit.php
+++ b/includes/routes/event/edit.php
@@ -8,6 +8,7 @@ use Wporg\TranslationEvents\Event\Event_Repository_Interface;
 use Wporg\TranslationEvents\Routes\Route;
 use Wporg\TranslationEvents\Stats\Stats_Calculator;
 use Wporg\TranslationEvents\Translation_Events;
+use Wporg\TranslationEvents\Urls;
 
 /**
  * Displays the event edit page.
@@ -39,20 +40,18 @@ class Edit_Route extends Route {
 		}
 
 		include ABSPATH . 'wp-admin/includes/post.php';
-		$event_page_title              = 'Edit Event';
-		$event_form_name               = 'edit_event';
-		$css_show_url                  = '';
-		$event_title                   = $event->title();
-		$event_description             = $event->description();
-		$event_status                  = $event->status();
-		list( $permalink, $post_name ) = get_sample_permalink( $event->id() );
-		$permalink                     = str_replace( '%pagename%', $post_name, $permalink );
-		$event_url                     = get_site_url() . gp_url( wp_make_link_relative( $permalink ) );
-		$event_timezone                = $event->timezone();
-		$event_start                   = $event->start();
-		$event_end                     = $event->end();
-		$create_delete_button          = false;
-		$visibility_delete_button      = 'inline-flex';
+		$event_page_title         = 'Edit Event';
+		$event_form_name          = 'edit_event';
+		$css_show_url             = '';
+		$event_title              = $event->title();
+		$event_description        = $event->description();
+		$event_status             = $event->status();
+		$event_url                = Urls::event_details_absolute( $event_id );
+		$event_timezone           = $event->timezone();
+		$event_start              = $event->start();
+		$event_end                = $event->end();
+		$create_delete_button     = false;
+		$visibility_delete_button = 'inline-flex';
 
 		if ( $event->end()->is_in_the_past() ) {
 			$this->die_with_error( esc_html__( 'You cannot edit a past event.', 'gp-translation-events' ), 403 );
--- a/tests/urls.php
+++ b/tests/urls.php
@@ -17,14 +17,10 @@ class Urls_Test extends GP_UnitTestCase {
 		$this->event_factory    = new Event_Factory();
 		$this->event_repository = new Event_Repository( new Attendee_Repository() );
 		$this->set_normal_user_as_current();
-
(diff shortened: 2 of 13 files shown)


akirk (changes requested):
Nice! Could you also find a way to add unit tests that use `define( 'GP_URL_BASE', '/' );` which we have on translate.wordpress.org?

Probably the `Urls` class fits our current plugin design but I was wondering if another option might have been to introduce a `gp_url_event()` similar to [`gp_url_project()` in GlotPress](https://github.com/GlotPress/GlotPress/blob/10644fd9bebd966c1327ff7fea3879556be7833b/gp-includes/url.php#L143).

includes/routes/event/edit.php
```diff
@@ -39,20 +40,18 @@ public function handle( int $event_id ): void {
...
-		$event_end                     = $event->end();
-		$create_delete_button          = false;
-		$visibility_delete_button      = 'inline-flex';
+		$event_page_title         = 'Edit Event';
+		$event_form_name          = 'edit_event';
+		$css_show_url             = '';
+		$event_title              = $event->title();
+		$event_description        = $event->description();
+		$event_status             = $event->status();
+		$event_url                = Urls::event_details_absolute( $event_id );
```
akirk, line 44: What do you think about introducing an alias `$event->url()`?

tests/urls.php
```diff
@@ -17,58 +17,62 @@ public function setUp(): void {
 		$this->event_factory    = new Event_Factory();
 		$this->event_repository = new Event_Repository( new Attendee_Repository() );
 		$this->set_normal_user_as_current();
+
+		if ( ! defined( 'GP_URL_BASE' ) ) {
+			define( 'GP_URL_BASE', '/' );
+		}
```
akirk: Oh, I wanted to suggest if we could test both, one with a GP_URL_BASE defined, and one without. So that it shows that it properly generates URLs for both cases.

akirk: AFAIK you can just define it on a per test basis, even though it's a constant; the next test will run with the constant undefined again.

includes/routes/event/edit.php
psrpinto: I'm not sure about this one because there's `Urls::event_details_absolute()` and `Urls::event_details()`. So we'd need to choose one to use in the `$event->url()` method.

I think then it could become confusing because you'd need to remember whether `$event->url()` returns the absolute or relative URL.

psrpinto: Or maybe all functions in `Urls` should return the absolute URL?

tests/urls.php
psrpinto: Gotcha, fixed.

psrpinto: > AFAIK you can just define it on a per test basis, even though it's a constant; the next test will run with the constant undefined again.

That's what I attempted to do indeed, but it's not the case, sadly. Because it's a PHP constant, once it's defined it can't be modified or unset. I also tried adding it to a different file, but that didn't work either.

I think since all tests are run in a single php process, once that constant is defined, it will keep the value until the process ends.

psrpinto:
> Probably the Urls class fits our current plugin design but I was wondering if another option might have been to introduce a gp_url_event() similar to [gp_url_project() in GlotPress](https://github.com/GlotPress/GlotPress/blob/10644fd9bebd966c1327ff7fea3879556be7833b/gp-includes/url.php#L143).

I think in practice the two approaches are similar, it's just that GlotPress defines those functions globally, and namespaces by prefixing the functions with `gp_url`, while here we use a namespaced Class. 

```php
# GlotPress/gp-includes/url.php

function gp_url_project( $project_or_path = '', $path = '', $query = null ) {}
```

```php
# wporg-gp-translation-events/includes/urls.php

namespace Wporg\TranslationEvents;

class Urls {
    public static function event_details( int $event_id ): string {}
}
```

Personally I like the PHP namespaces approach better, but I can change it to `gp_url_*()` methods if you prefer. Alternatively, we could also drop the `Urls` class and define the functions directly on the namespace:

```php
# wporg-gp-translation-events/includes/urls.php

namespace Wporg\TranslationEvents;

function event_details_url( int $event_id ): string {}
```
```php
# somefile.php

use function Wporg\TranslationEvents\event_details_url;

echo event_details_url( $event_id );
```

Let me know which approach you prefer.

includes/routes/event/edit.php
akirk: It's certainly not bad to always return absolute URLs, thus probably preferrable indeed, now that we have it unified.

akirk: I think just in this context it's preferrable to avoid id juggling but maybe it's one-off enough to be ok? Don't have a strong opinion.

akirk:
As I said earlier, the `Urls` approach probably fits better here after all. I just wanted to offer (or at least surface) the view to the other alternatives that are already in use elsewhere in GlotPress.

akirk (approved):
Nice, thanks! I think it's ok to address my feedback in an follow-up PR if we want to do them, I'm happy to get this over the finish line already in this state.

psrpinto:
Cool, merging!

PR #231, opened 2024-04-22
Use event repository when showing active events in translation table

psrpinto:
There was one place where we were not using the event repository: to show the "Contributing to X events" in the translation table:

Screenshot 2024-04-22 at 15 18 58

---

We're now also using the `Urls` class to generate the URLs for the links.


akirk (approved):
Tested through #235!

PR #233, opened 2024-04-25
Manage Attendees

### LINKED ISSUE #101: Manage attendee list on a dedicated view
Create a new view for hosts that shows the attendees in a table view with action buttons (e.g. Make host, remove) and sortable headers.

trymebytes:
Fixes #101 
**WIP**

**Screenshot**
image

image

### DIFF
--- a/templates/events-attendees.php
+++ b/templates/events-attendees.php
@@ -16,6 +16,7 @@ gp_tmpl_load( 'events-header', get_defined_vars(), __DIR__ );
 ?>
 <div class="event-page-wrapper">
 	<div class="event-details-stats">
+	<a href="<?php echo esc_url( Urls::event_details( $event->id() ) ); ?>" class="view-event-page">View Event</a>
 	<ul class="event-attendees-filter">
 		<li><a href="<?php echo esc_url( Urls::event_attendees( $event_slug ) ); ?>"><?php esc_html_e( 'All attendees', 'gp-translation-events' ); ?></a></a></li>
 		<li><a href="?filter=hosts"><?php esc_html_e( 'Hosts', 'gp-translation-events' ); ?></a></li>
--- a/templates/events-attendees.php
+++ b/templates/events-attendees.php
@@ -1,0 +1,16 @@
+<?php
+/**
+ * Events list page.
+ */
+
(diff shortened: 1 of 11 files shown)


templates/events-attendees.php
```diff
@@ -0,0 +1,63 @@
+<?php
+/**
+ * Events list page.
+ */
+
+namespace Wporg\TranslationEvents;
+
```
psrpinto, line 7:
```suggestion
use Wporg\TranslationEvents\Event\Event;

/** @var Event $event */
/** @var bool $is_active_filter */
```

templates/events-attendees.php
```diff
@@ -0,0 +1,63 @@
...
+gp_tmpl_load( 'events-header', get_defined_vars(), __DIR__ );
+
+
+
+?>
+<div class="event-page-wrapper">
+	<div class="event-details-stats">
+	<a href="<?php echo esc_url( Urls::event_details( $event->id() ) ); ?>" class="view-event-page">Go to event page</a>
+	<ul class="event-attendees-filter">
+		<li><a class="<?php echo ( ! $is_active_filter ) ? 'active-filter' : ''; ?>" href="<?php echo esc_url( Urls::event_attendees( $event_slug ) ); ?>"><?php esc_html_e( 'All attendees', 'gp-translation-events' ); ?></a></a></li>
```
psrpinto:
```suggestion
		<li><a class="<?php echo ( ! $is_active_filter ) ? 'active-filter' : ''; ?>" href="<?php echo esc_url( Urls::event_attendees( $event->slug() ) ); ?>"><?php esc_html_e( 'All attendees', 'gp-translation-events' ); ?></a></a></li>
```

psrpinto (approved):
Works great, well done!

trymebytes:
Thank you for catching these @psrpinto !

psrpinto:
Just to be clear, `$event_slug` would also work because the variable is defined on the route, but I find that it's more convenient/readable to use the getters on `$event` as then there's no need to declare `$event_slug` (or other variables) on the template. Just a detail :)

psrpinto (approved):

PR #235, opened 2024-04-29
Clean up stats calculator

psrpinto:
`Stats_Calculator` had accumulated some functions that aren't related to stats. This PR moves those functions to other places.

Everything works the same as it did before, but we're now doing fewer queries (see summary of changes below).

## Summary of changes

- Extract `get_projects()` out of `Stats_Calculator` into a new `Projects_Repository` (just a class that groups functions that retrieve projects)
- Extract `get_attendees_not_contributing()` out of `Stats_Calculator` into the `Attendee_Repository`
    - It also now returns instances of `Attendee`, to be consistent with other functions in that class.
        - This has the nice side effect that we're no longer doing n+1 queries to retrieve the attendees not contributing.


templates/event.php
```diff
@@ -88,24 +91,21 @@
...
-							<a href="<?php echo esc_url( get_author_posts_url( $_user->ID ) ); ?>" class="avatar"><?php echo get_avatar( $_user->ID, 48 ); ?></a>
-							<a href="<?php echo esc_url( get_author_posts_url( $_user->ID ) ); ?>" class="name"><?php echo esc_html( get_the_author_meta( 'display_name', $_user->ID ) ); ?></a>
-							<?php if ( $stats_calculator->is_new_translation_contributor( $event_start, $_user->ID ) ) : ?>
+							<a href="<?php echo esc_url( get_author_posts_url( $_attendee->user_id() ) ); ?>" class="avatar"><?php echo get_avatar( $_attendee->user_id(), 48 ); ?></a>
+							<a href="<?php echo esc_url( get_author_posts_url( $_attendee->user_id() ) ); ?>" class="name"><?php echo esc_html( get_the_author_meta( 'display_name', $_attendee->user_id() ) ); ?></a>
+							<?php if ( $stats_calculator->is_new_translation_contributor( $event_start, $_attendee->user_id() ) ) : ?>
 								<span class="first-time-contributor-tada" title="<?php esc_html_e( 'New Translation Contributor', 'gp-translation-events' ); ?>"></span>
 							<?php endif; ?>
 							<?php
-							$_attendee = $attendee_repo->get_attendee( $event_id, $_user->ID );
```
psrpinto, line 99: This is the query that was removed.

psrpinto: This is the query that was removed.

akirk (changes requested):
I noticed that a non-contributor is no longer detected as host:

no-host

Not sure why this breaks it but it works in `trunk`.

psrpinto:
I've found the cause of the issue above, working on a fix.

psrpinto:
Issue is now fixed.

akirk (approved):
Confirmed that this works now. Thanks!

psrpinto:
Thanks for testing @akirk !

PR #236, opened 2024-04-29
Improve event details page performance

### LINKED ISSUE #208: Improve Attendee/Hosts code
I think there's room for improvement on the Attendee/Host thing is currently represented in code. An [example](https://github.com/WordPress/wporg-gp-translation-events/pull/206#discussion_r1565866331) is:

Screenshot 2024-04-15 at 16 32 17

### LINKED ISSUE #225: Performance improvements for event details page
I noticed that the event details page takes a few seconds to load, for [Yoast Contributor day 2024 – April edition](https://translate.wordpress.org/events/yoast-contributor-day-2024-april-edition/).

We should look into what the culprits are and see if we can optimise it.

psrpinto:
Fixes #225 #208 

This PR reduces the time it takes to load the event details page. It does so by changing the logic so that all data is retrieved from the database, and then passed to the template, so that no queries are being done by the template. This reduces cases of needing to do n+1 queries, and makes it so that there's a single query that retrieves attendees/hosts/contributors.

Additionally, the `Attendee` model was improved so that an instance of `Attendee` now knows whether that attendee is a contributor or not (`$attendee->is_contributor()`). This simplifies things quite a bit IMHO.

Concerning finding who is a new contributor, that is now done in a single query for all contributors of an event, instead of doing one query per contributor. This was the biggest culprit of the performance issues, and where the most gains come from.

Note that the `Attendee` does **not** currently have an `$attendee->is_new_contributor()`. This is because retrieving this information in `get_attendees()` would make that method slow, and we don't necessarily need to know who is a new contributor in every place we call `get_attendees()`. When we do need that information, we can do something like [this](https://github.com/WordPress/wporg-gp-translation-events/pull/236/files#diff-191ce4bb541891d5409a539d712644c17ed150c3b7475e94444fb3a6c00eb2a3R80-R93).

## Metrics

Tested on my wporg sandbox, for the [Yoast Contributor Day](https://translate.wordpress.org/events/yoast-contributor-day-2024-april-edition/) event.

| Metric                       | Before | After |
|------------------------------|--------|-------|
| Page generation time         | 13.3s  | **4.3s**  |
| Query time                   | 10.4s  | **1.6s**  |
| Number of slow queries       | 10     | **3**     |
| Number of duplicated queries | 398    | **314**   |


includes/translation/translation-repository.php
```diff
@@ -0,0 +1,58 @@
...
+		global $wpdb, $gp_table_prefix;
+
+		// Prevent SQL injection.
+		foreach ( $user_ids as $user_id ) {
+			if ( ! is_int( $user_id ) ) {
+				return array();
+			}
+		}
+
+		$user_ids_param = implode( ',', $user_ids );
```
akirk: Another way how to do this with parameters is:
```suggestion
		$user_ids_param = implode( ',', array_fill( 0, count( $user_ids ), '%d' );
```

includes/translation/translation-repository.php
```diff
@@ -0,0 +1,58 @@
...
+	 * @param int[]             $user_ids Ids of users for which to count translations.
+	 * @param DateTimeImmutable $before   Only include translations before this datetime.
+	 *
+	 * @return array Associative array with user id as key and number of translations as value.
+	 */
+	public function count_translations_before( array $user_ids, DateTimeImmutable $before ): array {
+		global $wpdb, $gp_table_prefix;
+
+		// Prevent SQL injection.
+		foreach ( $user_ids as $user_id ) {
```
akirk, line 22: We also need to check whether `empty( $user_ids )` , otherwise the SQL statement below will be invalid.

includes/translation/translation-repository.php
```diff
@@ -0,0 +1,58 @@
...
+			group by user_id
+		";
+
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.NoCaching
+		// phpcs:disable WordPress.DB.PreparedSQL.NotPrepared
+		$rows = $wpdb->get_results(
+			$wpdb->prepare(
+				$query,
+				$before->format( 'Y-m-d H:i:s' ),
```
akirk: When using placeholders, see above:
```suggestion
				array_merge( $user_ids, $before->format( 'Y-m-d H:i:s' ) ),
```

includes/translation/translation-repository.php
```diff
@@ -0,0 +1,58 @@
...
+			),
+			OBJECT_K
+		);
+		// phpcs:enable
+
+		$results = array();
+		foreach ( $user_ids as $user_id ) {
+			$results[ $user_id ] = 0;
+		}
```
akirk: Could be written as:
```suggestion
		$results = array_fill_keys( $user_ids, 0 );
```

includes/attendee/attendee-repository.php
```diff
@@ -99,73 +106,76 @@ public function get_attendee( int $event_id, int $user_id ): ?Attendee {
...
 
-		$contributing_ids = $wpdb->get_col(
-			$wpdb->prepare(
-				"
-				select distinct user_id
-				from {$gp_table_prefix}event_actions
-				where event_id = %d
-			",
-				array(
+		return array_map(
```
akirk, line 152: Shouldn't we add at least in-request caching here? If I call something that invokes `get_attendees()` multiple times, I query the database multiple times.

includes/attendee/attendee.php
```diff
@@ -9,20 +9,26 @@ class Attendee {
...
+	/**
+	 * @var string[]
+	 */
+	private array $contributed_locales;
+
 	/**
 	 * @throws Exception
 	 */
-	public function __construct( int $event_id, int $user_id ) {
+	public function __construct( int $event_id, int $user_id, bool $is_host = false, array $contributed_locales = array() ) {
```
akirk, line 20: Nice! I was always a bit confused why we didn't support populating the full object.

includes/routes/event/details.php
```diff
@@ -45,18 +48,53 @@ public function handle( string $event_slug ): void {
...
-		$user_is_attending = $attendee instanceof Attendee;
-
 		$stats_calculator   = new Stats_Calculator();
 		$project_repository = new Project_Repository();
+
+		$current_user_attendee = $this->attendee_repository->get_attendee( $event->id(), $user->ID );
+		$user_is_attending     = $current_user_attendee instanceof Attendee;
+		$user_is_contributor   = $user_is_attending && $current_user_attendee->is_contributor();
+		$projects              = $project_repository->get_for_event( $event->id() );
+		$attendees             = $this->attendee_repository->get_attendees( $event->id() );
```
akirk:
```suggestion
		$projects              = $project_repository->get_for_event( $event->id() );
		$attendees             = $this->attendee_repository->get_attendees( $event->id() );
		$current_user_attendee = $attendees[ $user->ID ];
		$user_is_attending     = $current_user_attendee instanceof Attendee;
		$user_is_contributor   = $user_is_attending && $current_user_attendee->is_contributor();
```

includes/attendee/attendee-repository.php
```diff
@@ -99,73 +106,76 @@ public function get_attendee( int $event_id, int $user_id ): ?Attendee {
...
-				if ( '1' === $attendee_row->is_host ) {
-					$attendee->mark_as_host();
+	/**
+	 * Get attendees without contributions for an event.
+	 *
+	 * @return Attendee[]
+	 * @throws Exception
+	 */
+	public function get_attendees_not_contributing( int $event_id ): array {
+		return array_values(
```
akirk: If we opt to use user_ids as array keys (see above), we could forgo this.
```suggestion
		return 
```

includes/attendee/attendee-repository.php
```diff
@@ -99,73 +106,76 @@ public function get_attendee( int $event_id, int $user_id ): ?Attendee {
...
+						  and user_id = attendees.user_id
+					) as locales
+				from {$gp_table_prefix}event_attendees attendees
 				where event_id = %d
 			",
 				array(
 					$event_id,
 				)
-			),
+			)
```
akirk: If we add this, we can have the array user_id indexed (like you do in the `count_translations_before` function).
```suggestion
			), OBJECT_K
```

includes/attendee/attendee-repository.php
```diff
@@ -177,32 +187,14 @@ public function get_attendees_not_contributing( int $event_id ): array {
...
-			$wpdb->prepare(
-				"
-				select event_id, user_id
-				from {$gp_table_prefix}event_attendees
-				where event_id = %d and is_host = 1
-			",
-				array(
-					$event_id,
-				)
+		return array_values(
```
akirk:
```suggestion
		return
```

akirk (approved):
Generally this is great! I had a few suggestions, and one idea for a follow-up. I think we should consider dropping the `count_translations_before` function and move the `is_new_contributor` value to the `event_attendees` table. This would slightly change the definition of a new contributor, the "new contributor" status is determined when you join an event (vs. right now you could translate yourself out of being a new contributor between the attending click and the event start) but I think that's also valid and would remove these 1.2 seconds off the page load.

includes/translation/translation-repository.php
psrpinto: Did not know about `array_fill_keys()`! Nice

includes/attendee/attendee-repository.php
psrpinto: Agree, will be addressing it in an immediate follow-up.

psrpinto:
> I think we should consider dropping the count_translations_before function and move the is_new_contributor value to the event_attendees table.

This makes a lot of sense! Opened #241.

I've addressed all your remarks (great ones BTW!), could you take a quick look?

includes/routes/event/details.php
```diff
@@ -45,18 +52,50 @@ public function handle( string $event_slug ): void {
 		$event_start       = $event->start();
 		$event_end         = $event->end();
 
-		$attendee          = $this->attendee_repository->get_attendee( $event->id(), $user->ID );
-		$user_is_attending = $attendee instanceof Attendee;
+		$projects              = $this->project_repository->get_for_event( $event->id() );
+		$attendees             = $this->attendee_repository->get_attendees( $event->id() );
+		$current_user_attendee = $attendees[ $user->ID ];
```
akirk: I think we'll need an `isset()` check to avoid a warning here.

akirk (approved):
Looks good! Thanks for making the changes. I just think I found one thing that would generate a warning that we should avoid.

https://3v4l.org/HVTFE

trymebytes (commented):
Tested in my sandbox and pages take about 20s to load, not sure if this affects only me.

psrpinto:
@trymebytes I also had cases where the sandbox took longer to load the page. After a few refreshes it seemed to "pick up" and I got faster page loads.

akirk (approved):

PR #237, opened 2024-04-30
Add revisions for some postmeta keys

### LINKED ISSUE #204: Record Co-Host promotion/demotion in Revisions
Let's do some research how modifications to post_meta could be revealed in the revisions history.

An option would also be to just use [xwp/stream](https://github.com/xwp/stream).

amieiro:
With this PR, each time we update the post, it creates:
- A new post revision (this is the current behavior).
- 4 post meta (`_event_start`, `_event_end`, `_event_timezone`, `_hosts`) for each revision.
- The `_hosts` post meta key is a string with all the user ids that are hosts in this revision, separated with commas.

## `posts` table

image

## `postmeta` table

image

To test it, you can create and update an event, and change the hosts. Then you can review the updates in these tables: `posts` and `postmeta`.

Fixes https://github.com/WordPress/wporg-gp-translation-events/issues/204.

### DIFF
--- a/includes/event/event-repository.php
+++ b/includes/event/event-repository.php
@@ -413,8 +413,17 @@ class Event_Repository implements Event_Repository_Interface {
 	}
 
 	private function update_event_meta( Event $event ) {
+		$hosts     = $this->attendee_repository->get_hosts( $event->id() );
+		$hosts_ids = array_map(
+			function ( $host ) {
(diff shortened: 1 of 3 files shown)


psrpinto (approved):

trymebytes (approved):
Tested and works for me!

PR #240, opened 2024-05-02
Remove attendees

### LINKED ISSUE #63: Allow to remove somebody from an Event
For example when it's a local event and only people present in the room should be able to participate.

Also remove their contributions.

trymebytes:
Fixes #63  It adds a link to each row of attendee that allows event hosts or admins remove attendees from an event.

Screenshot 2024-05-02 at 13 38 51


akirk:
I took the liberty to do some style finetuning:
Screenshot 2024-05-07 at 08 01 18

akirk (dismissed):
When I try to remove myself, I get a whitepage with just the text "You do not have permission to remove this attendee."  This is not very nice UI. Why not hide the button?

akirk:
Tosin and I agreed to not show the button for hosts. If you'd want to remove a host, you need to remove them from being a host first.

wporg-gp-translation-events.php
```diff
@@ -106,6 +106,7 @@ public function gp_init() {
 		GP::$router->add( "/events/$slug/translations/$locale", array( 'Wporg\TranslationEvents\Routes\Event\Translations_Route', 'handle' ) );
 		GP::$router->add( "/events/$slug", array( 'Wporg\TranslationEvents\Routes\Event\Details_Route', 'handle' ) );
 		GP::$router->add( "/events/$slug/attendees", array( 'Wporg\TranslationEvents\Routes\Attendee\List_Route', 'handle' ) );
+		GP::$router->add( "/events/$slug/attendees/remove/$id", array( 'Wporg\TranslationEvents\Routes\Attendee\Remove_Attendee_Route', 'handle' ) );
```
psrpinto: The pattern I feel we have been following is to use the slug for pages (i.e. when the URL is visible to the user), and the event id for "api" calls, where the user does not see the URL. I think it would make sense to use the same pattern here.

includes/routes/attendee/remove-attendee.php
```diff
@@ -0,0 +1,68 @@
...
+		if ( ! current_user_can( 'edit_translation_event', $event->id() ) ) {
+			$this->die_with_error( esc_html__( 'You do not have permission to edit this event.', 'gp-translation-events' ), 403 );
+		}
+
+		$attendee = $this->attendee_repository->get_attendee( $event->id(), $user_id );
+		// The user is attending to the event, so if I don't find the attendee, I won't create it.
+		if ( $attendee instanceof Attendee ) {
+			if ( $attendee->is_host() || $attendee->user_id() === $user->ID ) {
+				$this->die_with_error( esc_html__( 'You do not have permission to remove this attendee.', 'gp-translation-events' ), 400 );
+			}
```
psrpinto: Should we use the new `edit_translation_event_attendees` capability here?

autoload.php
```diff
@@ -28,3 +28,4 @@
 require_once __DIR__ . '/includes/translation/translation-repository.php';
 require_once __DIR__ . '/includes/event-text-snippet.php';
 require_once __DIR__ . '/includes/routes/attendee/list.php';
+require_once __DIR__ . '/includes/routes/attendee/remove-attendee.php';
```
psrpinto: I think the pattern we have been following is to not suffix with the entity name. I think it would make sense to do the same here:

```suggestion
require_once __DIR__ . '/includes/routes/attendee/remove.php';
```

psrpinto (approved):
👍

PR #245, opened 2024-05-06
Trash and delete events

### LINKED ISSUE #230: Hide events
For administrators, it should be possible to change the post_status of an event so that unsuitable events can be marked as private or trashed.

On the events list page, there should be a button to hide an event quickly.

Create an admin view of hidden events, so that an event can be restored or fully deleted.

Maybe we can easily also add an edit link (maybe a pencil, and a trash can for deleting).

psrpinto:
Fixes #230 

This PR adds a "Trashed Events" page that shows all trashed events. Each displayed event has a "Restore" and a "Delete" button. Accessing this page requires the `manage_translation_events` capability, which is currently only assigned to GlopPress admins.

A new "Trashed Events" link has been added to the header, which is only visible for users with the `manage_translation_events` capability.

Permanently deleting an event also deletes attendees and stats, requires both the `manage_translation_events` and `delete_translation_event` capabilities.

## Summary of changes

- Rename "delete" to "trash" across the codebase
   - What we called "delete" was in fact "trashing"
- Add a `List_Trashed` route and template
- Add a `Trash` route, requires `trash_translation_event` capability
- Add a `Delete` route, requires `manage_translation_events` and `delete_translation_event` capabilities

## Next steps

- #252 

## Screen captures

Screenshot 2024-05-08 at 15 24 59


akirk (commented):
Everything works as you explained, it took me a bit to understand that in this PR you can only trash an event from the edit event page.

I am not super happy with using the word "trash" everywhere. For me, personally, this has a far more negative connotation than the word "delete". Would it work if we called "Trashed Events" just "Deleted Events" and say "Permanently Delete" ?

Screenshot 2024-05-08 at 16 03 44

Screenshot 2024-05-08 at 16 03 20

psrpinto:
>  it took me a bit to understand that in this PR you can only trash an event from the edit event page.

Indeed, it's a bit awkward now, I will address this in #252.

> Would it work if we called "Trashed Events" just "Deleted Events" and say "Permanently Delete" ?

Agree, I will make this change in this PR.

psrpinto:
I renamed "trash" to "delete" and "delete" to "delete permanently" in user-facing strings. I've update the screen capture in the PR description.

akirk (approved):
Thanks!

PR #246, opened 2024-05-06
New Event: Quickly Fill Upcoming WordCamp Data

akirk:
Loads the [next WordCamps](https://central.wordcamp.org/wp-json/wp/v2/wordcamps) and makes it easy to create an event from them:

quick-add

Along the way it fixes a bug where we try accessing the slug() on the not yet created event object.


psrpinto (approved):
Works very well, cool!

PR #248, opened 2024-05-06
Add Year to event URL

akirk:
By making the post type hierarchical, we can add the year to the slug.

WordCamp Europe in 2024 becomes https://translate.wordpress.org/events/2024/wordcamp-europe/ and in 2025 https://translate.wordpress.org/events/2025/wordcamp-europe/

add-year

### DIFF
--- a/includes/event/event-repository.php
+++ b/includes/event/event-repository.php
@@ -20,21 +20,26 @@ class Event_Repository implements Event_Repository_Interface {
 		$this->attendee_repository = $attendee_repository;
 	}
 
-	public function insert_event( Event $event ) {
-		$year        = $event->start()->utc()->format( 'Y' );
-		$parent_post = get_page_by_path( $year, OBJECT, self::POST_TYPE );
-		if ( ! $parent_post ) {
-			$parent_post = wp_insert_post(
+	private function get_year_post_id( $year ) {
+		$year_post = get_page_by_path( $year, OBJECT, self::POST_TYPE );
+		if ( ! $year_post ) {
+			return wp_insert_post(
 				array(
 					'post_type'    => self::POST_TYPE,
 					'post_title'   => $year,
+					'post_name'    => $year,
 					'post_status'  => 'publish',
 					'post_content' => '',
 				)
 			);
-			if ( $parent_post instanceof WP_Error ) {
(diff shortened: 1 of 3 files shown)


includes/event/event-repository.php
```diff
@@ -20,14 +20,36 @@ public function __construct( Attendee_Repository $attendee_repository ) {
 		$this->attendee_repository = $attendee_repository;
 	}
 
+	private function get_year_post_id( $year ) {
```
psrpinto:
```suggestion
        /*
         * @return int|WP_Error
         */
	private function get_year_post_id( string $year ) {
```

includes/event/event-repository.php
```diff
@@ -20,14 +20,36 @@ public function __construct( Attendee_Repository $attendee_repository ) {
...
+		}
+
 		$event_id_or_error = wp_insert_post(
 			array(
 				'post_type'    => self::POST_TYPE,
 				'post_name'    => $event->slug(),
 				'post_title'   => $event->title(),
 				'post_content' => $event->description(),
 				'post_status'  => $event->status(),
+				'post_parent'  => $post_parent,
```
psrpinto, line 57: I think it would make sense to modify `test_create_event()` to make sure the post parent is set correctly.

psrpinto (approved):
Would it be possible to achieve this without needing a parent post, just by prepending `$year/` to the slug? Not sure that's possible. 

I was thinking that we would need to alter the `get_*()` methods of the repository to not return parent events, but WordPress already seems to filter them out somehow. WordPress must be checking if a post has children, and not return them in that case?

includes/event/event-repository.php
akirk: Indeed, I'll add it!

akirk:
> Would it be possible to achieve this without needing a parent post, just by prepending $year/ to the slug? Not sure that's possible.

Unfortunately not, slashes are not allowed in slugs. I also considered adding the year with a dash (like `2024-wordcamp-europe`) but it's more hassle to try and bend the generated slug vs just adding a parent post.

Using the hierarchy allows us to keep everything else unchanged and yet benefit from the "namespace separation."

PR #249, opened 2024-05-06
Manage attendee cap

### LINKED ISSUE #242: Host cannot view attendees after first contribution
When an event has started and has at least a contribution, the edit event page where we have  the link to `Manage Attendees` is no longer accessible to the host. 

The manage events page checks for this permission `edit_translation_event` for the page to be accessible but since a contribution has been made, the attendees page is not also available.

Should the `edit_translation_event` permission be modified to allow for the attendees page to be accessible or should we have a new permission for this purpose?

trymebytes:
In this PR we add a new capability for managing event attendees, `manage_translation_event_attendees`. This fixes #242

### DIFF
--- a/includes/event/event-capabilities.php
+++ b/includes/event/event-capabilities.php
@@ -9,10 +9,11 @@ use Wporg\TranslationEvents\Attendee\Attendee_Repository;
 use Wporg\TranslationEvents\Stats\Stats_Calculator;
 
 class Event_Capabilities {
-	private const CREATE = 'create_translation_event';
-	private const VIEW   = 'view_translation_event';
-	private const EDIT   = 'edit_translation_event';
-	private const DELETE = 'delete_translation_event';
+	private const CREATE                 = 'create_translation_event';
+	private const VIEW                   = 'view_translation_event';
+	private const EDIT                   = 'edit_translation_event';
+	private const DELETE                 = 'delete_translation_event';
+	private const MANAGE_EVENT_ATTENDEES = 'manage_translation_event_attendees';
 
 	/**
 	 * All the capabilities that concern an Event.
@@ -22,6 +23,7 @@ class Event_Capabilities {
 		self::VIEW,
 		self::EDIT,
 		self::DELETE,
+		self::MANAGE_EVENT_ATTENDEES,
 	);
 
 	private Event_Repository_Interface $event_repository;
@@ -53,6 +55,7 @@ class Event_Capabilities {
 			case self::VIEW:
 			case self::EDIT:
 			case self::DELETE:
+			case self::MANAGE_EVENT_ATTENDEES:
 				if ( ! isset( $args[2] ) || ! is_numeric( $args[2] ) ) {
 					return false;
 				}
@@ -69,6 +72,9 @@ class Event_Capabilities {
 				}
 				if ( self::DELETE === $cap ) {
 					return $this->has_delete( $user, $event );
+				}
+				if ( self::MANAGE_EVENT_ATTENDEES === $cap ) {
+					return $this->has_manage_attendees( $user, $event );
 				}
 				break;
 		}
@@ -151,6 +157,30 @@ class Event_Capabilities {
 		}
 
 		if ( user_can( $user->ID, 'delete_post', $event->id() ) ) {
+			return true;
+		}
+
+		return false;
+	}
+
+	/**
+	 * Evaluate whether a user can manage attendees for a specific event.
+	 *
+	 * @param WP_User $user  User for which we're evaluating the capability.
+	 * @param Event   $event Event for which we're evaluating the capability.
+	 * @return bool
+	 */
+	private function has_manage_attendees( WP_User $user, Event $event ): bool {
+		if ( $this->is_gp_admin( $user ) ) {
+			return true;
+		}
+
+		if ( $event->author_id() === $user->ID ) {
+			return true;
+		}
+
+		$attendee = $this->attendee_repository->get_attendee( $event->id(), $user->ID );
+		if ( ( $attendee instanceof Attendee ) && $attendee->is_host() ) {
 			return true;
 		}
 
--- a/includes/event/event-capabilities.php
+++ b/includes/event/event-capabilities.php
@@ -74,7 +74,7 @@ class Event_Capabilities {
 					return $this->has_delete( $user, $event );
 				}
 				if ( self::EDIT_ATTENDEES === $cap ) {
-					return $this->has_manage_attendees( $user, $event );
+					return $this->has_edit_attendees( $user, $event );
 				}
 				break;
 		}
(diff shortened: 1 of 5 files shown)


templates/events-header.php
```diff
@@ -60,6 +60,8 @@
 			</span>
 			<?php if ( current_user_can( 'edit_translation_event', $event->id() ) ) : ?>
 				<a class="event-page-edit-link" href="<?php echo esc_url( Urls::event_edit( $event->id() ) ); ?>"><span class="dashicons dashicons-edit"></span><?php esc_html_e( 'Edit event', 'gp-translation-events' ); ?></a>
+			<?php elseif ( current_user_can( 'manage_translation_event_attendees', $event->id() ) ) : ?>
```
akirk: This also shows on the Manage Attendees page itself which is redundant:
Screenshot 2024-05-07 at 10 58 32

akirk (changes requested):
Works well, just the Manage Attendees link should be removed from the Manage Attendees page.

includes/event/event-capabilities.php
```diff
@@ -9,10 +9,11 @@
...
 class Event_Capabilities {
-	private const CREATE = 'create_translation_event';
-	private const VIEW   = 'view_translation_event';
-	private const EDIT   = 'edit_translation_event';
-	private const DELETE = 'delete_translation_event';
+	private const CREATE                 = 'create_translation_event';
+	private const VIEW                   = 'view_translation_event';
+	private const EDIT                   = 'edit_translation_event';
+	private const DELETE                 = 'delete_translation_event';
+	private const MANAGE_EVENT_ATTENDEES = 'manage_translation_event_attendees';
```
psrpinto: Could we call the capability `EDIT` instead of `MANAGE`? Manage makes me think of elevated permissions (e.g. super admin), but anyone who can edit the event should be able to edit the attendees. I see attendees as another property of the event (like title or start date), so to me changing attendees would be editing the event.

Additionally, because this is the `Event_Capabilities` class, we don't need the `_EVENT_`.

```suggestion
	private const EDIT_ATTENDEES = 'edit_translation_event_attendees';
```

trymebytes: I agree! Thank you!

includes/event/event-capabilities.php
```diff
@@ -70,6 +73,9 @@ private function has_cap( string $cap, array $args, WP_User $user ): bool {
 				if ( self::DELETE === $cap ) {
 					return $this->has_delete( $user, $event );
 				}
+				if ( self::EDIT_ATTENDEES === $cap ) {
+					return $this->has_manage_attendees( $user, $event );
```
psrpinto:
```suggestion
					return $this->has_edit_attendees( $user, $event );
```

includes/event/event-capabilities.php
```diff
@@ -157,6 +163,30 @@ private function has_delete( WP_User $user, Event $event ): bool {
 		return false;
 	}
 
+	/**
+	 * Evaluate whether a user can manage attendees for a specific event.
```
psrpinto:
```suggestion
	 * Evaluate whether a user can edit attendees for a specific event.
```

includes/event/event-capabilities.php
```diff
@@ -157,6 +163,30 @@ private function has_delete( WP_User $user, Event $event ): bool {
...
+	 *
+	 * @param WP_User $user  User for which we're evaluating the capability.
+	 * @param Event   $event Event for which we're evaluating the capability.
+	 * @return bool
+	 */
+	private function has_manage_attendees( WP_User $user, Event $event ): bool {
```
psrpinto:
```suggestion
	private function has_edit_attendees( WP_User $user, Event $event ): bool {
```

includes/routes/attendee/list.php
```diff
@@ -40,8 +40,8 @@ public function handle( string $event_slug ): void {
 		if ( ! $event ) {
 			$this->die_with_404();
 		}
-		if ( ! current_user_can( 'edit_translation_event', $event->id() ) ) {
-			$this->die_with_error( esc_html__( 'You do not have permission to edit this event.', 'gp-translation-events' ), 403 );
+		if ( ! current_user_can( 'edit_translation_event_attendees', $event->id() ) ) {
+			$this->die_with_error( esc_html__( 'You do not have permission to manage this event\'s attendees.', 'gp-translation-events' ), 403 );
```
psrpinto:
```suggestion
			$this->die_with_error( esc_html__( "You do not have permission to edit this event's attendees.", 'gp-translation-events' ), 403 );
```

psrpinto (approved):

templates/events-header.php
```diff
@@ -30,7 +30,7 @@
 			<?php endif; ?>
 		<?php endif; ?>
 	</ul>
-	<?php if ( isset( $event ) && ! isset( $event_form_name ) ) : ?>
+	<?php if ( isset( $event ) && ! isset( $event_form_name ) && ! isset( $hide_sub_head ) ) : ?>
```
trymebytes, line 33: In a follow-up PR, I want to introduce the `$hide_sub_head` flag to the forms. That way this line would become
```suggestion
	<?php if ( isset( $event ) && && ! isset( $hide_sub_head ) ) : ?>
```

psrpinto: Makes sense. Maybe the flag could be called `$show_sub_head`, to prevent a double negative, which I find harder to reason about.

psrpinto (approved):

templates/events-header.php
trymebytes: `Manage attendees` link removed.

image

templates/events-header.php
```diff
@@ -60,6 +60,8 @@
 			</span>
 			<?php if ( current_user_can( 'edit_translation_event', $event->id() ) ) : ?>
 				<a class="event-page-edit-link" href="<?php echo esc_url( Urls::event_edit( $event->id() ) ); ?>"><span class="dashicons dashicons-edit"></span><?php esc_html_e( 'Edit event', 'gp-translation-events' ); ?></a>
+			<?php elseif ( current_user_can( 'edit_translation_event_attendees', $event->id() ) ) : ?>
+				<a class="event-page-attendees-link" href="<?php echo esc_url( Urls::event_attendees( $event->slug() ) ); ?>"><?php esc_html_e( 'Manage attendees', 'gp-translation-events' ); ?></a>
```
akirk:
```suggestion
				<a class="event-page-attendees-link" href="<?php echo esc_url( Urls::event_attendees( $event->id() ) ); ?>"><?php esc_html_e( 'Manage attendees', 'gp-translation-events' ); ?></a>
```

akirk (changes requested):
Please rebase and change this line

templates/events-header.php
trymebytes: Done, thanks!

akirk (approved):

PR #253, opened 2024-05-08
Fix Manage Attendees Urls

akirk:
After #248, the Manage Attendees URL is broken because it directly uses the event slug. This makes it use the already existing `event_details()` method which fixes the links.


trymebytes (approved):
Tested and works well!

psrpinto:
Thanks for fixing!

PR #255, opened 2024-05-08
Fix old event routing

### LINKED ISSUE #254: Events on translate.wordpress.org returns 404
After introducing adding`year` slug to the event urls, all events created before this improvement now return 404.

eventspage

akirk:
Fixes #254.


akirk:
After we shipped this we can also update the slugs by adding the year page as a parent.

Screenshot 2024-05-08 at 14 41 32

trymebytes (approved):
Tested and works as expected, I didn't need to manually update the slug.

PR #256, opened 2024-05-08
Enable admin UI

### LINKED ISSUE #174: Allow managing events from wp-admin
Right now it cannot be edited because of errors with the event time.

akirk:
Fixes #174. If we want to add the year to old events, we need to add the right parent page (see #255). Let's enable the admin UI to achieve that. This PR fixes the event and timezone saving for that and adds a new meta box for hosts.

Screenshot 2024-05-14 at 15-05-44 Edit Translation Event „WordCamp Europe“ ‹ glotpress local — WordPress

### DIFF
--- a/wporg-gp-translation-events.php
+++ b/wporg-gp-translation-events.php
@@ -161,12 +161,47 @@ class Translation_Events {
 	 */
 	public function event_dates_meta_box( WP_Post $post ) {
 		wp_nonce_field( 'event_dates_nonce', 'event_dates_nonce' );
-		$event_start = get_post_meta( $post->ID, '_event_start', true );
-		$event_end   = get_post_meta( $post->ID, '_event_end', true );
-		echo '<label for="event_start">Start Date: </label>';
-		echo '<input type="date" id="event_start" name="event_start" value="' . esc_attr( $event_start ) . '" required>';
-		echo '<label for="event_end">End Date: </label>';
-		echo '<input type="date" id="event_end" name="event_end" value="' . esc_attr( $event_end ) . '" required>';
+		$event_start    = get_post_meta( $post->ID, '_event_start', true );
+		$event_end      = get_post_meta( $post->ID, '_event_end', true );
+		$event_timezone = get_post_meta( $post->ID, '_event_timezone', true );
+		$hosts          = explode( ',', get_post_meta( $post->ID, '_hosts', true ) );
+		if ( empty( $hosts ) ) {
+			$hosts = array( $post->post_author );
+		}
+		?>
+		<label for="event_start">Start Date: </label>
+		<input type="datetime-local" id="event_start" name="event_start" value="<?php echo esc_attr( $event_start ); ?>" required><br>
+		<label for="event_end">End Date: </label>
+		<input type="datetime-local" id="event_end" name="event_end" value="<?php echo esc_attr( $event_end ); ?>" required><br>
+		<label for="event-timezone">Timezone: </label>
+		<select id="event-timezone" name="event_timezone" required>
+			<?php
+			echo wp_kses(
+				wp_timezone_choice( $event_timezone, get_user_locale() ),
+				array(
+					'optgroup' => array( 'label' => array() ),
+					'option'   => array(
+						'value'    => array(),
+						'selected' => array(),
+					),
+				)
+			);
+			?>
+		</select><br>
+		<label for="event-hosts">Hosts: </label>
+		<?php
+		foreach ( $hosts as $host ) {
+			$user = get_user_by( 'ID', $host );
+			if ( $user ) {
+				?>
+				<a href="<?php echo esc_url( get_author_posts_url( $host ) ); ?>"><?php echo esc_html( $user->display_name ); ?></a>
+				<?php
+			} else {
+				?>
+				<i><?php echo esc_html( $host ); ?></i>
+				<?php
+			}
+		}
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


wporg-gp-translation-events.php
```diff
@@ -162,12 +161,47 @@ public function event_meta_boxes() {
...
-		$event_end   = get_post_meta( $post->ID, '_event_end', true );
-		echo '<label for="event_start">Start Date: </label>';
-		echo '<input type="date" id="event_start" name="event_start" value="' . esc_attr( $event_start ) . '" required>';
-		echo '<label for="event_end">End Date: </label>';
-		echo '<input type="date" id="event_end" name="event_end" value="' . esc_attr( $event_end ) . '" required>';
+		$event_start    = get_post_meta( $post->ID, '_event_start', true );
+		$event_end      = get_post_meta( $post->ID, '_event_end', true );
+		$event_timezone = get_post_meta( $post->ID, '_event_timezone', true );
+		$hosts          = explode( ',', get_post_meta( $post->ID, '_hosts', true ) );
+		if ( empty( $hosts ) ) {
```
amieiro: If `_hosts` is an empty value, the `$host` value will be `array(  0 => "")`, so we need to check the first value in the array: 

```suggestion
		if ( empty( $hosts[ 0 ] ) ) {
```

wporg-gp-translation-events.php
```diff
@@ -162,12 +161,47 @@ public function event_meta_boxes() {
...
+			$hosts = array( $post->post_author );
+		}
+		?>
+		<label for="event_start">Start Date: </label>
+		<input type="datetime-local" id="event_start" name="event_start" value="<?php echo esc_attr( $event_start ); ?>" required><br>
+		<label for="event_end">End Date: </label>
```
amieiro: As we are using UTC times, and we put a timezone selector, to avoid confusions, I suggest adding the `(UTC)` text.

```suggestion
		<label for="event_start">Start Date (UTC): </label>
		<input type="datetime-local" id="event_start" name="event_start" value="<?php echo esc_attr( $event_start ); ?>" required><br>
		<label for="event_end">End Date (UTC): </label>
```

amieiro:
Please, merge the last changes from `trunk`, because I was not able to access to the individual events without doing it.

akirk:
Sorry, I'm off today, feel free to take this over. Although it should work with just #255.

amieiro:
> Sorry, I'm off today, feel free to take this over. Although it should work with just #255.

Yes, it is working on production, so we can wait. Have a nice day!!!!

wporg-gp-translation-events.php
```diff
@@ -162,12 +162,58 @@ public function event_meta_boxes() {
...
 		wp_nonce_field( 'event_dates_nonce', 'event_dates_nonce' );
-		$event_start = get_post_meta( $post->ID, '_event_start', true );
-		$event_end   = get_post_meta( $post->ID, '_event_end', true );
-		echo '<label for="event_start">Start Date: </label>';
-		echo '<input type="date" id="event_start" name="event_start" value="' . esc_attr( $event_start ) . '" required>';
-		echo '<label for="event_end">End Date: </label>';
-		echo '<input type="date" id="event_end" name="event_end" value="' . esc_attr( $event_end ) . '" required>';
+		$event_start    = get_post_meta( $post->ID, '_event_start', true );
+		$event_end      = get_post_meta( $post->ID, '_event_end', true );
+		$event_timezone = get_post_meta( $post->ID, '_event_timezone', true );
```
psrpinto: I think it would make sense to use `self::get_event_repository()->get_event( $post->ID )` here, which will return an instance of `Event`. Then in the code below we can use the getters on the event instance to access the event properties, e.g. `$event->start()`.

wporg-gp-translation-events.php
```diff
@@ -162,12 +162,58 @@ public function event_meta_boxes() {
...
+		<?php
+	}
+
+	/**
+	 * Output the event dates meta box.
+	 *
+	 * @param  WP_Post $post The current post object.
+	 */
+	public function hosts_meta_box( WP_Post $post ) {
+		$hosts = explode( ',', get_post_meta( $post->ID, '_hosts', true ) );
```
psrpinto: Similarly to my other comment about using the event repository, it think it would make sense to use the `Attendee_Repository` to retrieve the list of hosts:

```php
$hosts = self::get_attendee_repository()->get_hosts( $post->ID )`;
```

wporg-gp-translation-events.php
akirk: The reason I went for using the post_meta directly is that I see this as a way to show the "real" database contents to the admin. Let's say we have a bug in our date functions, then here we can verify (and potentially fix) what's in the database since we show and save the data unmodified by our logic.

psrpinto: I see. In that case, since the purpose is to document for an admin how the data is stored, we could instead add that information as a comment. The purpose of the repositories was to abstract how the data is stored from the rest of the codebase, so it defeats the purpose a bit to not use them here. I understand your point, but personally I would optimise for making the code easier to maintain, instead of optimising for a situation that I think might not happen very often.

Just my perspective, I'm fine with leaving it as is. Though I think we shouldn't extend this practice to other places, if we need to do more things in wp-admin.

akirk: I think I tend to agree after all, so I have made your suggested changes. Thank you!

wporg-gp-translation-events.php
akirk: Implemented, thanks for the code snippet!

psrpinto (approved):
👍

Issue #247, opened 2024-05-06
Add edit and trash icons to events list


psrpinto:
- [ ] Add a pencil icon next to each entry on the events list page that the user can edit
- [ ] Add a trash icon next to each entry on the events list page that the user can trash

PR #257, opened 2024-05-08
Improve templates DevEx

psrpinto:
While dealing with templates, I've encountered some DevEx issues that I thought could be improved. This PR addresses such issues so that the templating system is easier to use and maintain. The issues I identified are:

1. Templates are called with `get_defined_vars()`, which makes it hard to track what data a template depends on, as it gets all the variables that are defined in the scope. For example, auxiliary variables that are defined only for the route logic, are also passed to the template, but the template does not use them.
2. There is no centralized place for the code necessary to set up the page header. This results in all templates needing to call a sequence of functions to set up the header.
3. The header template contains code that is specific to a given page. Ideally the header should only contain code that applies to all pages.
4. Instead of templates using the `$event` instance to access event information, additional variables are defined (e.g. `$event_title` and `$event_start`.

----

This PR addresses the above issues by doing the following.

When you need to render a template, you do something like:

```php
Templates::render(
    'my-template.php',
    array(
        'foo' => 'bar',
    ),
);
```

If rendering from a route, you should instead do:

```php
$this->render(
    'my-template.php',
    array(
        'foo' => 'bar',
    ),
);
```

There are convenience functions to render the header and the footer:

```php
# my-template.php

Templates::header(
	array(
		'html_title'  => esc_html__( 'My template', 'gp-translation-events' ),
		'page_title'  => esc_html__( 'My template', 'gp-translation-events' ),
		'breadcrumbs' => array( esc_html__( 'My template', 'gp-translation-events' ) ),
	),
);
?>

<p>hello</p>

<?php Templates::footer(); ?>
```

The `$page_title` can also be a function, which allows templates to add markup to the title:

```php

Templates::header(
	array(
		'page_title'  => function () use ( $event ) {
			echo esc_html( $event->title() );
			if ( 'draft' === $event->status() ) : ?>
				<span class="event-label-draft"><?php echo esc_html( $event->status() ); ?></span>
			<?php endif; ?>
			<?php
		},
	),
);
```

The header defines a "sub-header" section that can be "filled" by pages that need a sub-header:

```php
# my-template.php

Templates::header(
	array(
		'html_title' => esc_html__( 'My template', 'gp-translation-events' ),
		'page_title' => esc_html__( 'My template', 'gp-translation-events' ),
		'sub_head'   => function () {
			?>
				<p>This is the sub-header</p>
			<?php
		},
	),
);
?>
```


psrpinto:
I haven't touched the translations templates as they are more complicated to deal with, because the html is being rendered in the route. I plan on addressing those in a separate PR.

includes/routes/attendee/list.php
```diff
@@ -43,14 +43,20 @@ public function handle( string $event_slug ): void {
...
-		$attendees = array();
 		if ( gp_get( 'filter' ) && 'hosts' === gp_get( 'filter' ) ) {
 			$is_active_filter = true;
 			$attendees        = $this->attendee_repository->get_hosts( $event->id() );
 		} else {
 			$attendees = $this->attendee_repository->get_attendees( $event->id() );
 		}
 
-		$this->tmpl( 'events-attendees', get_defined_vars() );
+		$this->render(
```
psrpinto, line 53: If you prefer, I can change the code so that we call `$this->tmpl()` instead of `$this->render()`.

templates/event.php
```diff
@@ -9,39 +9,72 @@
...
+			?>
+			<span class="event-host">
+				<?php
+				if ( isset( $hosts ) ) :
+					if ( count( $hosts ) > 0 ) :
+						if ( 1 === count( $hosts ) ) :
+							esc_html_e( 'Host:', 'gp-translation-events' );
+						else :
+							esc_html_e( 'Hosts:', 'gp-translation-events' );
+						endif;
```
akirk, line 48: We should use a plural function here and insert the list of hosts as a variable.
```suggestion
						echo esc_html( _n( 'Host: %s' , 'Hosts: %s', count( $hosts ) 'gp-translation-events' ) );

```

templates/event.php
```diff
@@ -9,39 +9,72 @@
...
+					else :
+						esc_html_e( 'Created by:', 'gp-translation-events' );
```
akirk, line 50:
```suggestion
						esc_html_e( 'Created by: %s', 'gp-translation-events' );
```

templates/event.php
```diff
@@ -9,39 +9,72 @@
...
+						&nbsp;<a href="<?php echo esc_attr( get_author_posts_url( $event->author_id() ) ); ?>"><?php echo esc_html( get_the_author_meta( 'display_name', $event->author_id() ) ); ?></a>
+					<?php endif; ?>
+					<?php foreach ( $hosts as $host ) : ?>
+					&nbsp;<a href="<?php echo esc_attr( get_author_posts_url( $host->user_id() ) ); ?>"><?php echo esc_html( get_the_author_meta( 'display_name', $host->user_id() ) ); ?></a>
+						<?php if ( end( $hosts ) !== $host ) : ?>
+						,
+						<?php else : ?>
+						.
+						<?php endif; ?>
+					<?php endforeach; ?>
```
akirk, line 61: Can we do this using an `array_map()` and insert it above?

akirk (approved):
Thank you for this effort. I don't have a strong opinion about the use of `get_defined_vars()` but I am happy to support preference against it. I have added a few code remarks that we can also address in follow-ups so that we can unblock #262. I tested the flows and everything seemed fine in my testing.

Some higher-level thoughts:
- I like the idea that the header is its own template that has parameter callbacks. I am a bit worried if this really simplifies things. For example on the "Create Event" page I am not sure we want to show a "Create Event" button, how do we hide it just on that page?
- An idea could be to move to generating the header through options? We could try in a follow up to set header and footer options in the route before calling the `render()` function which would then automatically render header and footer before and after the template?
- In the upcoming redesign a lot of navigation should move to the footer, unifying this could be interesting.

psrpinto:
Great suggestions @akirk, I opened #271 and will address them once #247 is merged.

templates/event.php
psrpinto: After "Host:" (or "Hosts:") we display links to the profile of each host. Do you mean that we should pass that markup as the argument that will replace `%s`? I'm guessing that's not what you meant.

Screenshot 2024-05-13 at 15 22 19

templates/event.php
psrpinto: I'm not sure I understand, could you provide an example? Do you mean passing the separator as a field? Like:

```php
$hosts = array_map(
	function ( Attendee $host ) use ( $event, $hosts ) {
		return array(
			'url'       => get_author_posts_url( $event->author_id() ),
			'name'      => get_the_author_meta( 'display_name', $host->user_id() ),
			'separator' => end( $hosts ) === $host ? '.' : ', ',
		);
	},
	$hosts
);
?>
<?php foreach ( $hosts as $host ) : ?>
	<a href="<?php echo esc_attr( $host['url'] ); ?>">
		<?php echo esc_html( $host['name'] ); ?>
	</a><?php echo esc_html( $host['separator'] ); ?>
<?php endforeach; ?>
```

akirk: No, rather something like
```php
$hosts_list = array_map( function ( $host ) {
	return '<a href="' . esc_attr( get_author_posts_url( $host->user_id() ) ). '">' . esc_html( get_the_author_meta( 'display_name', $host->user_id() ) ). '</a>';
}, $hosts );
$hosts_list = implode( ', ', $hosts_list );
echo wp_kses(
	sprintf(
		/* translators: %s: Host(s) */
		_n( 'Host: %s', 'Hosts: %s', count( $hosts ), 'gp-translation-events' ),
		$hosts_list
	),
	array( 'a' => array( 'href' => array() ) )
);
```

templates/event.php
akirk: Yes, see below

PR #260, opened 2024-05-09
Add the events the user is or will attend

### LINKED ISSUE #152: Display the events the user has RSVPed for under My Events
Currently My Events displays:

- Events I host
- Events I have created 
- Events I attended 

I think it would make sense to also display the events the user is currently attending, as there's currently no way for the user to see their current and future events.

amieiro:
This PR adds:
- A list with the events the user is or will attend. It only shows the events that the user is attending or is going to attend that have not ended, sorted by the start date. 
- A top menu with links to the main sections.

This PR doesn't address the comments from the designers. I will create a new issue to do it.

image
 
Fixes #152.

### DIFF
--- a/includes/event/event-repository.php
+++ b/includes/event/event-repository.php
@@ -327,51 +327,6 @@ class Event_Repository implements Event_Repository_Interface {
 		// phpcs:enable
 	}
 
-	public function get_events_user_is_or_will_attend( int $user_id, int $page = -1, int $page_size = -1 ): Events_Query_Result {
-		global $wpdb, $gp_table_prefix;
-
-		// phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
-		// phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery
-		// phpcs:disable WordPress.DB.DirectDatabaseQuery.NoCaching
-		$events_user_is_or_will_attend_ids = $wpdb->get_col(
-			$wpdb->prepare(
-				"
-				select distinct event_id
-				from {$gp_table_prefix}event_attendees
-				where user_id = %d
-			",
-				array(
-					$user_id,
-				)
-			),
-		);
-		// phpcs:enable
-
-		if ( empty( $events_user_is_or_will_attend_ids ) ) {
-			return new Events_Query_Result( array(), 1, 1 );
-		}
-
-		$now = new DateTimeImmutable( 'now', new DateTimeZone( 'UTC' ) );
-
-		// phpcs:disable WordPress.DB.SlowDBQuery.slow_db_query_meta_query
-		// phpcs:disable WordPress.DB.SlowDBQuery.slow_db_query_meta_key
-		$query_args = array(
-			'post_status' => array( 'publish' ),
-			'meta_key'    => '_event_start',
-			'orderby'     => 'meta_value',
-			'order'       => 'ASC',
-			'meta_query'  => array(
-				array(
-					'key'     => '_event_end',
-					'value'   => $now->format( 'Y-m-d H:i:s' ),
-					'compare' => '>',
-					'type'    => 'DATETIME',
-				),
-			),
-		);
-		// phpcs:enable
-		return $this->execute_events_query( $page, $page_size, $query_args, $events_user_is_or_will_attend_ids );
-	}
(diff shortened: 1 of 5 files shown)


templates/events-my-events.php
```diff
@@ -18,10 +18,68 @@
 $event_page_title = __( 'My Events', 'gp-translation-events' );
 gp_tmpl_load( 'events-header', get_defined_vars(), __DIR__ );
 ?>
-
+<div class="events-links-to-anchors">
+	<ul>
+		<?php if ( ! empty( $events_i_am_or_will_attend_query->events ) ) : ?>
+			<li><a href="#events-i-am-or-will-attend"><?php esc_html_e( 'Events I am or will be attending', 'gp-translation-events' ); ?></a></li>
```
trymebytes, line 24:
```suggestion
			<li><a href="#events-i-am-or-will-attend"><?php esc_html_e( 'Events Joined', 'gp-translation-events' ); ?></a></li>
```

I would like to suggest that the header be changed to `Events joined` for sake of brevity. I understand that we will get the entire page redesigned soon , however for the list of events joined, we can have the title of past events in grey colour while the title of current or upcoming events stays the same. See example below,

image

amieiro: What I want to explain to the user with the title "Events I am or will be attending" is that in this list she can find only the events that she is attending now, or she will attend in the future. With the "Events Joined" I understand that it fits the past events, too.

includes/event/event-repository.php
```diff
@@ -327,6 +327,51 @@ public function get_events_created_by_user( int $user_id, int $page = -1, int $p
 		// phpcs:enable
 	}
 
+	public function get_events_user_is_or_will_attend( int $user_id, int $page = -1, int $page_size = -1 ): Events_Query_Result {
```
psrpinto: I think the existing `get_current_and_upcoming_events_for_user()` does the same as this method.

templates/events-my-events.php
```diff
@@ -18,10 +18,68 @@
...
+		<h2 id="events-i-am-or-will-attend"><?php esc_html_e( 'Events I am or will be attending', 'gp-translation-events' ); ?> </h2>
+		<ul>
+		<?php
+		foreach ( $events_i_am_or_will_attend_query->events as $event ) :
+			$stats_calculator = new Stats_Calculator();
+			$has_stats        = $stats_calculator->event_has_stats( $event->id() );
+			?>
+			<li class="event-list-item">
+				<a class="event-link-<?php echo esc_attr( $event->status() ); ?>" href="<?php echo esc_url( Urls::event_details( $event->id() ) ); ?>"><?php echo esc_html( $event->title() ); ?></a>
+				<?php if ( ! $event->end()->is_in_the_past() && ! $has_stats ) : ?>
```
psrpinto:
```suggestion
				<?php if ( current_user_can( 'edit_translation_event', $event->id() ) : ?>
```

amieiro: Good catch!!!! I have updated the logic here and in another 2 places!

psrpinto (approved):
👍

akirk (changes requested):
Let's hold off from merging just a bit. I think we should discuss the English wording.

akirk:
I've been thinking about this, and I think we're adding more complexity by adding one more section so that we even need a table of contents.

What if we just had two sections and used tags to explain the status of the event:
- My Upcoming Events (which can also include current events)
- My Past Events

Before:
Screenshot 2024-05-10 at 14 58 43

After (not fully accurate and just a mockup, but to convey the idea):
Screenshot 2024-05-10 at 15 04 57

In the next iteration we can then introduce filters like here in @jasmussen's design:
Screenshot 2024-05-10 at 15 05 49

psrpinto:
I like @akirk's suggestion of simplifying the page. Maybe we could even have a single list instead of splitting into "My next events" and "My past events", if we display "Attended" and "Attending" labels, that should be enough for the user to know the event is past.

amieiro:
I like the idea. As Paulo said, with a tag we can use only 1 list: 
- Will attend. For future events.
- Attending. For current events.
- Attended. For past events.

Is it ok if we merge this PR and I start working on the new approach in another PR?

akirk (approved):
Ok, let's merge this one and iterate as we agreed upon.

PR #261, opened 2024-05-09
Fix a bug showing the event creator when the event doesn't have any host

### LINKED ISSUE #259: Error showing the event creator
If an event doesn't have any hosts, it shows:
- If the user is logged, it says she is the creator. 
- If the user is logged out, it is empty.

image
image

The system should show the event creator in both situations.

amieiro:
If the event doesn't have any host, it should show the event creator. Currently, it shows the logged user of an empty string if the user is logged out.

This PR solves this problem, changing the method to get the ID of the author of the post.

To test it:
- Create an event.
- Remove all hosts.
- Open the event page with the event creator logged.
- Open the event page with a different user logged.
- Open the event page without any user logged.
- In these 3 situations, you should see the event creator after the "Created by: " text.

Fixes #259.

### DIFF
--- a/templates/events-header.php
+++ b/templates/events-header.php
@@ -45,7 +45,7 @@ use Wporg\TranslationEvents\Event\Event;
 					else :
 						esc_html_e( 'Created by:', 'gp-translation-events' );
 						?>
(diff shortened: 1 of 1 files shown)


psrpinto (approved):
Thanks for fixing!

trymebytes (approved):
Tested and works for me!

PR #262, opened 2024-05-09
Add buttons to event lists

### LINKED ISSUE #247: Add edit and trash icons to events list

psrpinto:
Fixes #247

This PR adds a "partial template" that renders a list of events, and uses that partial everywhere we need to render a list of events. This simplifies things quite a bit IMHO.

Additionally, "Edit", "Delete", "Restore" and "Delete permanently" buttons were added to each entry in the list, provided that the user has permissions to do those actions on a given event.

The buttons don't look great, but we can improve that later, if that's ok with you. Note that most users won't see the buttons, they will only see edit and delete for events they are hosts of.

## Testing instructions

This PR only changes the pages that render lists of events (listed below). Testing this PR means taking a look at those pages to make sure they render correctly.

- Home page
- My events
- Deleted events

## Screen captures

Screenshot 2024-05-14 at 14 56 21

Screenshot 2024-05-14 at 14 57 03
 

Screenshot 2024-05-10 at 15 33 33

### DIFF
--- a/assets/css/translation-events.css
+++ b/assets/css/translation-events.css
@@ -135,6 +135,7 @@ span.event-list-date.events-i-am-attending {
 	font-weight: normal;
 }
 
+/* Event title. */
 .event-list-item a:first-child {
 	font-weight: bold;
 	font-size: 1.2em;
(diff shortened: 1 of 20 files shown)


psrpinto:
The buttons don't look great, but we can improve upon that later, if that's ok with you. Also notice that most users won't see these buttons, they will only see edit and delete for the events they created or are hosts of.

trymebytes (commented):
I tested and noticed that the trash button is shown for events that are currently happening and have contributions. Also, when the trash button is clicked for this kind of events, I'm redirected to a page that has the text `You do not have permission to edit this event.` but the event has actuallly been deleted as it's no longer visible on the homepage or my events page.

psrpinto:
Great catches @trymebytes. I opened #274 and #275 to track those issues.

trymebytes (approved):
Tested and works for me!

PR #265, opened 2024-05-10
Generate unify generation of draft and published urls

akirk:
Previously, a draft post would get a temporary `Urls::event_details` URL like https://glotpress.local/?post_type=translation_event&p=570/ which also didn't work. This now also generates slug URLs for drafts.


amieiro (approved):
Tested and works fine.

PR #266, opened 2024-05-10
Check if the event has postmeta data

### LINKED ISSUE #264: Fatal error when a user tries to access to the year page
If a user tries to access to the year page, like https://glotpress.test/glotpress/events/2024/, she gets a fatal error.

amieiro:
If an event doesn't have postmeta data, it fires a fatal error.

This PR checks if the event has the regular postmeta data: `_event_start`, `_event_end` and `_event_end`. If the event doesn't have them, it returns a 404 error.

Fixes #264.

### DIFF
--- a/includes/event/event-repository.php
+++ b/includes/event/event-repository.php
@@ -131,6 +131,9 @@ class Event_Repository implements Event_Repository_Interface {
 
 		try {
 			$meta  = $this->get_event_meta( $id );
+			if ( ! $meta ) {
+				return null;
+			}
 			$event = new Event(
 				intval( $post->post_author ),
 				$meta['start'],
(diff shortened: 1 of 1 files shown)


akirk (approved):

includes/event/event-repository.php
```diff
@@ -501,10 +504,14 @@ private function get_event_post( int $event_id ): ?WP_Post {
 	/**
 	 * @throws Exception
 	 */
-	private function get_event_meta( int $event_id ): array {
+	private function get_event_meta( int $event_id ): ?array {
 		$meta = get_post_meta( $event_id );
 		$utc  = new DateTimeZone( 'UTC' );
 
+		if ( ! isset( $meta['_event_start'][0], $meta['_event_end'][0], $meta['_event_timezone'][0] ) ) {
```
akirk, line 511: Why the `[0]`?

akirk (commented):
I am not sure about a detail, let's discuss it before we merge.

includes/event/event-repository.php
amieiro: When we have data, this is an example of the `$meta` content:

```php
array(
  "_event_start" => array(
    0 => "2024-05-01 00:02:00"
  ),
  "_event_end" => array(
    0 => "2024-05-31 23:58:00"
  ),
  "_event_timezone" => array(
    0 => "Europe/Luxembourg"
  ),
  "_hosts" =>array(
    0 => ""
  ),
)
```
so we need to access to this `[0]` element.

akirk: Ok understood, thank you! Since we're not specifying the third parameter $single of get_post_meta it will always return an array.

PR #267, opened 2024-05-10
Show or hide subhead

trymebytes:
In this PR, we use a flag to hide or show the sub-head, the sub-head is a UI component that contains list of the hosts of an event , edit event and manage attendees links. Before this PR, we were hiding the sub heads only for the forms, we would check  if `$event_form_name` was not set and then we would hide the sub-head. However, there was a need to hide the sub-head for a page that isn't a form, hence the need for this approach.

Screenshot 2024-05-10 at 11 26 33


akirk:
Please add a description, I have made this a draft PR in the mean while.

psrpinto (approved):

PR #269, opened 2024-05-10
Add new capabilty to make a form field editable

### LINKED ISSUE #227: Make it possible to edit certain fields of a past event
Before the event has started, the hosts can edit:
- Title
- Description
- Dates

After the event has started but no contributions:
- Title
- Description
- Dates

After the event has started but with contributions:
- Title
- Description
- Extend the end date

1 hour after the event has ended:
- Title
- Description
- Extend the end date

After one hour after the event has ended, they can edit:
- Description

Reflect updates by a message "This event was last updated x minutes ago", linking to [a timeline](https://github.com/WordPress/wporg-gp-translation-events/issues/234).

trymebytes:
Fixes #227 

To test, kindly create the events that satisfies the conditions listed in the issue #227 

1. **Create an event that starts in the future, you would only be able to edit the following;**
Title
Description
Dates

 A. **Edit the event created in 1 to start now**

2. **With no contributions to the event created in A, you would only be able to edit the following;**
Title
Description
Dates

4. **Add contributions to the event created in A, you would only be able to edit the following;**
Title
Description
Extend the end date

B. **Find or create an event that has ended**

5. **Within 0-1 hour after the event has ended, you would only be able to edit the following;**
Title
Description
Extend the end date

6. **After one hour after the event has ended, you would only be able to edit the following;**
Description

### DIFF
--- a/includes/event/event-capabilities.php
+++ b/includes/event/event-capabilities.php
@@ -297,7 +297,6 @@ class Event_Capabilities {
 		}
 
 		return false;
-
 	}
 
 	public function register_hooks(): void {
--- a/includes/event/event-capabilities.php
+++ b/includes/event/event-capabilities.php
@@ -261,7 +261,7 @@ class Event_Capabilities {
 	 */
 	private function has_edit_form_field( WP_User $user, Event $event, $cap ): bool {
 		$event_start         = $event->start();
-		$happening_now       = $event->start()->is_in_the_past() && ! $event->end()->is_in_the_past();
+		$happening_now       = $event->is_active();
 		$now                 = new DateTimeImmutable( 'now', new DateTimeZone( 'UTC' ) );
 		$event_end_plus_1_hr = $event->end()->modify( '+1 hour' );
 
--- a/includes/event/event-capabilities.php
+++ b/includes/event/event-capabilities.php
@@ -105,7 +105,7 @@ class Event_Capabilities {
 					return $this->has_edit_attendees( $user, $event );
 				}
 				if ( self::EDIT_TITLE === $cap || self::EDIT_DESCRIPTION === $cap || self::EDIT_START === $cap || self::EDIT_END === $cap || self::EDIT_TIMEZONE === $cap ) {
-					return $this->has_edit_form_field( $user, $event );
+					return $this->has_edit_form_field( $user, $event, $cap );
 				}
 				break;
 		}
@@ -267,14 +267,35 @@ class Event_Capabilities {
 	 * @param Event   $event Event for which we're evaluating the capability.
 	 * @return bool
 	 */
-	private function has_edit_form_field( WP_User $user, Event $event ): bool {
-		$now          = new DateTimeImmutable( 'now', new DateTimeZone( 'UTC' ) );
-		$one_hour_ago = $now->modify( '-1 hour' );
+	private function has_edit_form_field( WP_User $user, Event $event, $cap ): bool {
+		$event_start         = $event->start();
+		$happening_now       = $event->start()->is_in_the_past() && ! $event->end()->is_in_the_past();
+		$now                 = new DateTimeImmutable( 'now', new DateTimeZone( 'UTC' ) );
+		$event_end_plus_1_hr = $event->end()->modify( '+1 hour' );
 
-		// Check if event ended at least 1 hour ago.
-		if ( $event->end()->is_in_the_past() && $event->end() < $one_hour_ago ) {
-			return false;
+		if ( self::EDIT_DESCRIPTION === $cap ) {
+			return true;
 		}
+
+		if ( $event_start > $now ) {
+			return true;
+		}
+
+		if ( $happening_now && ! $this->stats_calculator->event_has_stats( $event->id() ) ) {
+			return true;
+		}
+
+		if ( $happening_now && $this->stats_calculator->event_has_stats( $event->id() ) ) {
+			return ( self::EDIT_TITLE === $cap || self::EDIT_END === $cap );
+		}
+
+		if ( $event->end()->is_in_the_past() && $now < $event_end_plus_1_hr ) {
+			return ( self::EDIT_TITLE === $cap || self::EDIT_END === $cap );
+		}
+		if ( $event->end()->is_in_the_past() && $now > $event_end_plus_1_hr ) {
+			return ( self::EDIT_DESCRIPTION === $cap );
+		}
+
 		return false;
 
 	}
--- a/includes/event/event-capabilities.php
+++ b/includes/event/event-capabilities.php
@@ -260,7 +260,6 @@ class Event_Capabilities {
 	 * @return bool
 	 */
 	private function has_edit_field( WP_User $user, Event $event, $cap ): bool {
-		$event_start         = $event->start();
 		$happening_now       = $event->is_active();
 		$now                 = new DateTimeImmutable( 'now', new DateTimeZone( 'UTC' ) );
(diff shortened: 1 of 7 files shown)


includes/event/event-capabilities.php
```diff
@@ -17,6 +17,7 @@ class Event_Capabilities {
 	private const TRASH          = 'trash_translation_event';
 	private const DELETE         = 'delete_translation_event';
 	private const EDIT_ATTENDEES = 'edit_translation_event_attendees';
+	private const EDIT_FORM_FIELD = 'edit_translation_event_form_field';
```
akirk: We discussed adding new permissions per field, could you elaborate of the thinking to use `form_field` instead?
```suggestion
	private const EDIT_TITLE = 'edit_translation_event_title';
```

trymebytes: Yes, on another thought, I was trying to be mindful of creating a lot of permissions and also considering the need to update this list of permissions when we add a new field to the form. So, I wanted to have a single permission that checks for the permission for any form field in this manner
```
current_user_can( 'edit_translation_event_form_field', $event->id(), 'event_description' )
```
So to check for the permission for any field we only need to pass the field name as an argument as shown above.

psrpinto: I think in this case we could do something like:

```php
current_user_can( 'edit_translation_event', $event->id(), 'description' );
```

and modify the `edit_translation_event` capability so that it takes an optional field.

psrpinto: However, I think it might be more explicit to have `edit_translation_event_description` instead. In practice there's not much difference between the two options, I personally don't have a strong preference for either.

trymebytes: That's an interesting suggestion @psrpinto.

includes/event/event-capabilities.php
```diff
@@ -240,6 +252,45 @@ private function has_edit_attendees( WP_User $user, Event $event ): bool {
...
 	}
 
+	/**
+	 * Evaluate whether a user can edit event title for a specific event.
+	 *
+	 * @param WP_User $user  User for which we're evaluating the capability.
+	 * @param Event   $event Event for which we're evaluating the capability.
+	 * @return bool
+	 */
+	private function has_edit_form_field( WP_User $user, Event $event, $cap ): bool {
```
psrpinto: I think the edit field capabilities will be useful across the codebase (not just for the edit form), so I would drop the "form":

```suggestion
	private function has_edit_field( WP_User $user, Event $event, $cap ): bool {
```

includes/event/event-capabilities.php
```diff
@@ -240,6 +252,45 @@ private function has_edit_attendees( WP_User $user, Event $event ): bool {
...
+		$event_start         = $event->start();
+		$happening_now       = $event->start()->is_in_the_past() && ! $event->end()->is_in_the_past();
```
psrpinto:
```suggestion
		$happening_now       = $event->is_active();
```

includes/event/event-capabilities.php
```diff
@@ -240,6 +252,45 @@ private function has_edit_attendees( WP_User $user, Event $event ): bool {
...
 
+	/**
+	 * Evaluate whether a user can edit event title for a specific event.
+	 *
+	 * @param WP_User $user  User for which we're evaluating the capability.
+	 * @param Event   $event Event for which we're evaluating the capability.
+	 * @return bool
+	 */
+	private function has_edit_form_field( WP_User $user, Event $event, $cap ): bool {
+		$event_start         = $event->start();
```
psrpinto: We could drop this variable and call `$event->start()` directly whenever it's needed.

includes/event/event-capabilities.php
```diff
@@ -240,6 +252,45 @@ private function has_edit_attendees( WP_User $user, Event $event ): bool {
...
+		if ( $happening_now && ! $this->stats_calculator->event_has_stats( $event->id() ) ) {
+			return true;
+		}
+
+		if ( $happening_now && $this->stats_calculator->event_has_stats( $event->id() ) ) {
+			return ( self::EDIT_TITLE === $cap || self::EDIT_END === $cap );
+		}
+
+		if ( $event->end()->is_in_the_past() && $now < $event_end_plus_1_hr ) {
+			return ( self::EDIT_TITLE === $cap || self::EDIT_END === $cap );
```
psrpinto, line 283: I think if it's possible to edit the title, it should also be possible to edit the description, because the title is more restricted then description. Also applies to the condition above.

includes/event/event-capabilities.php
```diff
@@ -5,18 +5,25 @@
...
-	private const TRASH          = 'trash_translation_event';
-	private const DELETE         = 'delete_translation_event';
-	private const EDIT_ATTENDEES = 'edit_translation_event_attendees';
+	private const MANAGE           = 'manage_translation_events';
+	private const CREATE           = 'create_translation_event';
+	private const VIEW             = 'view_translation_event';
+	private const EDIT             = 'edit_translation_event';
+	private const TRASH            = 'trash_translation_event';
+	private const DELETE           = 'delete_translation_event';
+	private const EDIT_ATTENDEES   = 'edit_translation_event_attendees';
```
psrpinto, line 21: (Commenting here since I can't comment on `has_edit_attendees()` because it hasn't been modified)

I'm wondering, should we restrict editing attendees after the event has ended, similarly to what we're now doing for other fields? Currently, editing attendees is allowed after the event ends.

psrpinto (commented):
Looks great! Got a few comments

trymebytes:
Thank you for catching these, I'll address them.

includes/event/event-capabilities.php
trymebytes: Yes, that is correct and this [line](https://github.com/WordPress/wporg-gp-translation-events/blob/5bc84b7e9673f5ba1e658bc7032bfe3f2cc5db68/includes/event/event-capabilities.php#L266) covers for it.

includes/event/event-capabilities.php
trymebytes: I agree but we would need the attendees page to be accessible after the event ends. We would have to find a way to make the attendees page accessible if we add the logic to check for event end date in this capability.

templates/events-form.php
```diff
@@ -46,7 +46,7 @@
 	</div>
 	<div>
 		<label for="event-description"><?php esc_html_e( 'Event Description', 'gp-translation-events' ); ?></label>
-		<textarea id="event-description" name="event_description" rows="4" cols="40" required><?php echo esc_html( $event->description() ); ?></textarea>
+		<textarea id="event-description" name="event_description" rows="4" cols="40" required <?php echo esc_html( $is_create_form || current_user_can( 'edit_translation_event_description', $event->id() ) ?: 'readonly' ); ?> ><?php echo esc_html( $event->description() ); ?></textarea>
```
akirk: I think it'd be nice to avoid whitespace before a tag bracket:
```suggestion
		<textarea id="event-description" name="event_description" rows="4" cols="40" required <?php echo esc_html( $is_create_form || current_user_can( 'edit_translation_event_description', $event->id() ) ?: 'readonly' ); ?>><?php echo esc_html( $event->description() ); ?></textarea>
```

psrpinto (changes requested):
Code looks good 👍 

However, there's an issue when trying to update an event that is active now and has stats (step A.3 in the testing instructions), I get:

```
Error: Invalid time zone.
```

trymebytes:
Thanks for testing, taking another look.

psrpinto (approved):

PR #273, opened 2024-05-13
Remove some unwanted translations in the event stats

### LINKED ISSUE #258: Translation action captured twice in stats
It appears like we are counting the status change from `current` to `old` in the stats. To reproduce;
- Join an event as a GlotPress validator
- Go to a current GlotPress project and update the translation for a current translation and save ,  you would notice that the count for `created` in the stats table is increased by 2 instead of 1.
Screenshot 2024-05-08 at 15 50 58

amieiro:
The event page shows some incorrect stats in some situations: 
- When you replace one current translation, so you have one translation in `current` status and another in `old` status.
- When you reject one translation.
- When you have the same translated string in 2 different locales, they are counted in both locales.

This PR removes these problems. It only shows the translations in `current`, `fuzzy`, `waiting` and `changesrequested` status, avoiding the strings in `old` or `rejected` status.

To test it, try to reproduce the previous problems, see the stats, change to this branch and see the difference.

Let's see it with an example:

**Current situation**

image

**Proposed situation** 

image

Clicking on the links in the table, in the "Created" column, you can see the current translations (4) for Galician language and (1) for Spanish language.

image

image

I needed to update some tests, because now the tests need data in the `translation_set` and in the `translations` tables. 

Fixes #258.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,3 +42,9 @@ You can run tests in `wp-env` with the following command:
 ```shell
 composer dev:test
 ```
+
+If you want to run only one test, you can use the following command:
+
+```shell
+wp-env run tests-cli --env-cwd=wp-content/plugins/wporg-gp-translation-events ./vendor/bin/phpunit --filter methods_name
+```
--- a/README.md
+++ b/README.md
@@ -46,5 +46,5 @@ composer dev:test
 If you want to run only one test, you can use the following command:
 
 ```shell
-wp-env run tests-cli --env-cwd=wp-content/plugins/wporg-gp-translation-events ./vendor/bin/phpunit --filter methods_name
+composer dev:test -- --filter methods_name
 ```
--- a/tests/stats/stats-calculator.php
+++ b/tests/stats/stats-calculator.php
@@ -33,10 +33,10 @@ class Stats_Calculator_Test extends GP_UnitTestCase {
 		$this->set_normal_user_as_current();
 		$user_id = wp_get_current_user()->ID;
 
-		$event_id = $this->event_factory->create_active( array( $user_id ) );
-		$event    = get_post( $event_id );
-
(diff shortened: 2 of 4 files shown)


akirk (approved):
Thank you!

README.md
```diff
@@ -42,3 +42,9 @@ You can run tests in `wp-env` with the following command:
 ```shell
 composer dev:test
 ```
+
+If you want to run only one test, you can use the following command:
+
+```shell
+composer dev:test -- --filter methods_name
```
psrpinto, line 49: FYI, you can also do `--filter Class_Name` to run all the tests of a specific class/file.

tests/stats/stats-calculator.php
```diff
@@ -51,43 +51,62 @@ public function test_calculates_stats_for_event() {
...
+		$this->assertEquals( 2, $stats->rows()[ $translation_set_2->locale ]->reviewed );
+		$this->assertEquals( 3, $stats->rows()[ $translation_set_2->locale ]->users );
 
 		// Totals.
 		$this->assertEquals( 5, $stats->totals()->created );
 		$this->assertEquals( 5, $stats->totals()->reviewed );
 		$this->assertEquals( 3, $stats->totals()->users );
 	}
+
+	private function create_original_and_translation( $translation_set, $status = 'current' ) {
```
psrpinto, line 101: Instead of adding this method, we could add an optional `$translation_set` and `$status` argument to the existing `create()` function in `tests/lib/translation-factory.php`. Alternatively, we could move this function as is to `tests/lib/translation-factory.php`.

psrpinto (commented):
Thanks for fixing!

> When you have the same translated string in 2 different locales, they are counted in both locales.

Does this mean the string was translated twice (once in each locale) by the user? If so, should it not be counted once per locale?

PR #276, opened 2024-05-14
Add is_new_contributor to attendee

### LINKED ISSUE #241: Add is_new_contributor to attendees table
@akirk [said](https://github.com/WordPress/wporg-gp-translation-events/pull/236#pullrequestreview-2037997687):

> I think we should consider dropping the `count_translations_before` function and move the `is_new_contributor` value to the `event_attendees` table. This would slightly change the definition of a new contributor, the "new contributor" status is determined when you join an event (vs. right now you could translate yourself out of being a new contributor between the attending click and the event start) but I think that's also valid and would remove these 1.2 seconds off the page load.

psrpinto:
Fixes #241

This PR adds a new `is_new_contributor` column to the attendees table. The value of `is_new_contributor` is calculated when the user clicks the attend button. The `is_new_contributor` field is then used in the lists of attendees that are displayed in the event details page (contributors, attendees not contributing).

Additionally, the first time a request comes to wp-admin, `is_new_contributor` will be set for all contributors of existing events.

## Testing instructions

1. Log in to wp-admin, and create a user (`user2`)
3. Create a new event that starts before now, and ends some time in the future
4. Log in as `user2`
5. Add a translation
6. Navigate to the page of the event you created in 3. and click the attend button
7. Go to the page of the event and make sure `user2` is listed as a contributor, and has the 🎉 icon


tests/attendee/attendee-repository.php
```diff
@@ -239,9 +253,9 @@ public function test_is_new_translation_contributor() {
 		$this->repository->insert_attendee( $attendee12 );
 		$this->repository->insert_attendee( $attendee13 );
 
-		$this->assertTrue( $attendee11->is_new_contributor() );
-		$this->assertTrue( $attendee12->is_new_contributor() );
-		$this->assertFalse( $attendee13->is_new_contributor() );
+		$this->assertTrue( $attendee11->is_new_contributor_legacy() );
+		$this->assertTrue( $attendee12->is_new_contributor_legacy() );
+		$this->assertFalse( $attendee13->is_new_contributor_legacy() );
```
akirk: Is this just an intermediate step or are you planning to keep testing the legacy function? I'd argue in the above we should use the `Attendee_Adder` from #278 rather than directly inserting the attendee?

psrpinto: Yes indeed, it's an intermediate step, I renamed it so that tests would continue passing while I work on the PR. Everything will be moved to the `Attendee_Adder` by the end of this PR, like you suggested.

includes/upgrade.php
```diff
@@ -47,10 +61,77 @@ private static function get_database_schema_sql(): string {
...
+		foreach ( $events as $post ) {
+			$event = $event_repository->get_event( $post->ID );
+			if ( ! $event ) {
+				continue;
+			}
+
+			$attendees = $attendee_repository->get_attendees( $event->id() );
+
+			foreach ( $attendees as $attendee ) {
+				if ( ! $attendee->is_contributor() ) {
```
akirk: I think it should be possible that someone did not end up contributing but was a first contributor at the time

akirk (commented):
All worked well in my testing, just a remark about the transition.

includes/upgrade.php
psrpinto: Good point. I didn't consider that we show is_new_contributor in the list of non-contributing attendees. Fixed.

akirk (approved):

PR #277, opened 2024-05-14
Update the has_trash method

### LINKED ISSUE #274: Event delete capability allows deleting events that have stats
`has_trash()` should check if event has stats and refuse the capability, unless user has the `manage` capability.

amieiro:
This PR updates the `has_trash` method, checking if event has stats and refuse the capability, unless user has the `manage` capability.

Fixes https://github.com/WordPress/wporg-gp-translation-events/issues/274.

### DIFF
--- a/includes/event/event-capabilities.php
+++ b/includes/event/event-capabilities.php
@@ -174,16 +174,20 @@ class Event_Capabilities {
 	 * @throws Exception
 	 */
 	private function has_trash( WP_User $user, Event $event ): bool {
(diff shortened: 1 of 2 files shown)


psrpinto (approved):
👍

We could add a new test: `test_cannot_trash_event_with_stats()`. You could copy-paste the existing `test_cannot_edit_event_with_stats()` and alter it for trash.

trymebytes (approved):

PR #278, opened 2024-05-14
Fix issue that caused author stats to not be imported

### LINKED ISSUE #263: When a user creates an event, her stats are not added to the event
When a user creates an event, her stats are not added to the event. If she stops attending and then starts attending, the stats are added to the event.

psrpinto:
Fixes #263

This fixes the issue by adding an `Attendee_Adder` class that encapsulates everything that is needed to do when an attendee is added to the event. Currently it only imports stats, but in #276 it will also calculate the `is_new_contributor` flag for the attendee.


includes/routes/user/attend-event.php
```diff
@@ -54,13 +54,7 @@ public function handle( int $event_id ): void {
...
 		} else {
 			$attendee = new Attendee( $event->id(), $user_id );
-			$this->attendee_repository->insert_attendee( $attendee );
-
-			// If the event is active right now,
-			// import stats for translations the user created since the event started.
-			if ( $event->is_active() ) {
-				$this->stats_importer->import_for_user_and_event( $user_id, $event );
-			}
+			$this->attendee_adder->add_to_event( $event, $attendee );
```
akirk, line 57: I would love a function signature like this:
```suggestion
			$attendee = $event->add_user_as_attendee( $user_id );
```

akirk (commented):
I am a bit confused about introducing an `Attendee_Adder`. Couldn't we add this to the Event class like I suggested in the code snippet?

includes/routes/user/attend-event.php
psrpinto: > I am a bit confused about introducing an `Attendee_Adder`. Couldn't we add this to the Event class?

In the repository pattern that we're currently using, there are two types of classes:

1. Objects that encapsulate data (`Event`, `Attendee`, etc)
2. Services that retrieve those objects from the database, or write them to the database (`Event_Repository`, `Attendee_Repository`, etc)

In this pattern, the data objects (1. above) don't have logic, they are passed around the codebase, and the logic to write and read from the database is done in a repository.

However, repositories only deal with reading and writing data from/to the database, but often some logic needs to be done before the data is persisted. That's the case here, we want to import the stats, whenever an attendee is added to an event.

We don't want that logic to be in the repository however, because in some places of the codebase, we might want to use the repository just to write the data, without any side-effects.

So if we can't add logic to the data object (as per 1. above), and we can't add it to the repository, we need to add it somewhere else. Thus the `Attendee_Adder`.

----

I understand this pattern introduces some complexity, but there are advantages to using it.

For example, let's say that we drop the 1. "rule" from above, and logic is added to the `Event` class. In my experience, as more code/features are added, this will result in the `Event` class becoming a "god" class, that does everything related to an Event, which makes it harder to maintain.

If we do maintain rule 1., then we can add logic to different places, according to what the feature requires, which results in a more modular codebase, that is also easier to write tests for.

----

All this said, if I wasn't able to convince you :), I'm fine with removing rule 1. and add `add_user_as_attendee()` to `Event`.

akirk: Well, convinced to try your way, we can always refactor it :)

I think I am mostly coming from my own pattern, where I am not as strict about the boundaries and use more heavy objects that can do more than just store the respective data. Not recommending it, but like I described above, it can feel very natural.

I wonder if part of my irritation comes from the fact that `Attendee_Adder` seems too single purpose. Do you forsee this taking over functionality for other things (other than the calculation of the `is_new_attendee` that you mentioned in the PR description)?

psrpinto: I agree with what you say about `Attendee_Adder` being to single purpose here. I think we can move the [stats importing code](https://github.com/WordPress/wporg-gp-translation-events/blob/trunk/includes/stats/stats-importer.php) into the `Attendee_Adder`, and remove the `Stats_Importer` class. Because "stats importing" is only done when adding the attendee to the event, it's not used anywhere else.

In #276 I will also place the code for calculating `is_new_contributor` directly in `Attendee_Adder`.

akirk: Ok cool. Maybe we can do that when you add the is_new_contributor code and continue on this path to see how it holds up.

akirk (approved):

includes/routes/user/attend-event.php
psrpinto: I ended up removing the `Stats_Importer` and moving the code to `Attendee_Adder` here. I only saw your comment after I had already done it, let me know if you want me to drop the commit (it's the last commit in the PR).

PR #279, opened 2024-05-15
Update the redirection after trashing an event

### LINKED ISSUE #275: After trashing an event, user is redirected to the edit page of that event, but they might not be able to access it
To reproduce:

1. Create an event
2. Trash the event
3. You're redirected to the edit page but see an error message saying you cannot access that page

amieiro:
After trashing and untrashing an event, the user is redirected to the edit page for this event.

This PR changes this:
- Trashing: it redirects to the main page (events home).
- Untrashing: it redirects to the edit event page.

To test it:
- Create an event and publish it.
- Trash the event. You should be redirected to the main page.
- Go to the "Deleted Events" page. Untrash this event. You should be redirected to the edit page for this event.

Fixes #275.

### DIFF
--- a/includes/routes/event/trash.php
+++ b/includes/routes/event/trash.php
@@ -42,13 +42,14 @@ class Trash_Route extends Route {
 		if ( ! $event->is_trashed() ) {
 			// Trash.
 			$this->event_repository->trash_event( $event );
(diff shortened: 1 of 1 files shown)


psrpinto (approved):
👍

PR #280, opened 2024-05-16
Add social tags

amieiro:
This PR adds Twitter and Open Graph metadata, so when an event or is shared into a social network, the user can see a preview with information.

This is an example of the metadata added:

image

This PR adds:
- A new endpoint to get the event image. It returns a generic image if the event doesn't exist. This is useful for the other pages.
- The EBGaramond-Regular.ttf font, used in the WordPress.org titles, so we get the same typography, the same background color (`rgb(35, 40, 45)`) and font color (`rgb(255, 255, 255)`) that we have in the WordPress.org footer.
- The image can have 1 or 2 lines, limited to 22 chars each line, without breaking up the words, except at the end of the text.

Some examples of the event images:

One-line event:

image

Two-lines event:

image

Two-lines event with cut-out text:

image

Information about the Twitter and Open Graph metadata:
- [Overview of all Twitter Card Tags](https://developer.x.com/en/docs/twitter-for-websites/cards/overview/markup).
- [The Open Graph protocol](https://ogp.me/).

To test this PR in my local environment, I have used the [Social Share Preview Browser Extension](https://socialsharepreview.com/browser-extensions) to preview the metadata. You can see the metadata with the browser inspector.


templates/partials/header.php
```diff
@@ -6,9 +6,26 @@
...
+/** @var string $html_description */
 /** @var ?callable $sub_head */
 /** @var ?string[] $breadcrumbs */
+add_filter(
+	'gp_head',
+	function () use ( $html_title, $url, $html_description ) {
+		echo '<meta name="twitter:card" content="summary" />' . "\n";
+		echo '<meta name="twitter:site" content="@WordPress" />' . "\n";
+		echo '<meta name="twitter:title" content="' . esc_attr( $html_title ) . '" />' . "\n";
+		echo '<meta name="twitter:description" content="WordPress Translation Events" />' . "\n";
```
akirk: Forgotten to update?
```suggestion
		echo '<meta name="twitter:description" content="' . esc_html( $html_description ) . '" />' . "\n";
```

templates/partials/header.php
```diff
@@ -6,9 +6,26 @@
...
+		echo '<meta name="twitter:creator" content="@WordPress" />' . "\n";
+		echo '<meta name="twitter:image" content="https://live.staticflickr.com/2512/3967155269_f3246224c8_b.jpg" />' . "\n";
```
akirk: Could we omit the image as long as we don't have one? Ideally we'd generate the picture from the event stats, similar to https://jetpack.com/blog/jetpack-acquires-wordpress-plugin-social-image-generator/

templates/partials/header.php
```diff
@@ -6,9 +6,26 @@
...
+		echo '<meta name="twitter:card" content="summary" />' . "\n";
+		echo '<meta name="twitter:site" content="@WordPress" />' . "\n";
+		echo '<meta name="twitter:title" content="' . esc_attr( $html_title ) . '" />' . "\n";
+		echo '<meta name="twitter:description" content="' . esc_html( $html_description ) . '" />' . "\n";
+		echo '<meta name="twitter:creator" content="@WordPress" />' . "\n";
+		echo '<meta name="twitter:image" content="https://live.staticflickr.com/2512/3967155269_f3246224c8_b.jpg" />' . "\n";
 
+		echo '<meta property="og:url" content="' . esc_url( $url ) . '" />' . "\n";
+		echo '<meta property="og:title" content="' . esc_html( $html_title ) . '" />' . "\n";
+		echo '<meta property="og:description" content="' . esc_html( $html_description ) . '" />' . "\n";
```
akirk:
```suggestion
		echo '<meta property="og:title" content="' . esc_attr( $html_title ) . '" />' . "\n";
		echo '<meta property="og:description" content="' . esc_attr( $html_description ) . '" />' . "\n";
```

templates/partials/header.php
amieiro: I want to try to generate one image over the fly with the GD library, so we can have an image with the title. What do you think about?

akirk: I think it'd be great. How do you plan to do it? Introduce a new URL?

amieiro: Yes, adding a new endpoint for the images, something like 
`GP::$router->add( "/events/image/$id", array( 'Wporg\TranslationEvents\Routes\Event\Image_Route', 'handle' ) );`

akirk: That would be great! Looking forward to this! The advantage is that we can output the URL but won't be slowed down by image generation.

includes/routes/event/image.php
```diff
@@ -0,0 +1,112 @@
+<?php
+
+namespace Wporg\TranslationEvents\Routes\Event;
+
+use DateTimeImmutable;
+use DateTimeZone;
+use Wporg\TranslationEvents\Event\Event;
+use Wporg\TranslationEvents\Event\Event_End_Date;
+use Wporg\TranslationEvents\Event\Event_Start_Date;
```
psrpinto:
```suggestion
```

includes/routes/event/image.php
```diff
@@ -0,0 +1,112 @@
...
+
+	/**
+	 * Splits a string into two lines based on the maximum line length.
+	 *
+	 * @param string $text       The text to split.
+	 * @param int    $max_length The maximum length of each line.
+	 *
+	 * @return string[]
+	 */
+	private function splitText( string $text, int $max_length ): array {
```
psrpinto: I'm surprised phpcs doesn't catch this one

```suggestion
	private function split_text( string $text, int $max_length ): array {
```

includes/routes/event/image.php
```diff
@@ -0,0 +1,112 @@
...
+		if ( ! extension_loaded( 'gd' ) ) {
+			$this->die_with_error( esc_html__( 'The image cannot be generated because GD extension is not installed.', 'gp-translation-events' ) );
+		}
+
+		$event = $this->event_repository->get_event( $event_id );
+		$text  = ! $event ? esc_html__( 'Translation events', 'gp-translation-events' ) : $event->title();
+		$text  = '' === $text ? esc_html__( 'Translation events', 'gp-translation-events' ) : $text;
+		$text  = substr( $text, 0, 44 ); // Limit the text to 44 characters.
+
+		$lines = $this->splitText( $text, 22 ); // Limit each line to 22 characters.
```
psrpinto:
```suggestion
		$lines = $this->split_text( $text, 22 ); // Limit each line to 22 characters.
```

templates/partials/header.php
```diff
@@ -6,9 +6,40 @@
...
 /** @var string $html_title */
 /** @var string|callable $page_title */
+/** @var string $url */
+/** @var string $image_url */
+/** @var string $html_description */
 /** @var ?callable $sub_head */
 /** @var ?string[] $breadcrumbs */
 
+$html_title       = $html_title ?? esc_html__( 'Translation Events', 'gp-translation-events' );
+$url              = $url ?? URLS::events_home();
```
psrpinto:
```suggestion
$url              = $url ?? Urls::events_home();
```

tests/event/event-image.php
```diff
@@ -0,0 +1,53 @@
...
+
+class Event_Image_Test extends GP_UnitTestCase {
+	/**
+	 * Test that when the header template is fired, it generates the social metadata.
+	 *
+	 * @return void
+	 */
+	public function test_gp_head_metadata() {
+		$html_title       = 'Test HTML Title';
+		$url              = URLS::events_home();
```
psrpinto:
```suggestion
		$url              = Urls::events_home();
```

templates/partials/header.php
```diff
@@ -6,9 +6,40 @@
...
+$html_description = $html_description ?? esc_html__( 'WordPress Translation Events', 'gp-translation-events' );
+$image_url        = $image_url ?? URLs::event_default_image();
```
psrpinto:
```suggestion
$image_url        = $image_url ?? Urls::event_default_image();
```

tests/event/event-image.php
```diff
@@ -0,0 +1,53 @@
...
+		$html_description = 'This is a test description of the page.';
+		$image_url        = URLs::event_default_image();
```
psrpinto:
```suggestion
		$image_url        = Urls::event_default_image();
```

templates/partials/header.php
```diff
@@ -6,9 +6,40 @@
...
+
+add_action(
+	'gp_head',
+	function () use ( $html_title, $url, $html_description, $image_url ) {
+		echo '<meta name="twitter:card" content="summary" />' . "\n";
```
psrpinto, line 23: Instead of echos, we could do:

```php
function () use ( $html_title, $url, $html_description, $image_url ) {
?>
    <meta name="twitter:title" content="<?php echo esc_attr( $html_title ); ?>" />
<?php
```

psrpinto (approved):
Great work, the image looks nice!

Only found some cosmetic things.

templates/partials/header.php
amieiro: For me, it is more readable as it is, so I prefer to leave it as it is, but it is entirely up to opinion. Thanks for the feedback.

psrpinto: No problem, just a suggestion.

psrpinto (approved):

templates/partials/header.php
```diff
@@ -6,9 +6,40 @@
...
+		echo '<meta name="twitter:creator" content="@WordPress" />' . "\n";
+		echo '<meta name="twitter:image" content="' . esc_url( $image_url ) . '" />' . "\n";
+		echo '<meta name="twitter:image:alt" content="' . esc_attr( $html_title ) . '" />' . "\n";
+
+		echo '<meta property="og:url" content="' . esc_url( $url ) . '" />' . "\n";
+		echo '<meta property="og:title" content="' . esc_attr( $html_title ) . '" />' . "\n";
+		echo '<meta property="og:description" content="' . esc_attr( $html_description ) . '" />' . "\n";
+		echo '<meta property="og:site_name" content="' . esc_attr( get_bloginfo() ) . '" />' . "\n";
+		echo '<meta property="og:image:url" content="' . esc_url( $image_url ) . '" />' . "\n";
+		echo '<meta property="og:image:secure_url" content="' . esc_url( $image_url ) . '" />' . "\n";
```
akirk, line 36: I don't know if it's really necessary but I think we should show absolute URLs here.

amieiro: Done!

includes/urls.php
```diff
@@ -49,6 +49,26 @@ public static function event_create(): string {
...
 
+	/**
+	 * Returns the absolute URL to the image for the event.
+	 *
+	 * @param int $event_id The event ID.
+	 *
+	 * @return string
+	 */
+	public static function event_image( int $event_id ): string {
+		return trailingslashit( gp_url_public_root() ) . "events/image/$event_id";
```
akirk: I think we were using this convention:
```suggestion
		return home_url( gp_url( "events/image/$event_id" ) );
```

tests/event/event-image.php
```diff
@@ -0,0 +1,108 @@
...
+				'url'              => $url,
+				'html_description' => $html_description,
+				'image_url'        => $image_url,
+				'page_title'       => $page_title,
+			),
+		);
+
+		ob_start();
+		do_action( 'gp_head' );
+		$output = ob_get_clean();
```
akirk: Something is strange here. I noticed it because `Templates::header` outputs HTML already, so the phpunit output shows the HTML contents. But this also works, so do we even need to `do_action()` or are you testing something else here?

```suggestion
		ob_start();
		Templates::header(
			array(
				'html_title'       => $html_title,
				'url'              => $url,
				'html_description' => $html_description,
				'image_url'        => $image_url,
				'page_title'       => $page_title,
			),
		);
		$output = ob_get_clean();
```

akirk (commented):
Sorry another remark regarding the test that I saw now.

includes/urls.php
amieiro: I only see one occurrence, but it is ok for me to use it!

tests/event/event-image.php
amieiro: I used the `do_action` to fire the `add_action` in the template, but it is fired without this, so I removed it.

akirk (approved):
Thank you, this is really great!

akirk:
@amieiro I'm afraid the unit tests fail after being merged to trunk (probably a result of #294), as you can see in my new PR #302. Could you take a look?

amieiro:
> @amieiro I'm afraid the unit tests fail after being merged to trunk (probably a result of #294), as you can see in my new PR #302. Could you take a look?

Fixed!

akirk:
Thank you!

PR #282, opened 2024-05-17
Show a single event list in my events page

psrpinto:
Fixes #270

## Screen captures

Screenshot 2024-05-20 at 17 38 54

PR #283, opened 2024-05-20
Make it possible to query all events of a user

psrpinto:
In #282, I will need to query all events for a given user, but there was no method in the event repository to do so, so I'm doing that in this PR. Additionally, prior to this PR, whenever we needed to retrieve a list of events for a given user, we were doing the following:

1. Query for all the events a user is an attendee of
2. Pass the event ids returned in 1. as the `posts_in` condition for the `get_posts()`

This means that performance will degrade as time passes and more events are created, as 1. above returns all events the user is an attendee of.

This PR changes this, by making it possible to modify the where condition of the `get_posts()` to something like:

```php
$where and exists( select * from $attendees_table where user_id = $user_id and event_id = $posts_table.ID )
```

This makes it possible to query for events that the user is attending, so we no longer need step 1. above.

## Summary of changes

- `Event_Repository`:
    - Add `get_events_for_user()`
    - Change the internals so that we're more efficiently querying for the events of a user (as per description above)
- `Attendee_Repository`:
    - Rename some methods for clarity
    -  Add `get_attendees_for_events_for_user()`

## Testing instructions
There's no need for manual testing as the repositories are covered by tests.

### DIFF
--- a/includes/attendee/attendee-repository.php
+++ b/includes/attendee/attendee-repository.php
@@ -131,6 +131,72 @@ class Attendee_Repository {
 	}
 
 	/**
+	 * @var int[] $event_ids
+	 * @return Attendee[]
+	 * @throws Exception
+	 */
+	public function get_attendees_for_events_for_user( array $event_ids, int $user_id ): array {
+		if ( empty( $event_ids ) ) {
+			return array();
+		}
+
+		// Prevent SQL injection.
+		foreach ( $event_ids as $event_id ) {
+			if ( is_numeric( $event_id ) ) {
+				$event_id = intval( $event_id );
+			}
+			if ( ! is_int( $event_id ) || $event_id <= 0 ) {
+				return array();
+			}
+		}
+
+		global $wpdb, $gp_table_prefix;
+		$event_id_params = implode( ',', array_fill( 0, count( $event_ids ), '%d' ) );
+
+		// phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.NoCaching
+		$rows = $wpdb->get_results(
+			$wpdb->prepare(
+				"
+					select
+					    event_id,
+						user_id,
+						is_host,
+						is_new_contributor,
+						(
+							select group_concat( distinct locale )
+							from {$gp_table_prefix}event_actions
+							where event_id = attendees.event_id
+							  and user_id = attendees.user_id
+						) as locales
+					from {$gp_table_prefix}event_attendees attendees
+					where event_id in ($event_id_params)
+					  and user_id = %d
+				",
+				array_merge(
+					$event_ids,
+					array( $user_id ),
+				),
+			),
+		);
+		// phpcs:enable
+
+		return array_map(
+			function ( $row ) {
+				return new Attendee(
+					$row->event_id,
+					$row->user_id,
+					'1' === $row->is_host,
+					'1' === $row->is_new_contributor,
+					null === $row->locales ? array() : explode( ',', $row->locales ),
+				);
+			},
+			$rows,
+		);
+	}
+
+	/**
 	 * Retrieve all the attendees of an event.
 	 *
 	 * @return Attendee[] Attendees of the event. Associative array with user_id as key.
--- a/includes/attendee/attendee-repository.php
+++ b/includes/attendee/attendee-repository.php
@@ -87,47 +87,11 @@ class Attendee_Repository {
 	 * @throws Exception
 	 */
 	public function get_attendee_for_event_for_user( int $event_id, int $user_id ): ?Attendee {
-		global $wpdb, $gp_table_prefix;
-
-		// phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
-		// phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery
-		// phpcs:disable WordPress.DB.DirectDatabaseQuery.NoCaching
-		$row = $wpdb->get_row(
-			$wpdb->prepare(
-				"
-				select
-					user_id,
-					is_host,
-					is_new_contributor,
-					(
-						select group_concat( distinct locale )
-						from {$gp_table_prefix}event_actions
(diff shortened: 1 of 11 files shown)


includes/attendee/attendee-repository.php
```diff
@@ -86,47 +86,77 @@ public function remove_attendee( int $event_id, int $user_id ): void {
...
+		}
+		return $attendees[0];
+	}
+
+	/**
+	 * @var int[] $event_ids
+	 * @return Attendee[]
+	 * @throws Exception
+	 */
+	public function get_attendees_for_events_for_user( array $event_ids, int $user_id ): array {
```
trymebytes, line 102: I have no idea what we can rename this function to but it may be prone to typo errors when considering the preceding function in line 89 `public function get_attendee_for_event_for_user( int $event_id, int $user_id )` even though they serve different purposes.

includes/event/event-repository.php
```diff
@@ -223,6 +223,23 @@ public function get_trashed_events( int $page = - 1, int $page_size = - 1 ): Eve
...
+	public function get_events_for_user( int $user_id, int $page = -1, int $page_size = -1 ): Events_Query_Result {
+		// phpcs:disable WordPress.DB.SlowDBQuery.slow_db_query_meta_key
+		return $this->execute_events_query(
+			$page,
+			$page_size,
+			array(
+				'meta_key' => '_event_start',
+				'orderby'  => 'meta_value',
+				'order'    => 'DESC',
+			),
```
akirk, line 235: I am not super happy with this meta query. I just realized, could we not just sync the post's date with the event start time? Then we could use standard WordPress sorting. I've recorded this in https://github.com/WordPress/wporg-gp-translation-events/issues/292, we can keep it like this for now.

includes/attendee/attendee-repository.php
akirk: I agree that it's a bit of a verbose name but I'm ok with keeping it for now and updating it in a later PR if we find a better name.

includes/attendee/attendee-repository.php
```diff
@@ -86,47 +86,77 @@ public function remove_attendee( int $event_id, int $user_id ): void {
 	/**
 	 * @throws Exception
 	 */
-	public function get_attendee( int $event_id, int $user_id ): ?Attendee {
+	public function get_attendee_for_event_for_user( int $event_id, int $user_id ): ?Attendee {
+		$attendees = $this->get_attendees_for_events_for_user( array( $event_id ), $user_id );
+		if ( empty( $attendees ) ) {
```
akirk: Would this be too risky?
```suggestion
		if ( 1 !== count( $attendees ) ) {
```

includes/attendee/attendee-repository.php
```diff
@@ -86,47 +86,77 @@ public function remove_attendee( int $event_id, int $user_id ): void {
...
+					from {$gp_table_prefix}event_attendees attendees
+					where event_id in ($event_id_params)
+					  and user_id = %d
+				",
+				array_merge(
+					$event_ids,
+					array( $user_id ),
 				),
-			)
+			),
```
akirk, line 145: Would we get a benefit from this? (We'd need to change the `$attendees[0]` above.

```suggestion
			OBJECT_K
			)
```

Otherwise we should be careful with the trailing comma in the function, this is pretty new syntax for php ([since 7.3](https://wiki.php.net/rfc/trailing-comma-function-calls)).

includes/event/event-repository.php
```diff
@@ -533,3 +564,27 @@ function ( $host ) {
 		update_post_meta( $event->id(), '_hosts', $hosts_ids );
 	}
 }
+
+// phpcs:ignore Universal.Files.SeparateFunctionsFromOO.Mixed
+function add_user_id_where_clause_to_events_query( string $where, WP_Query $query ): string {
+	// Remove the filter, so it only applies to a single query.
+	remove_filter( 'posts_where', 'add_user_id_where_clause_to_events_query' );
```
akirk: I haven't tested it but as far as I know it needs to have the exact same function parameter except for the arguments:
```suggestion
	remove_filter( 'posts_where', 'Wporg\TranslationEvents\Event\add_user_id_where_clause_to_events_query', 10 );
```

akirk (approved):
Found a few small things and I think we can iterate on the overall approach regarding reducing complexity in the SQL queries and doing more on the PHP side in later PRs when we need to address performance again.

includes/attendee/attendee-repository.php
psrpinto: Maybe we can rename the other one to `get_single_attendee()`.

includes/event/event-repository.php
psrpinto: This is a bug indeed, great catch. However, setting the correct value here makes tests fail, so I'm looking into what's happening.

psrpinto: This is now fixed. Concerning the extra argument, `10`, it's not needed as `10` is the default value..

includes/attendee/attendee-repository.php
psrpinto: Makes sense, that will allow us to not have to do [this](https://github.com/WordPress/wporg-gp-translation-events/pull/282/files#diff-706179ae733561a4c131b9d753beb83d4cc944de6de99f5c8beb3c0fdc4c1f9dR52-R55).

psrpinto:
I've addressed your remarks @akirk, could you take a quick look?

includes/event/event-repository.php
```diff
@@ -454,10 +476,25 @@ private function execute_events_query( int $page, int $page_size, array $args, a
...
+			$args['translation_events_user_id']                 = $user_id;
+			$args['translation_events_include_created_by_user'] = $include_created_by_user;
+		}
 
+		$query = new WP_Query( $args );
+		$posts = $query->get_posts();
+
+		if ( isset( $user_id_filter_callback ) ) {
+			// Remove the filter, so it only applies to this query.
+			remove_filter( 'posts_where', $user_id_filter_callback );
```
akirk: Note this will work because we use the default priority 10, otherwise we'd need to specify it.

akirk (approved):

includes/event/event-repository.php
psrpinto: Good point. I've now added a variable so it will always be set to the same value as the `add_filter()`.

PR #289, opened 2024-05-21
Fix repeated footer

trymebytes:
A footer is displayed for every project translated on the [translations page](https://translate.wordpress.org/events/2024/wordcamp-porto-portugal-2024-contributor-day/translations/bn/), see below;

image

In this PR we fix that by loading the GlotPress footer at the end of the loop that renders the translated projects.

### DIFF
--- a/templates/event-translations-footer.php
+++ b/templates/event-translations-footer.php
@@ -38,4 +38,4 @@ foreach ( $editor_options as $translation_set_id => $options ) {
 </script>
 <?php
 gp_enqueue_script( 'wporg-translate-editor' );
-gp_tmpl_footer(); ?>
+ ?>
--- a/templates/event-translations-footer.php
+++ b/templates/event-translations-footer.php
@@ -40,4 +40,3 @@ foreach ( $editor_options as $translation_set_id => $options ) {
 <?php
 gp_enqueue_script( 'wporg-translate-editor' );
 Templates::footer();
-?>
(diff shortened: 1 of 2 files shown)


templates/event-translations-footer.php
```diff
@@ -33,4 +33,4 @@ function set_translation_table_<?php echo esc_html( $translation_set_id ); ?>()
 </script>
 <?php
 gp_enqueue_script( 'wporg-translate-editor' );
-gp_tmpl_footer(); ?>
```
psrpinto, line 41: I'm not sure I understand why this fixes the issue, does this mean that `templates/event-translations-footer.php` is being included multiple times? I searched for `event-translations-footer.php` but it only seems to be included once, in `includes/routes/event/translations.php`.

trymebytes: You are correct, just realised I hadn't moved it outside the for loop. Will continue investigating to determine the true cause or why this fixes it, thank you!

templates/event-translations-footer.php
```diff
@@ -38,4 +39,5 @@ function set_translation_table_<?php echo esc_html( $translation_set_id ); ?>()
...
+Templates::footer();
+?>
```
akirk: This can be omitted:
```suggestion

```

trymebytes:
This appears to have solved the breadcrumb issue for me.

psrpinto (approved):
👍

PR #290, opened 2024-05-21
Temporarily remove the waiting column from the stats table, to fix stats issues

### LINKED ISSUE #285: Inconsistent data between stats for each locale and the list of translations
In the [Contributor Day of WordCamp Porto 2024, Portugal](https://translate.wordpress.org/events/2024/wordcamp-porto-portugal-2024-contributor-day/), we have some inconsistent data between stats for each locale and the list of translations. 

E.g., the first locale ([Afrikaans](https://translate.wordpress.org/languages/af/)) says it has 36 translations created, 35 in waiting status:

image

The ["created"](https://translate.wordpress.org/events/2024/wordcamp-porto-portugal-2024-contributor-day/translations/af/) link shows only 1 string:

image

And the ["waiting"](https://translate.wordpress.org/events/2024/wordcamp-porto-portugal-2024-contributor-day/translations/af/waiting/) link shows the same string:

image

Props @pedro-mendonca.

psrpinto:
Fixes #285

I wasn't able to figure out yet what exactly causes the issue, but the issues with stats started once we introduced the waiting column. So in this PR, we're removing that column from the table. This is the quickest way to get a fix deployed, so that stats in production are correct. Then we can look into what it would take to bring back the waiting column.

The problem that the waiting column introduces is that we're now relying on joining with the translations table when computing stats. However, the model we had defined initially was based on the principle that stats would be calculated from the `actions` table.

Joining with the translations table means that, when calculating stats, we need to deal with the complexities of the lifecycle of translations. I think we should keep the principle that we defined initially that stats are calculated from the translations table alone.

So I think that when we bring back the waiting column after this PR is merged, we should look into how we can do so while only relying on the `actions` table for stats.

### DIFF
--- a/includes/stats/stats-calculator.php
+++ b/includes/stats/stats-calculator.php
@@ -9,14 +9,12 @@ use GP_Locales;
 class Stats_Row {
 	public int $created;
 	public int $reviewed;
-	public int $waiting;
 	public int $users;
 	public ?GP_Locale $language = null;
 
-	public function __construct( $created, $reviewed, $waiting, $users, ?GP_Locale $language = null ) {
+	public function __construct( $created, $reviewed, $users, ?GP_Locale $language = null ) {
 		$this->created  = $created;
 		$this->reviewed = $reviewed;
-		$this->waiting  = $waiting;
 		$this->users    = $users;
 		$this->language = $language;
 	}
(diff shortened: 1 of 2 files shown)


akirk (approved):
I agree this is the best path forward, also performance-wise. I think we could bring it back via AJAX later

amieiro (commented):
Tested in the sandbox and works fine. 

image

Fixes #291.

PR #293, opened 2024-05-23
Add the header in the new style of translate.wordpress.org/events

### LINKED ISSUE #286: Update header and footer designs
See Figma designs here https://www.figma.com/design/Q3RpUkAJhTaKVmXdGGJIBs/Translating-WordPress?node-id=41-950 for guidance.

amieiro:
This PR adds the header in the new style of https://translate.wordpress.org/events/, as a previous step to add it to the [wporg-gp-customizations](https://github.com/WordPress/wordpress.org/tree/66d51b3e40626d08912777a1ed00dff1576b70ce/wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations) plugin.

It adds:
- One new CSS file.
- A `Home` link in the breadcrumbs.

## Header

This is the current header:

image

This is the proposed header by the designer:

image

And this is the header we achieve with this PR (only the red square):

image

## Footer

This is the proposed footer by the designer:

Image 2024-05-23 19-29-06

I am not adding the block in the red square, because we don't have any content for this.

And this is the current design, so we don't need any work in the footer:

image


wporg-gp-translation-events.php
```diff
@@ -255,7 +255,11 @@ public function submit_event_ajax() {
 
 	public function register_translation_event_js() {
 		wp_register_style( 'translation-events-css', plugins_url( 'assets/css/translation-events.css', __FILE__ ), array( 'dashicons' ), filemtime( __DIR__ . '/assets/css/translation-events.css' ) );
-		gp_enqueue_style( 'translation-events-css' );
+		gp_enqueue_styles( 'translation-events-css' );
+		if ( $this->is_dot_org_or_dev_tld() ) {
```
akirk:
```suggestion
		if ( $this->should_load_new_design() ) {
```

wporg-gp-translation-events.php
```diff
@@ -428,5 +432,25 @@ public function pre_wp_unique_post_slug( $override_slug, string $slug, int $post
...
+		$parsed_url   = wp_parse_url( $site_url );
+		$host         = $parsed_url['host'];
+
+		$domain_parts = explode( '.', $host );
+		$tld          = end( $domain_parts );
+		if ( 'translate.wordpress.org' === $host || in_array( $tld, $allowed_tlds, true ) ) {
+			return true;
+		}
+		return false;
+	}
```
akirk: I use another local host format for this and it will be hard to include everyone. What if we did this?
```suggestion
	private function should_load_new_design(): bool {
		return 'translate.wordpress.org' === $host || (defined( 'TRANSLATION_EVENTS_NEW_DESIGN' ) && TRANSLATION_EVENTS_NEW_DESIGN );
	}
	
```

wporg-gp-translation-events.php
```diff
@@ -429,5 +433,17 @@ public function pre_wp_unique_post_slug( $override_slug, string $slug, int $post
...
+
+	/**
+	 * Check if the current site is a translate.wordpress.org or a development TLD.
+	 *
+	 * @return bool
+	 */
+	private function should_load_new_design(): bool {
+		$site_url   = get_site_url();
+		$parsed_url = wp_parse_url( $site_url );
+		$host       = $parsed_url['host'];
```
akirk:
```suggestion
		$host = wp_parse_url( get_site_url(), PHP_URL_HOST );
```

akirk:
In your PR, so far, you still use the [wporg-gp-customizations header.php](https://github.com/WordPress/wordpress.org/blob/trunk/wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/templates/header.php) and override some of it via CSS.

I wonder if another approach would be to do this:
```diff
diff --git i/templates/partials/header.php w/templates/partials/header.php
index 682038c..06487f7 100644
--- i/templates/partials/header.php
+++ w/templates/partials/header.php
@@ -42,7 +42,7 @@ add_action(
 );
 gp_title( $html_title );
 gp_breadcrumb_translation_events( $breadcrumbs ?? array() );
-gp_tmpl_header();
+echo do_blocks( '<!-- wp:wporg/global-header /-->' );
 ?>
 
 <div class="event-list-top-bar">
```

assets/css/new-dotorg-design.css
```diff
@@ -0,0 +1,50 @@
+body {
+	--wp--preset--color--charcoal-2: #23282d;
+	--wp-global-header--background-color: var(--wp--preset--color--charcoal-2);
+	--wp--preset--color--charcoal-4: #656a71;
+	--wp--preset--color--white-opacity-15: #ffffff26;
+}
```
psrpinto, line 9: I'm assuming you copied this CSS from somewhere else, could we add an inline comment that points to the original source?

I think once we have a first version of the design, we should look into how to not duplicate CSS of the theme, so it will be useful to know where the CSS came from.

psrpinto (approved):
👍

trymebytes:
I tested this PR and it looks good on the events page but this is how it looks on the translate.wordpress.org homepage, this test was done in my sandbox.
Screenshot 2024-05-24 at 14 57 07

wporg-gp-translation-events.php
```diff
@@ -429,5 +433,15 @@ public function pre_wp_unique_post_slug( $override_slug, string $slug, int $post
...
 	}
+
+	/**
+	 * Check if the current site is a translate.wordpress.org or a development TLD.
+	 *
+	 * @return bool
+	 */
+	private function should_load_new_design(): bool {
+		$host = wp_parse_url( get_site_url(), PHP_URL_HOST );
+		return 'translate.wordpress.org' === $host || ( defined( 'TRANSLATION_EVENTS_NEW_DESIGN' ) && TRANSLATION_EVENTS_NEW_DESIGN );
```
psrpinto: I'm wondering, should we disable the new design on `translate.wordpress.org` for now, until we have finalised it? This would allow us to still deploy the plugin to production (e.g. for bug fixes). Once we have completed the design, we would enable it for `translate.wordpress.org` here.

psrpinto (changes requested):
I can't undo an approve, so requesting changes instead, as I think we need to disable it for production.

psrpinto:
I agree with @akirk's [comment](https://github.com/WordPress/wporg-gp-translation-events/pull/293#issuecomment-2128963237) about using the header block. This appears to be the recommended way to do it, as documented in [Global Header and Footer - Classic themes in the w.org network](https://github.com/WordPress/wporg-mu-plugins/tree/trunk/mu-plugins/blocks/global-header-footer#classic-themes-in-the-worg-network).

We could do that in another PR, since it would be useful to have this one merged quickly, so that the new CSS file is on trunk.

wporg-gp-translation-events.php
amieiro: Yes, you are right. If a developer wants to have it enable in his sandbox, he can add the constant to the wp-config.php file. Removed [here](https://github.com/WordPress/wporg-gp-translation-events/pull/293/commits/e8c867be61ad2521409df0a985af8c9cfe7f89d7).

assets/css/new-dotorg-design.css
amieiro: I added the URL in the header of the file.

psrpinto (approved):

PR #294, opened 2024-05-23
Fix randomly failing tests

psrpinto:
As you might have noticed, the tests randomly fail, with something like:

image

I think this is due to the value of 'now' not staying the same throughout the test, which results in the order of events returned from the repository to not be consistent. If the seconds portion of 'now' changes while the test was running, the issue will manifest itself.

The solution to this is to make the value of 'now' the same throughout the test. That's what this PR does, along with some other small cleanups I noticed while fixing the 'now' issue.

### DIFF
--- a/tests/urls.php
+++ b/tests/urls.php
@@ -2,6 +2,9 @@
 
 namespace Wporg\Tests;
 
+use DateTime;
+use DateTimeImmutable;
+use DateTimeZone;
 use GP_UnitTestCase;
 use Wporg\TranslationEvents\Attendee\Attendee_Repository;
 use Wporg\TranslationEvents\Event\Event_Repository;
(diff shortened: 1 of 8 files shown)


trymebytes (approved):
Looks good to me!

tests/urls.php
```diff
@@ -64,37 +69,32 @@ public function test_event_translations() {
 	}
 
 	public function test_event_edit() {
-		$event_id = $this->event_factory->create_active();
-
+		$event_id = 42;
```
akirk, line 72: Maybe worth adding a comment here why we don't need a real event here?

akirk (approved):
Nice, thank you!

tests/urls.php
psrpinto: We're testing the `Urls` class which doesn't interact with the database, it takes the event id as argument, and returns the URL. It will work regardless of whether the event exists in the database or not.

I think what happened was that we copy-pasted from another test that was testing the slug, and since the slug is calculated automatically upon event creation, in those tests we do rely on creating the event to get the slug.

psrpinto:
Thanks @amieiro for bringing attention to this issue.

PR #309, opened 2024-05-28
Only show Permanently Delete Button on Deleted Events page

akirk:
## Before
Currently an Admin will see a permanently delete button on all pages. 

Screenshot 2024-05-28 at 13 40 51

## After
I'd like to avoid accidentally deleting an event, so this button now only shows on the trashed events page:

Screenshot 2024-05-28 at 13 42 08

Screenshot 2024-05-28 at 13 42 16

### DIFF
--- a/templates/events-list-trashed.php
+++ b/templates/events-list-trashed.php
@@ -28,7 +28,7 @@ Templates::header(
 					'show_start'             => true,
 					'show_end'               => true,
 					'relative_time'          => false,
-					'allow_permanent_delete' => true,
+					'show_permanent_delete'  => true,
 				),
 			);
 			?>
--- a/templates/events-list-trashed.php
+++ b/templates/events-list-trashed.php
@@ -28,6 +28,7 @@ Templates::header(
 					'show_start'             => true,
 					'show_end'               => true,
 					'relative_time'          => false,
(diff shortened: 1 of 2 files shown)


amieiro:
Currently, I can't see the "Delete permanently" in the trunk branch. 

image

image

image

I can see the button only in the "Deleted Translation Events" page.

image

akirk:
Well, you need a user that has the `delete_translation_event` permission, either through a filter or by being Super Admin.

trymebytes (approved):
Tested and works well!

amieiro:
I have tested it, and it works fine. In the `event_repository`, what do you think to check if the event has been trashed before deleted it? 

```
	public function delete_event( Event $event ) {
		if ( ! $event->is_trashed() ) {
			return false;
		}
```

psrpinto:
> In the event_repository, what do you think to check if the event has been trashed before deleted it? 

That check is being [done in the capabilities](https://github.com/WordPress/wporg-gp-translation-events/blob/trunk/includes/event/event-capabilities.php#L219). I think it shouldn't be in the repository because there might be situations where you want to be able to delete without it being trashed.

amieiro (approved):

PR #311, opened 2024-05-28
Remove the underline from the edit and delete buttons

amieiro:
In the event lists, the Edit and Delete buttons has an unusual underline.

image

This PR removes it.

image

### DIFF
--- a/assets/css/translation-events.css
+++ b/assets/css/translation-events.css
@@ -174,7 +174,9 @@ span.event-list-date {
 }
 
 /* Event list: Event buttons. */
-.event-list-item-button {
+a.event-list-item-button,
+a.event-list-item-button:visited,
+a.event-list-item-button:hover {
 	text-decoration: none;
 	vertical-align: bottom;
 }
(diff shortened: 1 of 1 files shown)


trymebytes (approved):
Tested and works as expected!

assets/css/translation-events.css
```diff
@@ -174,7 +174,9 @@ span.event-list-date {
 }
 
 /* Event list: Event buttons. */
-.event-list-item-button {
+a.event-list-item-button,
+a.event-list-item-button:visited,
+a.event-list-item-button:hover {
```
akirk: FWIW there is now a `:any-link` CSS Pseudo class: https://developer.mozilla.org/en-US/docs/Web/CSS/:any-link

psrpinto (approved):
Thanks for fixing!

I don't get why it worked in development environments, but not in production.

assets/css/translation-events.css
amieiro: Tested and works fine in all browsers, including Safari. Thank you!

PR #313, opened 2024-05-29
Change title to trash

trymebytes:
The icon title is different from the purpose of the icon.

**Before**

Screenshot 2024-05-29 at 11 35 50

**After**

Screenshot 2024-05-29 at 14 16 15

### DIFF
--- a/templates/partials/event-list.php
+++ b/templates/partials/event-list.php
@@ -89,7 +89,7 @@ $print_time = function ( $time ) use ( $date_format, $relative_time ): void {
 				<?php else : ?>
 					<a href="<?php echo esc_url( Urls::event_trash( $event->id() ) ); ?>"
 						class="event-list-item-button is-destructive"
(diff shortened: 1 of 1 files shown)


templates/partials/event-list.php
```diff
@@ -89,7 +89,7 @@ class="button is-small"
 				<?php else : ?>
 					<a href="<?php echo esc_url( Urls::event_trash( $event->id() ) ); ?>"
 						class="event-list-item-button is-destructive"
-						title="<?php echo esc_attr__( 'Delete', 'gp-translation-events' ); ?>">
+						title="<?php echo esc_attr__( 'Trash', 'gp-translation-events' ); ?>">
```
psrpinto: Maybe we could use 'Move to trash' instead, I think @akirk wasn't a fan of using Trash as verb.

akirk: I like "Move to trash"!

amieiro (approved):

psrpinto (approved):

PR #314, opened 2024-06-03
Reorganize templates, before new design

psrpinto:
When implementing the new design, I think it will be important to be able to tackle it one page at a time. So it's useful to have each page in its own template. I thought there was also room for improvement in the naming of the templates.

This PR makes each page have its own template, and improves the naming of each template.

I also think it's useful to start thinking of elements on the pages as "blocks" (for example breadcrumbs), so this PR also improves on that front, by creating template parts for the breadcrumbs and the event form.

Once this is merged, we can create a `templates/new-design/` directory, and create a copy of the template for each page there, while keeping the original template in place. This will allow us to iterate on the templates of the new design, without affecting the templates in production. (I will be setting this up in an upcoming PR).

## Summary of changes

- Add a dedicated template for the edit event and the create event page. Previously both pages shared the same template. We do this by making the form a template part, and then including the form part from the templates of both pages
- Make breadcrumbs a template part instead of a function that the template calls. Breadcrumbs are a "block" on the page, so I think it makes sense to make them a template part.
- Rename templates for clarity, so it's more obvious which page the template is for

## Screen captures

| Before | **After** |
|--------|-------|
|  Screenshot 2024-06-03 at 14 55 33 | Screenshot 2024-06-03 at 14 55 15 |


akirk (approved):
I clicked through the locally checked-out PR are and everything worked fine. Also extracting the breadcrumbs is fine to me. Let's proceed!

trymebytes (approved):
Tested and works as expected!

PR #315, opened 2024-06-03
Setup for new design

psrpinto:
This PR creates a new `templates/new-design/` directory, which would contain the templates for the new design. At this moment, it contains placeholders for the header, footer and the my-events page (this page should serve as an example).

The idea would be that we implement the new design page by page in the `templates/new-design/` directory, and leave the original templates under `templates/` untouched. 

Additionally, there is now a new `Templates::use_new_design()` function, that should be called by the routes. Whenever we create a template for the new design of a given page, we make the route call this function. When the design for said page is ready to use also in production, we can pass as extra argument to that function (`Templates::use_new_design( true )`) which will make it also be used in production.

Whenever we migrated all the pages to the new design, we can remove the old templates, and move the ones under `templates/new-design/` to the root `templates/` directory.

Additionally, `Templates::use_new_design()` also loads the `new-design.css`.

## Next steps

Once this PR merged, we can iteratively start migrating pages to the new design. It would probably make sense to start with the header and footer parts.


akirk (approved):
Nice! Worth mentioning that the new design will load on a dev machine only when you `define( 'TRANSLATION_EVENTS_NEW_DESIGN', true );`

amieiro (approved):

PR #316, opened 2024-06-04
Add the new site-header template for the DotOrg design

amieiro:
This PR implements the new design for the site-header at [translate.wordpress.org](https://translate.wordpress.org/), using only blocks. This PR:

- Adds a new `site-header` template in the new design.
  - Adds the `global-header`	block.
  - Adds a hard-coded menu.
  - Adds some inline CSS, copied from other site at wordpress.org, that should be removed in a future PR.
- Removes the call to the `gp_tmpl_header();` method.
- Adds the new header and footer in the new-design. We need to iterate in these templates.

Currently, this PR only works in [My events](https://translate.wordpress.org/events/my-events/), because we don't have another main templates in the new design.  

**New design with the current template system**

image

**New design with blocks**

image

Currently, we have a problem with the breadcrumb at [translate.wordpress.org](https://translate.wordpress.org/), because it is not working. We should address this problem in a future PR.


templates/parts/site-header.php
```diff
@@ -105,3 +105,24 @@ function add_css() {
...
+		echo '<meta property="og:url" content="' . esc_url( $url ) . '" />' . "\n";
+		echo '<meta property="og:title" content="' . esc_attr( $html_title ) . '" />' . "\n";
+		echo '<meta property="og:description" content="' . esc_attr( $html_description ) . '" />' . "\n";
+		echo '<meta property="og:site_name" content="' . esc_attr( get_bloginfo() ) . '" />' . "\n";
+		echo '<meta property="og:image:url" content="' . esc_url( $image_url ) . '" />' . "\n";
+		echo '<meta property="og:image:secure_url" content="' . esc_url( $image_url ) . '" />' . "\n";
+		echo '<meta property="og:image:type" content="image/png" />' . "\n";
+		echo '<meta property="og:image:width" content="1200" />' . "\n";
+		echo '<meta property="og:image:height" content="675" />' . "\n";
+		echo '<meta property="og:image:alt" content="' . esc_attr( $html_title ) . '" />' . "\n";
```
akirk: What about doing this with an array and a foreach loop? This way it's quite hard to read.

amieiro: Done [here](https://github.com/WordPress/wporg-gp-translation-events/pull/316/commits/41d2c2968990438822edc38a6f48e5e78bc55e57).

templates/parts/header.php
```diff
@@ -17,6 +17,7 @@
 $html_description = $html_description ?? esc_html__( 'WordPress Translation Events', 'gp-translation-events' );
 $image_url        = $image_url ?? Urls::event_default_image();
 
+// to-do: this action should be deleted when the new design is implemented, because it is in the "site-header" template.
```
psrpinto, line 20: What I was thinking was that we would replace this file with `new-design/parts/header.php`, once the new design implementation is complete.

templates/new-design/parts/site-header.php
```diff
@@ -0,0 +1,150 @@
...
+/** @var string $html_title */
+/** @var string $url */
+/** @var string $html_description */
+/** @var string $image_url */
+
+// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+echo do_blocks( '<!-- wp:wporg/global-header /-->' );
+
+gp_enqueue_styles( array( 'gp-jquery-webui-popover', 'driver-js' ) );
+gp_enqueue_scripts( array( 'gp-tour' ) );
```
psrpinto, line 13: Would it make sense to use `wp_enqueue_styles()` and `wp_enqueue_scripts()` instead? Also, I'm wondering if we need the tour in the Translation Events (I'm assuming `jquery-webui-popover` and `driver-js` are for the tour).

psrpinto (commented):
Tested locally and works nicely 👍

templates/new-design/parts/header.php
```diff
@@ -1,4 +1,52 @@
 <?php
+namespace Wporg\TranslationEvents\Templates\NewDesign\Parts;
```
psrpinto, line 2: I noticed that `glotpress/assets/css/style.css` is being included on the page, I'm not sure exactly what is causing that. The idea with the `new-design/` directory was that that would be a "clean slate" that doesn't have any of the glotpress styles.

Any idea what's causing this style to be included? It's visible in the screenshot that the style is included because the Translation Events header has the old styling.

templates/new-design/parts/header.php
```diff
@@ -1,4 +1,52 @@
...
+$html_description = $html_description ?? esc_html__( 'WordPress Translation Events', 'gp-translation-events' );
+$image_url        = $image_url ?? Urls::event_default_image();
+
+gp_title( $html_title );
+Templates::part( 'site-header', get_defined_vars() );
+Templates::part( 'breadcrumbs', array( 'extra_items' => $breadcrumbs ?? array() ) );
+
+?>
+
+<div class="event-list-top-bar">
```
psrpinto, line 26: I think it could make sense to remove this from this PR for now, as it doesn't currently adhere to the new design, and implement it in a later PR with the new design.

templates/parts/header.php
amieiro: We should do this based on the answer to: do we want this template system to be tied to the DotOrg blocks or do we want to allow a user to install GlotPress, then this plugin and everything will work correctly? I think we should talk about this shortly.

templates/new-design/parts/header.php
amieiro: I think this is because at translate.w.org, GlotPress loads on every call, so it loads its own CSS. It happens the same with other plugins like gp-translation-helpers.

templates/parts/header.php
psrpinto: I agree indeed, we need to discuss this.

psrpinto (approved):
Approving since we we can improve it other PRs, if needed.

psrpinto:
Merging, we can iterate in future PRs.

PR #319, opened 2024-06-05
Fix flaky tests due to value of now changing

### LINKED ISSUE #308: test_get_past_events_for_user is flaky
It failed here: https://github.com/WordPress/wporg-gp-translation-events/actions/runs/9268332809/job/25496731899
Screenshot 2024-05-28 at 13 22 21

And worked on the re-run: https://github.com/WordPress/wporg-gp-translation-events/actions/runs/9268332809/job/25496794775

psrpinto:
Fixes #308 

This PR (hopefully) fixes the issues with flaky tests, which we assume have to do with the value of _now_ not staying the same throughout the request. To fix the issue, the value of _now_ must be the same through the codebase, so we introduce a `Translation_Events::now()` function from which the value of _now_ must come from.

This means that we should never directly create an instance of `DateTimeImmutable` that represents _now_, we should always use `Translation_Events::now()` by injecting it into whoever needs it, or by calling it directly if injection is not an option.

This PR also adds a `Base_Test` that all tests should extend. The base test defines _now_, so all tests have `$this->now` available.

There were also issues with ordering not being guaranteed in tests, because of events that are created with the exact same start and end, which would result in randomness in the ordering, for example:

```php
$event1_id = $this->event_factory->create_active( $this->now );
$event2_id = $this->event_factory->create_active( $this->now );

$events = $this->repository->get_events_created_by_user( $user_id )->events;
```

In the code above, it will be random which event ends up as `$events[0]`. So instead we need to do:

```php
$event1_id = $this->event_factory->create_active( $this->now );
$event2_id = $this->event_factory->create_active( $this->now->modify( '+1 minute' ) );

$events = $this->repository->get_events_created_by_user( $user_id )->events;
```

which will result in `event_2` always being at `$events[0]`, because `get_events_created_by_user()` orders by `event_start` descending.

### DIFF
--- a/includes/event/event-capabilities.php
+++ b/includes/event/event-capabilities.php
@@ -43,15 +43,18 @@ class Event_Capabilities {
 		self::EDIT_TIMEZONE,
 	);
 
+	private DateTimeImmutable $now;
 	private Event_Repository_Interface $event_repository;
 	private Attendee_Repository $attendee_repository;
 	private Stats_Calculator $stats_calculator;
 
 	public function __construct(
+		DateTimeImmutable $now,
 		Event_Repository_Interface $event_repository,
 		Attendee_Repository $attendee_repository,
 		Stats_Calculator $stats_calculator
 	) {
+		$this->now                 = $now;
 		$this->event_repository    = $event_repository;
 		$this->attendee_repository = $attendee_repository;
 		$this->stats_calculator    = $stats_calculator;
@@ -260,14 +263,13 @@ class Event_Capabilities {
 	 * @return bool
 	 */
 	private function has_edit_field( WP_User $user, Event $event, $cap ): bool {
-		$now                 = new DateTimeImmutable( 'now', new DateTimeZone( 'UTC' ) );
 		$event_end_plus_1_hr = $event->end()->modify( '+1 hour' );
 
 		if ( self::EDIT_DESCRIPTION === $cap ) {
 			return true;
 		}
 
-		if ( $event->start() > $now ) {
+		if ( $event->start() > $this->now ) {
 			return true;
 		}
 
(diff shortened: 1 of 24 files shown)


akirk (approved):
This is great, systematically implemented! Thank you! Fine to go ahead with this, I just have one remark:

> which will result in event_2 always being at $events[0], because get_events_created_by_user() orders by event_start descending.

Could we add sorting by ID ASC to the sort order? https://developer.wordpress.org/reference/classes/wp_query/#order-orderby-parameters says:

> order (string | array) – Designates the ascending or descending order of the ‘orderby‘ parameter. Defaults to ‘DESC’. An array can be used for multiple order/orderby sets.

psrpinto:
Currently, in `get_events_created_by_user()` (the example I used in the PR description) , we are already [ordering by `event_start` descending](https://github.com/WordPress/wporg-gp-translation-events/blob/trunk/includes/event/event-repository.php#L341-L343), because I think we want show the latest events first.

For other repository methods, we are ordering in other ways. For example, `get_current_and_upcoming_events_for_user()` orders by event start ascending, because in that case the results go from now to the future.

akirk:
What I am saying is that we additionally sort by ID ASC, so that if two events start at the same time (like in the tests) we have a predictable sort order.

psrpinto:
Ah, that makes sense! Sorry, I didn't correctly read what you said. I will do it here.

psrpinto:
While making it so that we also order by `ID`, as @akirk suggested above, I realised there were actually bugs with ordering, because the syntax for `orderby` with multiple fields doesn't seem to work as we expected. Instead we need to do:

```php
'meta_key' => '_event_start',
'meta_type => 'DATETIME',
'orderby'    => array(
	'meta_value' => 'ASC',
	'ID'         => 'ASC',
),
```

There were also other potential issues, like some queries not specifying `meta_key` and `meta_type`.

All this is now addressed and I think ordering is now working as expected in all methods.

While I was at it, I also made it so that past events are now ordered `DESC`, which will fix the issue with older events being shown first in the "Past events" section of the homepage.

akirk (approved):
Great update to the PR, and also great that it revealed some bugs! Thanks for addressing this!

PR #320, opened 2024-06-05
Also show absolute time in the events list

### LINKED ISSUE #288: In events time line there should be exact date and time stamp
rtcamp

**Currently in all the events, its just displaying the weeks, not the exact date and time**

akirk:
Fixes #288.

This now also shows the absolute date in list view and moves the relative day into parenthesis.

### Old
rtcamp

### New
Screenshot 2024-06-05 at 16 04 17

### DIFF
--- a/templates/parts/event-list.php
+++ b/templates/parts/event-list.php
@@ -25,7 +25,7 @@ $current_user_attendee_per_event = $current_user_attendee_per_event ?? array();
 /**
  * @param Event_Start_Date|Event_End_Date $time
  */
-$print_time = function ( $time ) : void {
+$print_time = function ( $time ): void {
 	$time->print_absolute_and_relative_time_html();
 };
 ?>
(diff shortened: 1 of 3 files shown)


templates/parts/event-list.php
```diff
@@ -12,29 +12,21 @@
...
 /**
  * @param Event_Start_Date|Event_End_Date $time
  */
-$print_time = function ( $time ) use ( $date_format, $relative_time ): void {
-	if ( $relative_time ) {
-		$time->print_relative_time_html( $date_format );
-	} else {
-		$time->print_time_html( $date_format );
-	}
+$print_time = function ( $time ) : void {
```
psrpinto: We could remove the `$print_time` function as the logic it contained has now been moved to the `Event_Date`.

psrpinto (commented):
I'm a bit confused by the JavaScript changes, I thought the JS was only for the event form, is it also used for rendering the event lists?

akirk:
The JS autoupdates the time every 10 seconds. It's for situations where you have the event page open and then the event starts. It's a bit of a gimmick.

psrpinto (approved):

PR #321, opened 2024-06-07
Set up block theme

psrpinto:
- Please review only the final result instead of individual commits as there was a lot of trial and error.
- I would recommend checking out the branch and taking a look at the `themes/wporg-translate-events-2024` directory, which contains the new theme.
- The new theme is where the bulk of the changes are, other changes were just to "fake" that theme as the active theme for the current request.

----

This PR introduces a new block theme to be used for translation events, and potentially in the future for the whole of translate.wordpress.org (`wporg-translate`). For the moment, the theme is contained in this repository, but we can extract it into its own repository later, once/if we decide to use the theme for the rest of `wporg-translate`.

**Before proceeding, please read the new theme's [README](https://github.com/WordPress/wporg-gp-translation-events/blob/new-design-setup/themes/wporg-translate-events-2024/README.md)**, which contains valuable information that is useful to review this PR.

## Summary of changes

- [Document](https://github.com/WordPress/wporg-gp-translation-events/blob/new-design-setup/themes/wporg-translate-events-2024/README.md#context) how themes work at `translate.wordpress.org`
- Remove the `templates/new-design` directory
- Create a a new block theme under `themes/wporg-translate-events-2024`
- Add a mechanism to enable the new theme only for specific pages (call `$this->use_theme()` on the `Route` of the page)
- Implement the block theme as described in [this section of the theme's README](https://github.com/WordPress/wporg-gp-translation-events/blob/new-design-setup/themes/wporg-translate-events-2024/README.md#how-it-works-now-with-this-theme)

## How to review and test

1. Pull `translate-w-org-env` and run the provision script again (it will take a while, probably). This is to get the changes introduced by:
    - https://github.com/WordPress/translate-w-org-env/pull/7
    - https://github.com/WordPress/translate-w-org-env/pull/8, 
2. Make sure the Gutenberg plugin is installed/activated, and if not install/activate it
3. Checkout this branch
4. Set the `TRANSLATION_EVENTS_NEW_DESIGN` env variable to `true`
5. Navigate to the "My Events" page
    - You should see the same as the screenshot below
6. Navigate to other pages and make sure they look and function as they did

## Screen capture

This is how a page looks when using the theme.

Screenshot 2024-07-10 at 15 03 10

## TODO

- [x] https://github.com/WordPress/translate-w-org-env/pull/7
- [x] Fix issue with styles when new design not in use
- [x] Remove `add_css()` function
- [x] Remove styles from active theme
- [x] Remove wporg-gp-customizations styles and scripts
- [x] Include styles from parent theme
- [x] Remove `global-styles-inline-css` in favour of the parent theme styles
- [x] Undeprecate header and footer?
- [x] Create a theme
- [x] Load the custom theme
- [x] Implement the basic theme
- [x] Figure out why styles are broken when the theme is not the active theme (it works correctly when it is the active theme)
- [x] Figure out the correct way to use templates/parts/etc
- [x] Remove breadcrumbs block as it's only there as an example block and is not used
- [x] Turn php parts into patterns
- [x] Implement breadcrumbs
- [x] Align navigation to the right
- [x] Fix issue with parent theme style giving a 404
- [x] https://github.com/WordPress/translate-w-org-env/pull/8


psrpinto:
After quite some effort and digging into WordPress internals of how themes are loaded, I've made some progress here but there are still some things to fix. See the Screen capture section in the PR description.

akirk:
Fantastic! Looks like we're getting close!

akirk (commented):
Thanks for making this happen, I can see how this has been a lot of trial and error. While trying to wrap my head around the technique you are proposing, I was wondering if it would be wiser to reverse the template hierarchy:

Instead of wrapping everything with a "page" that then renders header, page content, footer, we could provide header and footer and explicitly include them as blocks in the specific page template.

Since it might be a little hard to explain in text, I created this branch showing what I mean: https://github.com/WordPress/wporg-gp-translation-events/compare/new-design-setup...new-design-setup-suggestion

I think one key thought that might take your approach further is that `do_blocks()` can already render multiple blocks. So instead of having lots of `do_blocks()` calls in your templates, we can just run `do_blocks( ob_get_contents() )` and write the blocks in the PHP HTML verbatim.

Using my approach, the My Events page can have this HTML:
```html
<!-- wp:wporg-translate-events-2024/header /-->
<!-- wp:paragraph --><p>my-events</p><!-- /wp:paragraph -->
<!-- wp:wporg-translate-events-2024/footer /-->
```

What do you think?

akirk:
Furthermore we can then replace the filter with just a block attribute: https://github.com/WordPress/wporg-gp-translation-events/commit/8ea412dd23fe7dac67cd7caedf40708e3c70a164

psrpinto:
Great points @akirk! I'm working on incorporating your suggestions into this PR.

psrpinto:
@akirk I've incorporated all suggestions you made:

- Make header its own block, which takes the page title as attribute.
- Make footer its own block.
- Minimize the number of `do_blocks()` calls.
- Remove the `page` "wrapper" block and make the route call the page-specific block instead.

I also took the liberty of adding a `render_header( string $title )` convenience function so that individual pages don't need to do the `wp_json_encode()` and phpcs ignore dance.

themes/wporg-translate-events-2024/blocks/header/render.php
```diff
@@ -0,0 +1,22 @@
...
+		<meta charset="<?php bloginfo( 'charset' ); ?>" />
+		<meta name="viewport" content="width=device-width, initial-scale=1" />
+		<title><?php echo esc_html( $html_title ); ?></title>
+		<?php wp_head(); ?>
+	</head>
+	<body <?php body_class(); ?>>
+		<?php wp_body_open(); ?>
+		<div class="wp-site-blocks">
+			<?php // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>
+			<?php echo $site_header; ?>
```
akirk, line 20: I am not quite sure what the benefit is to use a variable here vs using the outside do_blocks. Then we could just do this:

```suggestion
<?php require __DIR__ . '/site-header.php'; ?>
```

themes/wporg-translate-events-2024/functions.php
```diff
@@ -0,0 +1,145 @@
...
+			echo '<meta ' . esc_attr( $name ) . '="' . esc_attr( $key ) . '" content="' . esc_attr( $value ) . '" />' . "\n";
+		}
+	}
+}
+
+function render_header( string $title ): void {
+	$json = wp_json_encode( array( 'title' => $title ) );
+	// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+	echo "<!-- wp:wporg-translate-events-2024/header $json /-->";
+}
```
akirk: I am not sure I am very happy with such a function. If we want to make it easier, we could do something more generic like this:
```php
<?php block_with_attributes( 'wp:wporg-translate-events-2024/header', array( 'title' => 'My Events' ) );
```

```suggestion
function block_with_attributes( string $block, array $attributes ): void {
	$json = wp_json_encode( $attributes );
	// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
	echo "<!-- $block $json /-->";
}
```

akirk (approved):
Nice! It's ok for me to move ahead with this, though I have added a couple remarks. I guess in particular I am unsure about your hesitation to just use `do_blocks()` over the whole PHP output, why is that?

psrpinto:
> I am unsure about your hesitation to just use do_blocks() over the whole PHP output, why is that?

This is something that I should've have explained earlier, apologies. The reason there can't be a single `do_blocks()` call over the entire page (header, page content, footer) is described as follows.

The header block calls `wp_head()`, which results in the styles that have been registered being output as `<link>` tags. Similarly, the footer block calls `wp_footer()` which will output the `<script>` tags.

This means that when `do_blocks()` is called on the header, all styles must have already been registered. Similar for footer and scripts.

So before calling `do_blocks()` on the header and footer, we must first call `do_blocks()` on the page content, so that styles and scripts required by the page content are registered.

A similar thing is done on the header block itself, because for example the local-navigation-bar block has its own styles. So we must call `do_blocks()` on it before the header calls `wp_head()`.

---

There was actually an issue in the code you reviewed, because this behaviour wasn't happening: we were not correctly registering styles and scripts because we were not calling `do_blocks()` on the page content before the header/footer.

I've now pushed a commit that fixes this.

themes/wporg-translate-events-2024/functions.php
psrpinto: I've pushed a commit that removed that function. If in the future we end up requiring such a function, I agree with your suggestion of having a generic  `block_with_attributes()`.

themes/wporg-translate-events-2024/blocks/header/render.php
psrpinto: I provided some information on why I think this is needed in https://github.com/WordPress/wporg-gp-translation-events/pull/321#issuecomment-2225359200

psrpinto:
Just to add to what my comment above, the `global-header-footer` block uses a similar strategy: https://github.com/WordPress/wporg-mu-plugins/blob/trunk/mu-plugins/blocks/global-header-footer/blocks.php#L387-L421

akirk (approved):
Thanks for explaining! Looks good to move forward with this!

themes/wporg-translate-events-2024/functions.php
lynnbgriffith: Not a big deal in this context, especially because you've removed the function for now, however: it's worth noting that taking the output of `wp_json_encode`/`json_encode` directly into an HTML context like this shouldn't be assumed to be safe. There are a variety of flags that can passed along in the second argument to ensure that the encoded JSON doesn't break in that context. Some combination of the following bitwise flags would do it: `JSON_HEX_QUOT|JSON_HEX_TAG|JSON_HEX_AMP|JSON_HEX_APOS`.

psrpinto: Great point about not assuming the data to be safe. Before we use this theme in production, I will make sure that all `wp_json_encode`/`json_encode` calls are safe. Thanks!

PR #322, opened 2024-06-07
Don't show active events in translation table if user is not logged in

psrpinto:
The list of active events was shown above the translation table when user is not logged in:

image

This PR fixes that.

### DIFF
--- a/includes/event/event-repository.php
+++ b/includes/event/event-repository.php
@@ -480,6 +480,11 @@ class Event_Repository implements Event_Repository_Interface {
 	): Events_Query_Result {
 		$this->assert_pagination_arguments( $page, $page_size );
 
+		// A user who is not logged-in, does not have any events.
+		if ( 0 === $user_id ) {
+			return new Events_Query_Result( array(), $page, 0 );
+		}
+
 		$args = array_replace_recursive(
 			$args,
 			array(
--- a/wporg-gp-translation-events.php
+++ b/wporg-gp-translation-events.php
@@ -367,6 +367,10 @@ class Translation_Events {
 	 * @throws Exception
 	 */
 	public function add_active_events_current_user(): void {
+		if ( ! is_user_logged_in() ) {
+			return;
+		}
+
 		$event_repository            = self::get_event_repository();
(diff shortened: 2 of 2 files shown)


includes/event/event-repository.php
```diff
@@ -480,6 +480,11 @@ private function execute_events_query(
 	): Events_Query_Result {
 		$this->assert_pagination_arguments( $page, $page_size );
 
+		// A user who is not logged-in, does not have any events.
+		if ( 0 === $user_id ) {
+			return new Events_Query_Result( array(), $page, 0 );
```
psrpinto, line 485: This is not required to fix the bug, but the repository should not return any events for user `0`.

amieiro: I am not sure about this, because this `execute_events_query` is used in methods that don't require a logged user, like `get_upcoming_events`. Maybe in the future, we want to show the upcoming events to the visitors without forcing them to be logged in. I think this check should be moved to methods like `get_events_for_user`, `get_current_events_for_user`,... in another PR, because these methods involves a logged user. What do you think?

amieiro (approved):
The PR resolves the bug. It works fine!

includes/event/event-repository.php
psrpinto: Just to make sure we're on the same page, normally there are two possible scenarios:

1. `$user_id` is not passed as an argument, and thus takes the default value of `null`. This is the case for methods that are not about the events of a user (e.g. `get_upcoming_events()` like you mention).
2. `$user_id` is set to an int greater than 0: This is the case for methods that are about the events of user, like `get_current_events_for_user()`.

There's a third "hidden" case (that should not happen), where `$user_id` is 0, and this change is only for this case.

----

I agree that `execute_events_query()` should not be called with `$user_id` set to zero. I added this change here so that it is guaranteed that when it is erroneously called with 0, it returns nothing.

Though I agree that ideally that check should be done on the individual methods instead, like you suggested 👍

psrpinto:
Merging so we can get the fix deployed.

PR #324, opened 2024-06-12
Increase width for Contributors

akirk:
I'd like to propose to increase the width for Contributors. [Before](https://translate.wordpress.org/events/2024/wordcamp-europe/):
Screenshot 2024-06-12 at 10 33 18

After:
Screenshot 2024-06-12 at 10 33 11


psrpinto (approved):
Thanks for fixing!

trymebytes (approved):
Tested and works as expected!

amieiro (approved):
Tested and it works fine!

PR #325, opened 2024-06-15
Add a notice about hosting an event

akirk:
Adds this new notice:
Screenshot 2024-06-15 at 13 39 37

### DIFF
--- a/templates/home.php
+++ b/templates/home.php
@@ -19,8 +19,9 @@ Templates::header(
 	),
 );
 ?>
-
 <div class="event-page-wrapper">
+	<div class="notice">
+		<p><?php echo wp_kses( sprintf( __( 'Do you want to host your own event? <a href="%s">Find more information here</a>.', 'gp-translation-events' ), 'https://make.wordpress.org/polyglots/2024/05/29/translation-events-inviting-gtes-to-create-and-manage-events/' ), array( 'a' => array( 'href' => array() ) ) ); ?></p>
 <div class="event-left-col">
 <?php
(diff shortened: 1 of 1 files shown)


templates/home.php
```diff
@@ -19,8 +19,19 @@
...
 ?>
-
 <div class="event-page-wrapper">
+	<div class="notice" style="padding: .5rem">
+		<?php
+		echo wp_kses(
+			sprintf(
+			// translators: %s is a link to a page about hosting events.
+				__( 'Do you want to host your own event? <a href="%s">Find more information here</a>.', 'gp-translation-events' ),
+				'https://make.wordpress.org/polyglots/2024/05/29/translation-events-inviting-gtes-to-create-and-manage-events/'
```
psrpinto, line 29: The post only has information on how GTEs can create events, right? If the user reading this message is not a GTE, I think they will be confused because they will follow the link but will not find an answer on how they can create an event.

akirk: The link is somewhat temporary (also inserted via a variable) as long we don't have a Polyglots Handbook page. I agree it's not great but this is the best that we have right now.

psrpinto (approved):

amieiro (approved):
Tested and it works fine.

templates/home.php
amieiro, line 29: I have added a section in the [GlotPress handbook](https://make.wordpress.org/polyglots/handbook/for-editors/translation-collaboration-team-building-tips/translation-events-kit/#tracking-the-event-with-translation-events), so we can update the post URL to a more stable URL.
```suggestion
				'https://make.wordpress.org/polyglots/handbook/for-editors/translation-collaboration-team-building-tips/translation-events-kit/#tracking-the-event-with-translation-events'
```

PR #327, opened 2024-06-19
Fix issue in event form handler due to missing argument

psrpinto:
Since https://github.com/WordPress/wporg-gp-translation-events/pull/319, `Notifications_Schedule` constructor requires a `DateTimeImmutable` as first argument.

Thanks @trymebytes for finding this issue. I guess this means the form is currently broken.


trymebytes (approved):
Tested and works! Thank you for fixing!

PR #331, opened 2024-06-25
Track Onsite and Offsite attendees

### LINKED ISSUE #328: Track Onsite and Offsite attendees
It'd be useful to know who is attending remotely, so let's add another option to track this.

We could try to allow users to submit this themselves

Screenshot 2024-06-25 at 12 19 23

but also add a way to set this on the Manage Attendees page:

Screenshot 2024-06-25 at 12 23 48

trymebytes:
Fixes #328

In this PR we add a feature that allows attendees choose how they want to attend an event either as an onsite attendee or remote attendee. We added an extra button to the event page for this.
Screenshot 2024-06-25 at 12 19 23

We also updated the attendees table by adding a column for "Remote", we also added a button that allows the host update the attendance status of an attendee. 

Screenshot 2024-06-25 at 12 23 48

### DIFF
--- a/includes/attendee/attendee-adder.php
+++ b/includes/attendee/attendee-adder.php
@@ -93,6 +93,4 @@ class Attendee_Adder {
 
 		return intval( $translation_count ) <= 10;
 	}
-
-
 }
--- a/includes/attendee/attendee-adder.php
+++ b/includes/attendee/attendee-adder.php
@@ -21,13 +21,9 @@ class Attendee_Adder {
 	 *
 	 * @throws Exception
 	 */
-	public function add_to_event( Event $event, Attendee $attendee, $is_remote_attendee = false ): void {
+	public function add_to_event( Event $event, Attendee $attendee ): void {
 		if ( $this->check_is_new_contributor( $event, $attendee->user_id() ) ) {
 			$attendee->mark_as_new_contributor();
-		}
-
-		if ( $is_remote_attendee ) {
-			$attendee->mark_as_remote_attendee();
 		}
 
 		$this->attendee_repository->insert_attendee( $attendee );
--- a/includes/attendee/attendee-adder.php
+++ b/includes/attendee/attendee-adder.php
@@ -21,9 +21,13 @@ class Attendee_Adder {
 	 *
 	 * @throws Exception
 	 */
-	public function add_to_event( Event $event, Attendee $attendee ): void {
+	public function add_to_event( Event $event, Attendee $attendee, $is_remote_attendee = false ): void {
 		if ( $this->check_is_new_contributor( $event, $attendee->user_id() ) ) {
 			$attendee->mark_as_new_contributor();
+		}
+
+		if ( $is_remote_attendee ) {
+			$attendee->mark_as_remote_attendee();
 		}
 
 		$this->attendee_repository->insert_attendee( $attendee );
@@ -89,4 +93,6 @@ class Attendee_Adder {
 
 		return intval( $translation_count ) <= 10;
 	}
+
+
 }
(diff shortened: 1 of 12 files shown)


includes/attendee/attendee.php
```diff
@@ -65,6 +71,14 @@ public function mark_as_new_contributor(): void {
 		$this->is_new_contributor = true;
 	}
 
+	public function mark_as_remote_attendee(): void {
+		$this->is_remote = true;
+	}
+
+	public function mark_as_in_person_attendee(): void {
```
psrpinto, line 78: 👌

includes/attendee/attendee-adder.php
```diff
@@ -21,11 +21,15 @@ public function __construct( Attendee_Repository $attendee_repository ) {
 	 *
 	 * @throws Exception
 	 */
-	public function add_to_event( Event $event, Attendee $attendee ): void {
+	public function add_to_event( Event $event, Attendee $attendee, $is_remote_attendee = false ): void {
```
psrpinto: I think we can remove the `$is_remote_attendee` argument, since the `Attendee` already carries that information. This means we need to set it in the code that calls this method (I'm leaving another comment there).

includes/routes/user/attend-event.php
```diff
@@ -46,15 +53,17 @@ public function handle( int $event_id ): void {
...
+
 		if ( $attendee instanceof Attendee ) {
 			if ( $attendee->is_contributor() ) {
 				$this->die_with_error( esc_html__( 'Contributors cannot un-attend the event', 'gp-translation-events' ), 403 );
 			}
 			$this->attendee_repository->remove_attendee( $event->id(), $user_id );
 		} else {
 			$attendee = new Attendee( $event->id(), $user_id );
-			$this->attendee_adder->add_to_event( $event, $attendee );
+			$this->attendee_adder->add_to_event( $event, $attendee, $is_remote_attendee );
```
psrpinto: As per my other comment, we can instead pass `$is_remote_attendee` to the constructor of the `$attendee` above.

includes/routes/user/attendance-mode.php
```diff
@@ -0,0 +1,64 @@
...
+use Wporg\TranslationEvents\Routes\Route;
+use Wporg\TranslationEvents\Translation_Events;
+use Wporg\TranslationEvents\Urls;
+
+/**
+ * Toggle whether the current user is attending an event onsite or remotely.
+ * If the user is not currently marked as remote attendee, they will be marked as remote attendee.
+ * If the user is currently marked as as remote attendee, they will be marked as not remote attendee.
+ */
+class Attendance_Mode_Route extends Route {
```
psrpinto, line 17: As a suggestion for a future PR, I think it could make sense to consolidate in a single route the functionality for managing attendance of a user to a given event.

There are now four routes that deal with attendance of a user to a given event:

- `/events/attend/$event_id`
- `/events/host/$event_id/$user_id`
- `/events/attendance-mode/$event_id/$user_id`
- `/events/$event_id/attendees/remove/$event_id`

We could replace the first three with a single route that creates or updates an `attendee`, e.g. `/events/$event_id/attendees/$user_id/set`, which could take the following parameters:

- `is_host`
- `is_remote`

This route can then be called both by the user (when clicking the attending buttons), and from an admin, when using the Manage Attendees page.


Additionally, we could change the structure of the existing `remove` route (the last one on the list above), so we end up with the following two routes:

- `/events/$event_id/attendees/$user_id/set`
- `/events/$event_id/attendees/$user_id/remove`

psrpinto (commented):
Well done, I tested and works 👌

Only got one suggestion for this PR, and another one for a future PR

psrpinto (approved):

includes/routes/user/attendance-mode.php
trymebytes: I agree with this!

PR #333, opened 2024-07-04
Add attendance mode to form

### LINKED ISSUE #332: Add option for offsite and remote attendance
Event creators should be able to select whether they want their event attendance mode to be either onsite or remote or both at the point of creating the event.

trymebytes:
Fixes #332 
In this PR we had another field to the event form that allows users choose the attendance mode for their event.

Screenshot 2024-07-10 at 15 14 38

### DIFF
--- a/templates/event-details.php
+++ b/templates/event-details.php
@@ -289,7 +289,7 @@ Templates::header(
 						<input type="submit" class="button is-secondary attending-btn" value="<?php esc_attr_e( "You're attending", 'gp-translation-events' ); ?>" />
 					<?php else : ?>
 						<?php if ( ! $event->is_remote() ) : ?>
-							<input type="submit" class="button is-primary attend-btn" value="<?php esc_attr_e( 'Attend Event Onsite', 'gp-translation-events' ); ?>"/>
+							<input type="submit" class="button is-primary attend-btn" value="<?php esc_attr_e( 'Attend Event On-site', 'gp-translation-events' ); ?>"/>
 						<?php endif; ?>
 						<?php if ( $event->is_remote() || $event->is_hybrid() ) : ?>
 							<input type="submit" name="attend_remotely" class="button is-primary attend-btn" value="<?php esc_attr_e( 'Attend Event Remotely', 'gp-translation-events' ); ?>"/>
--- a/templates/event-details.php
+++ b/templates/event-details.php
@@ -290,9 +290,11 @@ Templates::header(
 					<?php else : ?>
 						<?php if ( ! $event->is_remote() ) : ?>
 							<input type="submit" class="button is-primary attend-btn" value="<?php esc_attr_e( 'Attend Event On-site', 'gp-translation-events' ); ?>"/>
-							<p class="onsite-btn-note">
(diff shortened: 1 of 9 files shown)


akirk:
I think we need to find a way to discourage remote attendees for onsite-only events. Maybe like this?

Screenshot 2024-07-11 at 11 10 12

akirk:
Regarding spelling, I am not sure if we need to spell this `On Site` or `on-site` but I don't think it can be `Onsite`.

trymebytes:
I would go with `On-site`

templates/parts/event-form.php
```diff
@@ -71,6 +71,17 @@
 				?>
 			</select>
 		</div>
+		<div>
+			<label for="event-attendance-mode"><?php esc_html_e( 'Attendance Mode', 'gp-translation-events' ); ?></label>
+				<input type="radio" id="event-attendance-mode-hybrid" name="event_attendance_mode" <?php echo $event->is_hybrid() ? esc_attr( 'checked' ) : ''; ?> value="hybrid" required>
+				<span title="Attendees can attend remotely and on-site"><?php esc_html_e( 'Hybrid (Remote and On-site)', 'gp-translation-events' ); ?></span>
```
akirk: If you use a `label` HTML tag like this, the text is clickable to activate the radio box. Please also do this for the other radios.
```suggestion
				<label title="Attendees can attend remotely and on-site"><input type="radio" id="event-attendance-mode-hybrid" name="event_attendance_mode" <?php echo $event->is_hybrid() ? esc_attr( 'checked' ) : ''; ?> value="hybrid" required>
				<?php esc_html_e( 'Hybrid (Remote and On-site)', 'gp-translation-events' ); ?></label>
```

templates/event-details.php
```diff
@@ -288,8 +288,15 @@ function ( $contributor ) {
 					<?php if ( $user_is_attending ) : ?>
 						<input type="submit" class="button is-secondary attending-btn" value="<?php esc_attr_e( "You're attending", 'gp-translation-events' ); ?>" />
 					<?php else : ?>
-						<input type="submit" class="button is-primary attend-btn" value="<?php esc_attr_e( 'Attend Event Onsite', 'gp-translation-events' ); ?>"/>
-						<input type="submit" name="attend_remotely" class="button is-primary attend-btn" value="<?php esc_attr_e( 'Attend Event Remotely', 'gp-translation-events' ); ?>"/>
+						<?php if ( ! $event->is_remote() ) : ?>
+							<input type="submit" class="button is-primary attend-btn" value="<?php esc_attr_e( 'Attend Event On-site', 'gp-translation-events' ); ?>"/>
+							<p class="onsite-btn-note">
```
akirk: This label is also shown for Hybrid events but shouldn't be.

akirk (changes requested):
Almost works well, just the message about "only onsite" is displaye for hybrid events also.

templates/event-details.php
trymebytes: Thanks for catching this

akirk (approved):
Looks good now!

trymebytes:
Thanks!

psrpinto (commented):
Code looks great! 👌

PR #334, opened 2024-07-10
Add edit-attendance-mode capability

trymebytes:
In this PR we add a new capability to handle when the attendance mode on the event form can be edited.


psrpinto (approved):
👍

PR #336, opened 2024-07-12
Fix page layout

psrpinto:
There was an issue where the content of the page was not center-aligned on wide viewports. This PR fixes that so that on narrow viewports the content is left-aligned, and on wide ones it's center-aligned. It now behaves the same as [events.wordpress.org](https://events.wordpress.org/upcoming-events/).

This fix also has the nice side effect of also fixing the footer, so it stays anchored to the bottom of the window.

Additionally, this PR also fixes a bug that caused the styles of blocks referenced by the footer not to be registered. The bug is visible in the screenshots below (they were taken before the fix), where there are dots before each social link in the footer, and the social images are not coloured.

## Screen captures
The red border shows the size of the container.

### Before
Screenshot 2024-07-12 at 16 19 26

### After
Screenshot 2024-07-15 at 15 42 23

---
Screenshot 2024-07-15 at 15 42 38

---
Screenshot 2024-07-15 at 15 43 16

### DIFF
--- a/themes/wporg-translate-events-2024/blocks/footer/index.php
+++ b/themes/wporg-translate-events-2024/blocks/footer/index.php
@@ -3,13 +3,10 @@
 register_block_type(
 	'wporg-translate-events-2024/footer',
 	array(
-		// The $attributes argument cannot be removed despite not being used in this function,
-		// because otherwise it won't be available in render.php.
-		// phpcs:ignore Generic.CodeAnalysis.UnusedFunctionParameter.Found
-		'render_callback' => function ( array $attributes ) {
+		'render_callback' => function () {
 			ob_start();
-			include_once __DIR__ . '/render.php';
-			return do_blocks( ob_get_clean() );
+			require __DIR__ . '/render.php';
+			return ob_get_clean();
 		},
 	)
 );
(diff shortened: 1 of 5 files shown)


psrpinto:
I'm putting this back in draft as I identified some issues:

1. There's now a deprecation warning
2. We should do a single do_blocks for the page and its container

psrpinto:
The issues mentioned above are now fixed. The footer was being incorrectly rendered, so its styles were not being ouput. This is now fixed, so the social links render correctly now, you can see they are now colored.

psrpinto:
@trymebytes @amieiro Could you take a look at this one? It fixes the layout and I think it should be risk-free to merge it.

trymebytes (commented):
I tested and it looks good but I noticed some excess padding, screenshot below;

Screenshot 2024-08-06 at 21 59 22

Here is how it looks  currently on the trunk branch
Screenshot 2024-08-06 at 22 00 03
It doesn't align with the breadcrumb above, I suppose this some styling issues.

psrpinto:
@trymebytes The padding is deliberate, if you look at https://events.wordpress.org/upcoming-events/ you will see that it also has that padding. That's because the page content is centered, while the breadcrumbs are left-aligned.

trymebytes (approved):
Looks good to me!

PR #337, opened 2024-07-15
Evaluate whether we should use the query loop block

psrpinto:
This PR is not intended to be merged.

---

## Goal
The easiest/more convenient way to create pages when using the block theme introduced in #321, would probably be to use the block editor. For example, in the "My events" page, we would need to display the list of events of the current user. To do this, we would:

1. Open the block editor
2. Add an `event-list` block
3. Configure the filters and other attributes for the `even-list` (e.g. show upcoming events only, for currently active user)
4. Configure the "template" that should be used for each entry in the list
5. Copy the markup into the "My events" page (`themes/wporg-translate-events-2024/blocks/pages/events/my-events/render.php`)

Ideally, the markup of the page would look something like:

```html
<!-- wp:wporg-translate-events/event-list {"filter":"current"} -->
    <!-- wp:wporg-translate-events/event-title /-->
    <!-- wp:wporg-translate-events/event-start /-->
<!-- /wp:wporg-translate-events/event-list -->
```

This PR explores two options for implementing such an `event-list`:

1. Using the Query Loop block (more concretely, a custom variation of the Query Loop block)
2. Using a custom `event-list` block.

## The goal is kind of to remove the need for the `Routes`

Another way to look at what we're trying to achieve here is that there would no longer be a need for the `Routes`. The logic of the route would instead be expressed through block attributes. For example, imagine that we want to display two pages:

- Page A displays the list of past events
- Page B displays the list of upcoming events

Previously, we would create two Routes, and each route would fetch the list of events from the database, then pass that list to the template of the respective page.

With the new approach we're exploring here, the only thing the routes would do is render the block for the respective page (e.g. `themes/wporg-translate-events-2024/blocks/pages/events/A.php`).

Then each page would render the `event-list` block with different attributes:

```html
<!-- Page A -->
<!-- wp:wporg-translate-events/event-list {"filter":"past"} /-->
```

```html
<!-- Page B -->
<!-- wp:wporg-translate-events/event-list {"filter":"upcoming"} /-->
```

## Using a variation of the Query Loop block

> Note that the Query Loop example in this PR is not a complete implementation, it's only provided as a very simple example of how to implement a variation of the Query Loop.

AFAICT, this option would have the following issues:

- It only works for post types, the same pattern would not be usable for things for which there is no post type. It would not be possible to implement, for example, an `attendee-list` using this pattern, because there is no `attendee` post type.
- It would not be possible to use existing code as is, as the Query Loop block works directly with `WP_Query`. We have created our own abstractions on top of `WP_Query` (e.g. `Event_Repository`). This would be unfortunate because the repositories are tested, and provide an abstraction that I consider to be useful.
- The same problem as in the previous point would apply to the rest of `translate.wordpress.org` if we eventually apply the same patterns there. It will likely be a lot of effort to turn glotpress entities (`translation`, `locale`, etc) into something that is usable with the Query Loop block.
- I'm not sure it's possible to provide custom attributes to the Query Loop block through the block editor. We would likely be limited to the attributes that the Query Loop block supports. This would be a serious limitation IMO.

## Using an `event-list` block

Using an `event-list` block that does not rely on the Query Loop block would solve all of the issues described above, while providing a very similar experience: you add the block through the block editor, configure how you want the list to be filtered (and potentially provide other attributes), and the list immediately renders in the block editor.

This would mean that we can reuse the existing code for retrieving events from the database, and other code as well. In practice, this option moves some of the logic out of the `Route` and into the block itself. 

So that we can use the existing code, the block must be rendered on the server. So we use `ServerSideRender` in the block editor. In the future, if need be, the blocks can be adapted to also render in JavaScript (for the editor), but this will require implementing custom REST routes.

## What remains to be done

Above I provided the following snippet as the goal for this PR:

```html
<!-- wp:wporg-translate-events/event-list {"filter":"current"} -->
    <!-- wp:wporg-translate-events/event-title /-->
    <!-- wp:wporg-translate-events/event-start /-->
<!-- /wp:wporg-translate-events/event-list -->
```

However, I haven't yet figured out how the inner blocks can be passed to the `event-list` block to be used as template for each entry in the list. This would be the next step here.

## References

- https://github.com/ryanwelcher/advanced-query-loop
- Query Loop Block
    - [Query Loop block](https://wordpress.org/documentation/article/query-loop-block/)
    - [Extending the Query Loop block](https://developer.wordpress.org/block-editor/how-to-guides/block-tutorial/extending-the-query-loop-block/)
    - [How to Modify the Query Loop Block to Filter by a Custom Field](https://wpfieldwork.com/modify-query-loop-block-to-filter-by-custom-field/)
    - [Building a book review grid with a Query Loop block variation](https://developer.wordpress.org/news/2022/12/20/building-a-book-review-grid-with-a-query-loop-block-variation/)
    - [`query_loop_block_query_vars`](https://developer.wordpress.org/reference/hooks/query_loop_block_query_vars/)
- Block variations
    - [Block variations](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-variations/)
    - [How to register block variations with PHP](https://developer.wordpress.org/news/2024/03/14/how-to-register-block-variations-with-php/)
- Passing data to blocks
    - [New Feature: The Block Bindings API](https://make.wordpress.org/core/2024/03/06/new-feature-the-block-bindings-api/)
    - [Using Block Variations with the Block Bindings API (video)](https://www.youtube.com/watch?v=qD5U7TNUKTQ)

### Query Loop
Screenshot 2024-07-16 at 18 24 19

### Custom block
Screenshot 2024-07-18 at 15 40 38


akirk:
Here I proposed to cache the event objects when querying the repository, then only pass the ids to the blocks, and then use the event repository cache to get the events: https://github.com/WordPress/wporg-gp-translation-events/compare/example-loop-event-object

PR #338, opened 2024-07-15
Add event list item component

trymebytes:
In this PR we start implementing the pages in blocks and we have focused on the `My Events` page. Here's the figma design we need to implement.

image

**Update**
Here is how the `My Events` page looks at this moment

image

In this PR, the focus is to implement the block theme for the `My Events` page and to start with, we would create blocks 
 for these event properties;
- [x] Start Date
- [x] Title
- [x] Attendance Mode
- [x] Host/Attending label

**Update**
image

As we don't collect event location at the moment, I have replaced the event location with the event attendance mode that shows if an event is remote/onsite/hybrid.


themes/wporg-translate-events-2024/blocks/components/events/start-date/index.php
```diff
@@ -0,0 +1,17 @@
...
+	array(
+		// The $attributes argument cannot be removed despite not being used in this function,
+		// because otherwise it won't be available in render.php.
+		// phpcs:ignore Generic.CodeAnalysis.UnusedFunctionParameter.Found
+		'render_callback' => function ( array $attributes ) {
+			global $post;
+			ob_start();
+			$event_start = get_post_meta( $post->ID, '_event_start', true );
+			include_once __DIR__ . '/render.php';
+			return do_blocks( ob_get_clean() );
```
akirk: The `do_blocks()` is not necessary since we just have plain HTML rendered.

themes/wporg-translate-events-2024/blocks/components/event/index.php
```diff
@@ -0,0 +1,77 @@
...
+			}
+			ob_start();
+			?>
+	
+			<h3 class="wporg-marker-list-item__title">
+					<a href="<?php echo esc_url( \Wporg\TranslationEvents\Urls::event_details( $attributes['id'] ) ); ?>">
+						<?php echo esc_html( $translation_events_lookup[ $attributes['id'] ]->title() ); ?>
+					</a>
+					<?php
+					if ( $event_flag ) :
```
akirk: This can generate a notice/warning. Better to do this:
```suggestion
					if ( isset( $event_flag ) ) :
```

trymebytes:
In this PR, a cache was introduced to store a copy of an event when fetched which has led to the modification of the `delete_event()` and `trash_event()` functions, where the stored event is removed when an event is trashed or deleted. If the event is not removed, the tests would fail.

themes/wporg-translate-events-2024/blocks/components/event/attendance-mode.php
```diff
@@ -0,0 +1,14 @@
+<?php namespace Wporg\TranslationEvents\Theme_2024;
+use Wporg\TranslationEvents\Translation_Events;
+
+register_block_type(
+	'wporg-translate-events-2024/components-event-attendance-mode',
```
psrpinto: We could place these blocks directly under `blocks/` instead of under an extra `components/` directory. It would keep the directory structure simpler, and we could also remove the `components-` prefix in the block slug.

themes/wporg-translate-events-2024/blocks/components/event/attendance-mode.php
```diff
@@ -0,0 +1,14 @@
...
+	array(
+		// phpcs:ignore Generic.CodeAnalysis.UnusedFunctionParameter.Found
+		'render_callback' => function ( array $attributes ) {
+			$event = Translation_Events::get_event_repository()->get_event( $attributes['id'] );
+			return '<div class="wporg-marker-list-item__location">
```
psrpinto: This `wporg-marker-list-item__location` appears to be specific to WordCamp Events, so I don't think we should rely on them. I think we should only rely on classes from the parent theme, or classes from our own theme.

themes/wporg-translate-events-2024/blocks/components/event/list.php
```diff
@@ -0,0 +1,49 @@
...
+			if ( empty( $event_ids ) ) {
+				return;
+			}
+			ob_start();
+			?>
+			<div class="wp-block-wporg-event-list">
+			<ul class="wporg-marker-list__container">
+				<?php
+				foreach ( $event_ids as $event_id ) {
+					$event = Translation_Events::get_event_repository()->get_event( $event_id );
```
psrpinto, line 21: I think we should add a new `get_events( array $event_ids ): Event[]` method to the repository, so that we don't do one query per event.

themes/wporg-translate-events-2024/blocks/components/event/date.php
```diff
@@ -0,0 +1,24 @@
+<?php namespace Wporg\TranslationEvents\Theme_2024;
+use Wporg\TranslationEvents\Translation_Events;
+
+register_block_type(
+	'wporg-translate-events-2024/components-event-start',
+	array(
+		// phpcs:ignore Generic.CodeAnalysis.UnusedFunctionParameter.Found
+		'render_callback' => function ( array $attributes ) {
```
psrpinto: I think each block should only get as attribute the data that it requires. So instead of the block getting an event id and then using that id to get the event from the database, and then retrieving the data it cares about from the event, the block would instead get the data it needs directly as an attribute. In this case, the block would get the start datetime, instead of the event id.

The responsibility of this block is to render the start datetime of the event, but the block itself should not need to know that events are stored in the database. For example, you should be able to call this block with an instance of an event that is already in memory.

I think applying this principle will simplify things quite a bit, for example, we probably won't need to add caching to the repository because the blocks won't need to interact with the database.

This would apply to all blocks we're adding, not just this one.

themes/wporg-translate-events-2024/blocks/components/event/excerpt.php
```diff
@@ -0,0 +1,13 @@
+<?php namespace Wporg\TranslationEvents\Theme_2024;
+use Wporg\TranslationEvents\Translation_Events;
+
+register_block_type(
+	'wporg-translate-events-2024/components-event-excerpt',
+	array(
+		// phpcs:ignore Generic.CodeAnalysis.UnusedFunctionParameter.Found
+		'render_callback' => function ( array $attributes ) {
+			$event = Translation_Events::get_event_repository()->get_event( $attributes['id'] );
```
psrpinto: I don't think we need to get the event from the database, we can pass the `id` attribute directly to the `get_the_excerpt()` function.

themes/wporg-translate-events-2024/blocks/components/event/title.php
```diff
@@ -0,0 +1,40 @@
...
+			$event = Translation_Events::get_event_repository()->get_event( $attributes['id'] );
+			$event_ids = $attributes['event_ids'] ?? array();
+			$user_id = get_current_user_id();
+			$current_user_attendee_per_event = Translation_Events::get_attendee_repository()->get_attendees_for_events_for_user( $event_ids, $user_id );
+			$current_user_attendee = $current_user_attendee_per_event[ $event->id() ] ?? null;
+
+			if ( $current_user_attendee ) {
+				$event_flag = 'Attending';
+				if ( $current_user_attendee->is_host() ) {
+					$event_flag = 'Host';
```
psrpinto: I think this block should only render the title, and we should have another block that renders the attendance/hosting labels. We might not want to always render the attendance/hosting labels every time we render the title.

themes/wporg-translate-events-2024/style.css
```diff
@@ -9,3 +9,137 @@ License: GNU General Public License v2 or later
 License URI: http://www.gnu.org/licenses/gpl-2.0.html
 Text Domain: wporg-translate-events-2024
 Template: wporg-parent-2021
+*/
+
+.wporg-marker-list__container {
```
psrpinto, line 26: I'm not sure why other themes named these classes like this, do you know? I mean, why "marker"? I think we could use clearer/simpler names, unless there's some convention that they are following.

includes/event/event-repository.php
```diff
@@ -96,6 +97,8 @@ public function trash_event( Event $event ) {
 		if ( ! $result ) {
 			return false;
 		}
+		unset( $this->cache[ $event->id() ] );
```
psrpinto: I think we shouldn't add caching here, because there's a version of the repository that is specific for caching (`Cached_Event_Repository`), so adding caching in the non-cached repository feels a bit weird to me. In addition, as I mention in another comment, I think that if we pass only the data that the block requires as an attribute, instead of passing the event id, there will be no need for caching.

(The `Cached_Event_Repository` might not have been the greatest idea, but given it currently exists, I think we should first re-evaluate whether we should remove it before adding caching directly in this repository).

psrpinto (commented):
I think we can simplify some things here, specifically by passing as attributes to the blocks only the data that they require.

themes/wporg-translate-events-2024/blocks/components/event/attendance-mode.php
trymebytes: The idea was to have the directory a bit more organized so we can have something like this, see below;
`/blocks/components/event` - holds all the different components of an event (title, date, excerpt, list etc)
`/blocks/components/attendee` - holds all the components of an attendee e.g (Name, profile url, profile picture url etc.) 
Similar to the `/blocks/pages/events` directory.

themes/wporg-translate-events-2024/blocks/components/event/date.php
trymebytes: I agree, thank you for this.

themes/wporg-translate-events-2024/style.css
trymebytes: I have no idea why

themes/wporg-translate-events-2024/blocks/components/event/attendance-mode.php
psrpinto: I see what you mean, and I agree it's useful to somehow group related blocks together. However, I think the term "component" when in the context of the block editor is commonly used to refer to React components, and not blocks. So I think that using the term component to refer to a block can be a bit confusing.

For example, if you look at [the core `button` block](https://github.com/WordPress/gutenberg/tree/trunk/packages/block-library/src/button), you can see that, internally, [it uses the `RichText` component](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/button/edit.js#L270C6-L270C14), which comes from [`gutenberg/packages/block-editor/src/components/rich-text/`](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/components/rich-text/native/index.native.js#L104). Note that `RichText` is not a block, it's a React component that is meant to be reused by blocks.

Currently, we are not using React components because we haven't had to manipulate how blocks render in the block editor, but if in the future we need to do so, at that point we would have both React components, and block components, which I think would be confusing.

So instead of grouping related blocks by placing them in a `components/foo` directory, we could instead prefix the block with it's logical "thing", e.g. `foo-title`. In this scenario, the blocks directory would look something like:

- `blocks/`
    - `event-title/`
    - `event-date/`
    - ...
    - `attendee-name/`
    - `attendee-profile-url/`
    - ...

These blocks can then be referenced with `wporg-translate-events/event-title` or `wporg-translate-events/attendee-name`.

This is what I was suggesting, sorry I should I have made it clearer.

trymebytes: Thank you so much for the detailed response, it's a lot clearer now and thanks for the reference to [gutenberg core](https://github.com/WordPress/gutenberg/tree/trunk/packages/block-library/src), very helpful.

includes/event/event-repository.php
trymebytes: I think it might be helpful to remove the caching from this PR and create an issue where we decide on the best approach.

themes/wporg-translate-events-2024/blocks/event-attendance-mode/index.php
```diff
@@ -0,0 +1,23 @@
...
+	array(
+		// phpcs:ignore Generic.CodeAnalysis.UnusedFunctionParameter.Found
+		'render_callback' => function ( array $attributes, $content, $block ) {
+			if ( ! isset( $block->context['postId'] ) ) {
+				return '';
+			}
+			$event_id = get_the_ID();
+			$event = Translation_Events::get_event_repository()->get_event( $event_id );
+			if ( ! $event ) {
+				return $event_id;
```
akirk: This should rather be this:
```suggestion
				return '';
```

Generally, since we have this section for retrieving the event repeated across many blocks, I propose to create a utility funtion or a callback wrapper in a follow-up.

akirk (approved):
As I mentioned, I think the `render_callback` could use a refactoring but ok for me to do this in a follow-up.

themes/wporg-translate-events-2024/blocks/event-attendance-mode/index.php
```diff
@@ -0,0 +1,23 @@
...
+
+register_block_type(
+	'wporg-translate-events-2024/event-attendance-mode',
+	array(
+		// phpcs:ignore Generic.CodeAnalysis.UnusedFunctionParameter.Found
+		'render_callback' => function ( array $attributes, $content, $block ) {
+			if ( ! isset( $block->context['postId'] ) ) {
+				return '';
+			}
+			$event_id = get_the_ID();
```
psrpinto: Could/should this be `$block->context['postId']`?

themes/wporg-translate-events-2024/blocks/event-excerpt/index.php
```diff
@@ -0,0 +1,13 @@
+<?php namespace Wporg\TranslationEvents\Theme_2024;
+use Wporg\TranslationEvents\Translation_Events;
+
+register_block_type(
+	'wporg-translate-events-2024/event-excerpt',
+	array(
+		// phpcs:ignore Generic.CodeAnalysis.UnusedFunctionParameter.Found
+		'render_callback' => function ( array $attributes ) {
+			$event = Translation_Events::get_event_repository()->get_event( $attributes['id'] );
```
psrpinto: The `$event` variable can be removed as it's not being used.

themes/wporg-translate-events-2024/blocks/event-flag/index.php
```diff
@@ -0,0 +1,40 @@
...
+			if ( ! isset( $block->context['postId'] ) ) {
+				return '';
+			}
+			$event_id = get_the_ID();
+			$event = Translation_Events::get_event_repository()->get_event( $event_id );
+			if ( ! $event ) {
+				return $event_id;
+			}
+
+			global $current_user_attendee_per_event;
```
psrpinto: I think we shouldn't rely on global variables, as the block can thus not be called without knowing that this variable needs to be set. Not sure what the correct solution is though, maybe the `Attendee_Repository` also needs caching?

psrpinto (commented):
Looks great! I have a couple of comments

themes/wporg-translate-events-2024/blocks/event-list/index.php
```diff
@@ -0,0 +1,40 @@
+<?php
```
akirk, line 1: Do we still need that file given that we'll want the flexibility that we get by [just embedding the loop directly in pages like My Events](https://github.com/WordPress/wporg-gp-translation-events/pull/338/files#diff-55573c810a23aecb046558264089922b2193247fff039d8697d65495d682e363R7-R28)?

akirk (approved):
I am happy with this, just a question whether we should remove the event-list block.

themes/wporg-translate-events-2024/blocks/event-list/index.php
trymebytes: The event-list block is being used in #342 where we have about 3 different lists on a page. Having the list block makes the code for the homepage look clean, apart from the fact that it makes it easier to make modifications to the list in just one place all instances of the list are updated, I don't see any other reason to keep it. But the reason given above is also valid, perhaps we can remove it now and abstract it later when we have a better way of doing it.

akirk: Hm. What about we re-introduce its use in this PR (i.e. replace the foreach loop in the my-events with the event-list block again), and in #342 (or when we need a variation) we add innerBlocks support to it?

trymebytes: This means we'll have to also re-introduce the `event-flag` attribute too

akirk: I am not sure why. I did what I suggested in [`d32cf43` (#338)](https://github.com/WordPress/wporg-gp-translation-events/pull/338/commits/d32cf4315368d4c8608dc42503ae6d6be084c701)

trymebytes: Oh, got it. Thanks!

themes/wporg-translate-events-2024/blocks/event-list/index.php
```diff
@@ -0,0 +1,45 @@
...
+					?>
+					<li class="wporg-marker-list-item">
+						<!-- wp:wporg-translate-events-2024/event-template <?php echo wp_json_encode( array( 'id' => $event_id ) ); ?> -->
+						<div>
+							<!-- wp:wporg-translate-events-2024/event-title /-->
+							<!-- wp:wporg-translate-events-2024/event-flag /-->
+						</div>
+						<!-- wp:wporg-translate-events-2024/event-attendance-mode /-->
+						<!-- wp:wporg-translate-events-2024/event-start /-->
+						<!-- /wp:wporg-translate-events-2024/event-list-->
```
akirk, line 32: I noticed this typo here. It is not a problem because it's ignored but it should be removed.

trymebytes: Thank you for catching this! it has been fixed in another [commit in this Home page block PR](https://github.com/WordPress/wporg-gp-translation-events/pull/342/commits/eb5292b199a22ab479fe8a5cc6e2abb7d8e674d0)

PR #340, opened 2024-07-17
Set up wp-scripts

psrpinto:
This PR sets up `wp-scripts`, needed for block development.

Install it with:
```
npm install
```

Then build all blocks under `src/` with:

```
npm run build
```

Note this doesn't yet build blocks that are under the theme, at `themes/worg-translate-2024/`.

I've also added an example block, that we can delete when we create the first real block.


trymebytes (approved):
Tested and works for me!

PR #344, opened 2024-07-31
Add the RSS feed in the header

amieiro:
This PR:
- Adds the rel="alternate" link for the events' RSS feed.
- Removes two invalid RSS feeds:
    - https://translate.wordpress.org/feed/
    - https://translate.wordpress.org/comments/feed/

To test it, you need to:
- Inspect the source code before switching to this branch, so you can see the 2 incorrect RSS feeds.
- Switch to this branch.
- Inspect the source code. Now you have a new RSS feed, linking to https://translate.wordpress.org/events/feed/, and you don't have the 2 incorrect RSS feeds.


akirk (approved):
Tested and looks good! Thanks!

PR #326, opened 2024-06-19
Create the RSS 2.0 feed

amieiro:
This PR adds:
- A new class to generate an RSS 2.0 feed.
- Two routes (`/events/feed` and `/events/rss`) to get the RSS feed.
- A new method (`get_current_and_upcoming_events`) in the event repository to get all events that are currently active or happening in the future, in order by the start date.

The:
- The [pubDate](https://www.w3schools.com/xml/rss_tag_pubdate_item.asp) element defines the last publication date for an item. 
  - In the main file, the `pubDate` is the date of the last update of all events in the RSS file. 
  - In the event (item), the `pubDate` is the last update of this event.
- The [lastBuildDate](https://www.w3schools.com/xml/rss_tag_lastbuilddate.asp) element defines the last-modified date of the content of the feed, so it is the same value that the `pubDate` in the main file.

I have used the http://purl.org/rss/1.0/modules/event/ namespace to be able to add the `ev:startdate` and `ev:enddate` elements to each event.

Date format:
- The `pubDate` has the DATE_RSS format: 'D, d M Y H:i:s O'.
- the `ev:startdate` and `ev:enddate` have the DateTimeInterface::ATOM format: 'Y-m-d\TH:i:sP'.

You can validate the RSS feed using the [W3C Feed Validation Service](https://validator.w3.org/feed/#validate_by_input), coping and pasting the generated content. You will get some warning like this, because the local URL is not accessible from this tool:

image

One example with the generated RSS feed:

``` xml
<?xml version="1.0" encoding="UTF-8"?>
<rss xmlns:atom="http://www.w3.org/2005/Atom" xmlns:ev="http://purl.org/rss/1.0/modules/event/" version="2.0">
   <channel>
      <title>WordPress.org Global Translation Events</title>
      <link>https://glotpress.test/glotpress/events/</link>
      <description>WordPress.org Global Translation Events</description>
      <language>en-us</language>
      <pubDate>Sat, 29 Jun 2024 10:13:45 +0000</pubDate>
      <lastBuildDate>Sat, 29 Jun 2024 10:13:45 +0000</lastBuildDate>
      <docs>https://www.rssboard.org/rss-specification</docs>
      <generator>Translation Events</generator>
      <atom:link href="https://glotpress.test/glotpress/events/rss/" rel="self" type="application/rss+xml" />
      <item>
         <title>July event</title>
         <link>https://glotpress.test/glotpress/events/july-event__trashed/</link>
         <description>Lorem ipsum</description>
[...]

PR #347, opened 2024-08-21
Fix broken form

trymebytes:
Found out that the event form is broken on form submission and returns the error below;

`Uncaught Error: Argument 8 passed to Wporg\TranslationEvents\Event\Event::__construct() must be an instance of DateTimeImmutable or null, string given, called in /Users/tosinoguntuyi/Sites/translate/wp-content/plugins/wporg-gp-translation-events/includes/event/event-form-handler.php on line 233`

In this PR, we add `null` values for the newly introduced `created_at` and `updated_at` function arguments.

### DIFF
--- a/includes/event/event.php
+++ b/includes/event/event.php
@@ -42,7 +42,6 @@ class Event {
 	private string $status;
 	private string $title;
 	private string $description;
-	private DateTimeImmutable $created_at;
 	private DateTimeImmutable $updated_at;
 	private string $attendance_mode;
 
@@ -59,7 +58,6 @@ class Event {
 		string $status,
 		string $title,
 		string $description,
-		DateTimeImmutable $created_at = null,
 		DateTimeImmutable $updated_at = null,
 		string $attendance_mode = 'onsite'
 	) {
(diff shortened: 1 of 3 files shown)


includes/event/event.php
```diff
@@ -59,8 +59,8 @@ public function __construct(
 		string $status,
 		string $title,
 		string $description,
-		DatetimeImmutable $created_at = null,
-		DatetimeImmutable $updated_at = null,
+		DateTimeImmutable $created_at = null,
+		DateTimeImmutable $updated_at = null,
```
psrpinto, line 61: I'm a bit confused by `$created_at` and `$updated_at`, they don't seem to be used anywhere, should we just remove them?

psrpinto (approved):

psrpinto:
Thanks for fixing!

includes/event/event.php
trymebytes: Thank you for reviewing! These properties were introduced in the [RSS PR](https://github.com/WordPress/wporg-gp-translation-events/pull/326/) and `$updated_at` was used [here](https://github.com/WordPress/wporg-gp-translation-events/blob/4eccfd26a3a42bd3fc68b1b07ab45faee0721cf5/includes/routes/event/rss.php#L105).

Perhaps, it's `$created_at` that may be removed.

psrpinto: Ah indeed, I missed that, I only searched for `created_at`.

amieiro (approved):
Works nice! Thank you for catching this bug!

PR #348, opened 2024-08-22
Merge Event_Repository_Cached into Event_Repository class

trymebytes:
In this PR we merging the classes Event_Repository_Cached and Event_Repository into a single `Event_Repository` class to allow the usage of cached results. We have only implemented caching for the `get_event()` function at the moment.

The Event_Repository_Cached class will then be deleted on completion of the merging.


psrpinto (commented):
I think there's two types of caching that we should consider:

1. Single events cache: so that you can call `$repository->get_event( $event_id )` and the Event is retrieved from cache instead of hitting the database. This is useful, for example, for a _block_ that gets the event id as attribute, and needs to retrieve the `Event`.
2. Query results cache: caches the results of queries (e.g. "Upcoming events").

In the current implementation of `Cached_Event_Repository`, it only deals with 2., and it only caches a single type of query ("Current events"). Additionally, it has specific code for that specific query, but I think we can cache all queries, without requiring specific code for each. (More on this below.)

Since that 1. is (I think?) the only caching type that we need right now (so that blocks can get a specific event from cache), and that the implementation in `Cached_Event_Repository` for 2. is "wrong", I would first remove `Cached_Event_Repository` from the codebase, along with `Event_Repository_Interface` (which is no longer needed since there is now a single implementation of `Event_Repository`).

Once `Cached_Event_Repository` and `Event_Repository_Interface` are out of the way, I would implement 1. into `Events_Repository::get_event()`, by caching the whole `Event` instance, indexed by its id.

I believe that's all we would need right now, but if later we implement 2., I would add caching in `execute_events_query()`, which is the method that all queries call. Given that calling this method with the same arguments must result in the same results being returned (provided of course that event information hasn't changed in the database), we can cache the queries by indexing them by a hash of the function's arguments. Something like:

```php
# pseudocode
private function execute_events_query(
		int $page,
		int $page_size,
		array $args,
		array $filter_by_ids = array(),
		int $user_id = null,
		bool $include_created_by_user = false
	): Events_Query_Result {
	$hash = hash_function([
		$page,
		$page_size,
		serialize_function($args),
		serialize_function($filter_by_ids),
		$user_id,
		$include_created_by_user
	]);
	
	$cached = get_cache($hash);
	if ($cached) {
		return $cached;
	}
	
	// ...
	
	$query_result = new Events_Query_Result( $events, $page, $query->max_num_pages )
	cache( $hash, $query_result );

	// Also cache single events.
	for ($event of $events) {
		cache($event->id, $event);
	}

	return $query_result;
}
```

trymebytes:
Thank you @psrpinto for the review!

psrpinto (approved):

trymebytes:
Caching for `get_event()` for fetching a single event has been implemented in this PR, we will handle other cache implementations in a follow up PR.

PR #349, opened 2024-08-29
Add a date in the RSS header when we don't have current and future events

amieiro:
In the RSS feed, if we don't have current and future events, some fields in the header (pubDate and lastBuildDate) are empty, so the RSS feed is not valid:

### Empty header
image

### Invalid check
image

This PR uses:
- The last updated event from the past events, if we don't have future events.
- The Unix epoch date if we don't have any event in the system.

### Last updated event from the past events
image

### Unix epoch date
image

### DIFF
--- a/includes/routes/event/rss.php
+++ b/includes/routes/event/rss.php
@@ -92,17 +92,30 @@ class Rss_Route extends Route {
 	/**
 	 * Get the most recent event's pub date.
 	 *
-	 * @param Event[] $events Array of events to use for the pub date.
+	 * Returns the last updated_at date if there are any current and upcoming events.
+	 * Otherwise, returns the last end date of the past events.
+	 * If there are no events at all, returns the date of the Unix epoch.
+	 *
+	 * @param Event[] $events Array of current and upcoming events to use for the pub date.
 	 *
 	 * @return string|null
 	 */
 	private function document_pub_and_build_date( array $events ): ?string {
 		if ( empty( $events ) ) {
 			$events = $this->event_repository->get_past_events()->events;
-		}
-		if ( empty( $events ) ) {
-			$zero_date = new DateTimeImmutable( '@0' );
-			return $zero_date->format( DATE_RSS );
+			if ( empty( $events ) ) {
+				$zero_date = new DateTimeImmutable( '@0' );
+				return $zero_date->format( DATE_RSS );
+			}
+
+			$end_date = $events[0]->end();
+			foreach ( $events as $event ) {
+				if ( $event->end() > $end_date ) {
+					$end_date = $event->end();
+				}
+			}
+
+			return $end_date->format( DATE_RSS );
 		}
 
 		$pub_date = $events[0]->updated_at();
--- a/includes/routes/event/rss.php
+++ b/includes/routes/event/rss.php
@@ -39,10 +39,7 @@ class Rss_Route extends Route {
 		);
 		$last_20_events_post = wp_get_recent_posts( $args );
 
-		if ( isset( $last_20_events_post[0]['post_modified_gmt'] ) ) {
-			$this->send_headers( $last_20_events_post[0]['post_modified_gmt'] );
-		}
-
+		$this->send_headers( $this->document_pub_and_build_date( $last_20_events_post, 'Y-m-d H:i:s' ) );
 		$rss_feed = $this->get_rss_20_header( $last_20_events_post );
 		foreach ( $last_20_events_post as $event_post ) {
 			$event = $this->event_repository->get_event( $event_post['ID'] );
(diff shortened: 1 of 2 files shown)


akirk (commented):
Could we also use another query than `get_current_and_upcoming_events`?

amieiro:
> Could we also use another query than `get_current_and_upcoming_events`?

In the [event repository](https://github.com/WordPress/wporg-gp-translation-events/blob/cb6cef19c0178afc8acdf34122e165e6a558abd3/includes/event/event-repository-interface.php) we don't have a public method to get all events, so we can:
- Make public the [get_events_active_between()](https://github.com/WordPress/wporg-gp-translation-events/blob/cb6cef19c0178afc8acdf34122e165e6a558abd3/includes/event/event-repository.php#L435) method and use the Unix epoch date and the current date as parameters.
- Use this approach: use the `get_current_and_upcoming_events` and if its output is empty, use the `get_past_events` method to use them, if it is not empty.

akirk:
I am coming from the thought that the RSS feed content should only change when a new event is added. I'd vote for creating a new function that will just retrieve the top 10 events sorted by event start date.

amieiro:
> I am coming from the thought that the RSS feed content should only change when a new event is added. I'd vote for creating a new function that will just retrieve the top 10 events sorted by event start date.

If a current or upcoming event changes some relevant info, like the description, the RSS should change. 

The idea with the `pubDate` and the `lastBuildDate` fields in the header (both fields have the same value) is that they will reflect the last time the RSS feed has been build (last time a current or upcoming event has changed): e. g., if I update the description of a current or upcoming event, this `updated_at` date will be used in the `pubDate` and the `lastBuildDate` fields, in the header.

If we don't have current or upcoming events, we should use the last finished date for the past events (here I need to change the code), because this last event has been removed from the RSS feed when the event has finished, so the file has changed. 

And if we don't have any event in the system, we show the Unix epoch date.

akirk:
@amieiro will change this to just use `get_posts()` with the right CPT.

includes/routes/event/rss.php
```diff
@@ -91,18 +101,25 @@ private function get_item( Event $event ) {
...
+	 *
+	 * @param WP_Post[] $events_post Array of last events, as WP_Post objects.
 	 *
 	 * @return string|null
 	 */
-	private function document_pub_and_build_date( array $events ): ?string {
-		if ( empty( $events ) ) {
-			return null;
+	private function document_pub_and_build_date( array $events_post ): ?string {
+		$pub_date = new DateTimeImmutable( '@0' );
```
akirk, line 196: What's `@0`? Could you add a comment?

akirk:
@amieiro I took the liberty to add HTTP caching, so something like this works:
```
curl -vH'if-none-match: "b9fa8c94ce941f16cafd14f1103f1e95"' https://glotpress.local/events/feed/
> GET /events/feed/ HTTP/1.1
> Host: glotpress.local
> User-Agent: curl/8.7.1
> Accept: */*
> if-none-match: "b9fa8c94ce941f16cafd14f1103f1e95"
> 
< HTTP/1.1 304 Not Modified
< Server: nginx/1.27.1
< Date: Tue, 10 Sep 2024 14:52:47 GMT
< Connection: keep-alive
< X-Powered-By: PHP/8.3.11
< Vary: origin
< Last-Modified: Tue, 27 Aug 2024 09:38:38 GMT
< ETag: "b9fa8c94ce941f16cafd14f1103f1e95"
< 
```
[
I took most of the code from Core WordPress](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/class-wp.php#L435-L581), I would love to use the core code itself but I didn't find a good way since GlotPress interferes.

akirk:
Can we try to stay close to what WordPress does, for example here: https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/feed-rss2.php Maybe we can also use some functions that it uses to ensure correctness.

akirk (approved):
It works well for me, I'll approve already but please take a look at my comments.

includes/routes/event/rss.php
amieiro: Done!

amieiro:
Thank you for adding this cache layer. I have been testing it in Chrome and Firefox and I have added some small tweaks:
- I use the last update event to send the cache headers to the browser, not the first one, because we can have an update in another event and, in this situation, we need to update the RSS file.
- The browsers (both Chrome and Firefox) add "W/" at the begging of the Etag, to indicate [weak validation.](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag#directives) If it exists, I remove from `$client_etag` to avoid problems in the `$client_etag === $wp_etag` check. 

Please, take a look at this and if it is ok to you, we merge this PR.

trymebytes (approved):
Tested and works!

PR #350, opened 2024-08-29
Do not change negative symbol in the diff

amieiro:
Currently, the human diff in the `WordCamp Rio` event is incorrect. 

image

This PR solves it.

image


akirk (approved):
Thanks!

PR #351, opened 2024-09-02
Event details page block

trymebytes:
In this PR, we implement the event-details page block.

**Figma design**
image

**Current Status**
image

### DIFF
--- a/themes/wporg-translate-events-2024/blocks/pages/events/event-details/index.php
+++ b/themes/wporg-translate-events-2024/blocks/pages/events/event-details/index.php
@@ -30,7 +30,7 @@ register_block_type(
 
 			render_page(
 				__DIR__ . '/render.php',
-				__( $event->title(), 'wporg-translate-events-2024' ),
+				esc_html( $event->title() ),
 				$attributes
 			);
 		},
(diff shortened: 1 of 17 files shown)


themes/wporg-translate-events-2024/blocks/pages/events/event-details/index.php
```diff
@@ -20,9 +21,16 @@ function ( $breadcrumbs ): array {
...
+			$event = Translation_Events::get_event_repository()->get_event( $event_id );
+			if ( ! $event ) {
+				return '';
+			}
+			$attributes['event'] = $event;
+
 			render_page(
 				__DIR__ . '/render.php',
-				__( 'Event', 'wporg-translate-events-2024' ),
+				__( $event->title(), 'wporg-translate-events-2024' ),
```
akirk: Never pass variables into translation functions:
```suggestion
				$event->title(),
```

trymebytes: Thanks for catching this!

akirk:
I tested it and it looks quite good! I did notice a problem with the Attendee count. Old (2 attendees):
Screenshot 2024-09-13 at 14 56 57
New (3 attendees):
Screenshot 2024-09-13 at 14 57 31

Did you swap the Attendees and Contributors by accident? 

This now also lost the display in the current user time, could you fix that in an upcoming PR?

trymebytes:
Oh thanks for testing! I'm taking a look.

akirk (approved):
This works well for me, I think we should update the button style of "You attended" in a follow up so that it doesn't look like a clickable button.

PR #353, opened 2024-09-13
Manage attendees page block

trymebytes:
In this PR, we implement the manage-attendees page block.

**Figma design**

image

**Before**

image


trymebytes:
Current preview;

image

akirk (approved):
It works well but in the previous template we had a link to get the to the event page. Would be nice to also get to the Event edit page. You can also do it in a follow up but right now this makes it hard to navigate.

PR #355, opened 2024-09-17
Fix event times stored in wrong timezone in the database

### LINKED ISSUE #354: Bug in timezone calculation
I created the WCUS Translation event and set the time to Los Angeles timezone but on the event details page the time calculated in my timezone is not correct.

Screenshot 2024-09-17 at 10 05 04

Screenshot 2024-09-17 at 10 04 32

akirk:
Fixes #354 

When storing an event in another timezone we need to convert it to UTC before saving it to the db:
Screenshot 2024-09-17 at 14 36 47

9am in Los Angeles needs to be saved as 4pm UTC. 

When editing the event, we need to convert it back into the even timezone so that the local time is being edited.


trymebytes (approved):
Works as expected, thank you!

PR #357, opened 2024-09-17
Fix missing CSS for event bubble

akirk:
Before:
Screenshot 2024-09-17 at 17 05 21

After:
Screenshot 2024-09-17 at 17 58 01


trymebytes (approved):
Tested and works! Thanks for fixing!

PR #361, opened 2024-09-24
Event has ended notice

trymebytes:
In this PR we display a notice if the event has ended to allow users easily see that the event has ended, also the button is replaced with a simple text `You attended this event.`

**Before**

image

**After**

Screenshot 2024-09-24 at 19 06 06


akirk (approved):
Great work using the existing block!

PR #362, opened 2024-09-24
Fix event flag error

### LINKED ISSUE #359: Event List shows "Attending" when an event is over
I've got this old event that correctly shows that I attended (past tense).

Screenshot 2024-09-20 at 14 21 18

In the list, it shows as "Attending" (present progressive). This should also be past tense for past events.

Screenshot 2024-09-20 at 14 20 52

trymebytes:
Fixes #359 
The event flag that tells if a user is attending or hosting on the my events page and it had a bug which this PR fixes. In this PR we ensure that for an event that has ended;
if user is a host, we display `Hosted`
if user is only an attendee, we display `Attended`

If the event has not ended;
if user is a host, we display `Hosting`
if user is only an attendee, we display `Attending`

**Before**
Screenshot 2024-09-24 at 23 51 14

**After**
Screenshot 2024-09-24 at 23 47 13


akirk (approved):
Nice, thanks!

PR #370, opened 2024-10-14
Add video icon to flag remote events

### LINKED ISSUE #335: Flag for remote event
We could consider adding an icon to the titles of remote events on the homepage to distinguish them easily. See mockup below;

Screenshot 2024-07-12 at 13 31 20

trymebytes:
Fixes #335 To allow easy identification of remote-only events.

**Before**
Screenshot 2024-10-14 at 16 12 22

**After**
Screenshot 2024-10-14 at 16 10 49

### DIFF
--- a/themes/wporg-translate-events-2024/blocks/attendee-list/render-list.php
+++ b/themes/wporg-translate-events-2024/blocks/attendee-list/render-list.php
@@ -33,7 +33,7 @@ if ( empty( $attendees_not_contributing ) || ! current_user_can( 'edit_translati
 			?>
 			/-->
 			<?php if ( $attendee->is_remote() ) : ?>
-				<!-- wp:wporg-translate-events-2024/remote-attendance-icon /-->
+					<!-- wp:wporg-translate-events-2024/remote-attendance-icon <?php echo wp_json_encode( array( 'css_class' => 'show-on-gravatar' ) ); ?> /-->
 			<?php endif; ?>
 			</div>
 		<!-- /wp:group -->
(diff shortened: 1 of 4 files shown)


themes/wporg-translate-events-2024/blocks/event-title/index.php
```diff
@@ -19,6 +19,12 @@
 			}
 			$url = Urls::event_details( $event->id() );
 			?>
+			<?php
+			if ( 'remote' === $event->attendance_mode() ) {
+				// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+				echo do_blocks( sprintf( '<!-- wp:wporg-translate-events-2024/remote-attendance-icon %s  /-->', wp_json_encode( array( 'inline_css' => 'float:left; margin-right:3px' ) ) ) );
```
amieiro: I think we should add some margin in the top. You can see the 5px in the row (1) vs the current margin in the row (2). 

image

```suggestion
				echo do_blocks( sprintf( '<!-- wp:wporg-translate-events-2024/remote-attendance-icon %s  /-->', wp_json_encode( array( 'inline_css' => 'float:left; margin:5px 3px 0 0' ) ) ) );
```

One question about adding the inline CSS: what do you thing about extracting these CSS to a CSS file instead of adding them inline?

amieiro (changes requested):

themes/wporg-translate-events-2024/blocks/event-title/index.php
trymebytes: Resolved in #377

trymebytes:
Resolved in #377

PR #371, opened 2024-10-15
Deleted events page

### LINKED ISSUE #368: Update design for deleted events page
We need to use the new theme on the trash events page.

**Before**

Image

trymebytes:
Fixes #368 

**Before**

image

**After**
Screenshot 2024-10-16 at 18 55 34


amieiro (changes requested):
I have tested it and the look and feel is nice. 

image

Some comments:
- I think we should move the "Edit icon" to a button like the other two ("Restore" and "Delete permanently").
- I think we should have a confirm button in both buttons: something like "Are you sure you want to delete this event", and something similar for the "Restore" button. These actions are important, and the "Delete" is destructive, so I think a confirmation button will improve the UI.
- When I click in the  "Restore" button in the "WordCamp Kerala" event, the event is not restored.

akirk:
We could also use icons instead of buttons?

For example for restoring: 

Screenshot 2024-11-18 at 12 23 23

And delete permanently:
Screenshot 2024-11-18 at 12 24 26

trymebytes:
Updated view
image

akirk:
@trymebytes That's a lot of icons now next to the title (I know this is just for admins but still). Can you try some UI variations with putting them in a dedicated column or just right aligning them? Thanks!

trymebytes:
Here are the variations;

**Mockup 1**
Screenshot 2024-12-02 at 14 54 07

**Mockup 2**
Screenshot 2024-12-02 at 14 17 00

cc @akirk @amieiro

akirk:
Thanks! Visually I like version 1 better, what do you think?

trymebytes:
I agree, it's visually appealing and user-experience wise, I also prefer it to version 2.

obenland:
Closing as this repository is being archived. Translation Events is now maintained in the WordPress.org Meta repository (https://meta.trac.wordpress.org/ticket/8368). If this is still relevant, please open a Meta Trac ticket under "Translate Site & Plugins" (https://meta.trac.wordpress.org/newticket?component=Translate+Site+%26+Plugins), or a PR against https://github.com/WordPress/wordpress.org.

PR #376, opened 2024-11-05
Define $editor_options which is used but not set.

dd32:
Fixes a PHP warning when an event doesn't have any translation sets.

```
E_WARNING: Invalid argument supplied for foreach() in wporg-gp-translation-events/templates/translations/footer.php:14
```


akirk (approved):
Thank you!

PR #379, opened 2024-11-22
Load more button

### LINKED ISSUE #378: Load more button
At the moment we use pagination links which makes the user experience not so great when there are multiple list sections on a page, we need to add a `Load more` button to append the next page of events to the current list.

Below is how it currently works;
Image

trymebytes:
Fixes #378 

- [x] Modify existing list endpoint to fetch next list of events
- [x] Call endpoint  via AJAX when button is clicked

### DIFF
--- a/assets/js/translation-events.js
+++ b/assets/js/translation-events.js
@@ -112,18 +112,21 @@
 								if (!response.ok) {
 									throw new Error(`HTTP error! Status: ${response.status}`);
 								}
-								return response.text();
+								return response.json();
 							})
 							.then(response => {
-								$(this).hide();
-								const $html = $(response);
+								const $html = $(response.data.html);
 								const $listItems = $html.find('li');
+								if ( 0 === response.data.nextPage ){
+									$(this).hide();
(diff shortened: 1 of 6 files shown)


assets/js/translation-events.js
```diff
@@ -99,6 +99,37 @@
 
 					}
 				);
+				$( document ).on(
+					'click',
+					'.load-more-events-btn',
+					function ( e ) {
+						let eventType = $( this ).data('event-type');
+						let nextPage = $( this ).data('event-next-page');
+						const url = `http://translate.test/events/?${encodeURIComponent(eventType)}=${encodeURIComponent(nextPage)}&format=html`;
```
amieiro:
```suggestion
						const url = `https://translate.wordpress.org/events/?${encodeURIComponent(eventType)}=${encodeURIComponent(nextPage)}&format=html`;
```

amieiro (changes requested):

amieiro (approved):

